PR #2, opened 2019-12-23
Initial setup.py & resource loading via pkg_resources

Korijn:
This PR brings:

- Creating platform wheels with `python setup.py bdist_wheel`
- Loading files from the `resources` subdir in development, installed and frozen states (via `pkg_resources.resource_filename`)

Also note:

* I don't think `resource_filename` will work without installing `wgpu` in development mode using `python setup.py develop`, so don't forget about that.


setup.py
```diff
@@ -0,0 +1,38 @@
...
+setup(
+    name=NAME,
+    version=VERSION,
+    packages=find_packages(exclude=["tests", "tests.*", "examples", "examples.*"]),
+    package_data={
+        f'{NAME}.resources': ['*.dll', '*.so', '*.dylib', '*.h', '*.idl'],
+    },
+    python_requires=">=3.6.0",
+    license=open("LICENSE").read(),
+    description=open("README.md").readlines()[2],
```
almarklein: This seems fragile, but let's fix that later. A one-line string in this file is good enough imo :)

PR #14, opened 2019-12-24
Wheels are no longer bound to CPython

### LINKED ISSUE #6: Make binary wheels that are not bound to Python version
The wheels that are currently build are for CPython 3.6+. Wheels can also be made to work with any (py3) version. This makes it easier for ppl to try it out on very new Python versions, and people on other implementations (thinking mostly of Pypy) can also just use it.

Examples where it's used:
* https://pypi.org/project/imageio-ffmpeg/#files
* https://pypi.org/project/glfw/#files

Korijn:
Fixes #6 

Wheels are now bound to any version of Python 3 and the platform on which the wheel is built.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -1,6 +1,8 @@
 import re
 
 from setuptools import find_packages, setup
+from wheel.bdist_wheel import bdist_wheel as _bdist_wheel
+from wheel.pep425tags import get_platform
 
 
 NAME = "wgpu"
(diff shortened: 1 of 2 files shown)


setup.py
```diff
@@ -10,24 +12,17 @@
...
         _bdist_wheel.finalize_options(self)
-        self.root_is_pure = False
 
 
 setup(
     name=NAME,
     version=VERSION,
     packages=find_packages(exclude=["tests", "tests.*", "examples", "examples.*"]),
     package_data={f"{NAME}.resources": ["*.dll", "*.so", "*.dylib", "*.h", "*.idl"]},
-    python_requires=">=3.6.0",
```
almarklein: don't we still need this? Or does it prevent us from creating the wheels we want. We can also add trove classifiers to specify the supported versions.

Korijn: I thought that it was an issue, but when I tried it again just now it worked just fine. So I restored it. Good call!

PR #17, opened 2019-12-24
Perform imports in Qt GUI backend

Korijn:
Changed the import mechanism a little bit to actually perform the import.


almarklein:
I'm a bit reluctant. Other code will have to import Qt anyway, to e.g. start the application, and things go bad when it imports *the other* qt lib.

Korijn:
> Other code will have to import Qt anyway, to e.g. start the application

This is a strange argument, we import GUI libraries in the other backends too. Also, it's standard practice to import modules in Python, since the importlib will take care of figuring out if something was already imported before or not.

> and things go bad when it imports the other qt lib.

Now this is the real problem; but I've added some code to check if either one has been imported already before going ahead with the import ourselves.

We need to have the import in order to define the widget, so there's no way around it. The code is very friendly towards users. The only scenario that could still pose a problem is where the user has both libraries in their environment, imports neither, and then we import the wrong one. I can add a check to guard against that specific scenario.

almarklein:
> > Other code will have to import Qt anyway, to e.g. start the application
> 
> This is a strange argument, we import GUI libraries in the other backends too.

Haha, true. I should have reversed the order of my arguments: it's dangerous to import a Qt lib here, because we might import the wrong one. Since the user will import it elsewhere anyway, it's not a big deal to require the user to do so before importing `wgpu.gui.qt`.

I think I still prefer doing that. With the current proposal the behavior will depends on the user's environment, which can be confusing. It also means that a script can work fine on one user's machine and not on another.

I like your error message better though :)

Korijn:
Do you want me to change anything here?

almarklein:
IMO we should not change behavior based on the user's system here, so I think we should leave it as it was, except use your better error message ("Import one of ..."). But if you disagree we should discuss :)

Korijn:
I'll trim this down to just the error message then :)

Korijn:
Done. Ready to merge.

PR #18, opened 2019-12-24
Simplify initialization

Korijn:
I felt like the backend import mechanism is a good fit for our needs _internally_, but harder than it needs to be for _users_. So I left the import mechanism intact, but provided an `init` function on the root module for users. Advantages:

* Users can simply `import wgpu` and call `wgpu.init()` to get started right away, without needing to make a choice of backend (Rust is the default). I think this will be just fine for the majority of users.
* Users can still pick a specific backend by providing the `backend` kwarg to `init`. Example: `wgpu.init(backend='js')`

(I also removed `os.environ["RUST_BACKTRACE"] = "0"` since that is something both developers and users should be able to set on their environment, rather than in code.)

### DIFF
--- a/README.md
+++ b/README.md
@@ -56,10 +56,12 @@ The full API is accessable via the main namespace:
 import wgpu
 ```
 
-But to use it, you need to select a backend first. You do this by importing it.
-There is currently only one backend:
+But to use it, you need to initialize the library first. You can provide
+configuration overloads here, such as the choice of backend (default backend is
+Rust, ie: `backend='rs'`):
+
 ```py
-import wgpu.backend.rs
+wgpu.init()
 ```
 
 To give an idea of what this API looks like, here's the API code from the triangle example:
@@ -67,6 +69,7 @@ To give an idea of what this API looks like, here's the API code from the triang
 # ... code to create shaders and GUI are omitted for brevity
 
 import wgpu
+wgpu.init()
 
 async def main(canvas):
 
--- a/examples/triangle_glfw.py
+++ b/examples/triangle_glfw.py
@@ -126,7 +126,7 @@ import glfw
 from wgpu.gui.glfw import WgpuCanvas  # WgpuCanvas wraps a glfw window
 import wgpu
 
-wgpu.init("rs")
+wgpu.init()
 
 
 glfw.init()
--- a/examples/triangle_glfw.py
+++ b/examples/triangle_glfw.py
@@ -124,7 +124,8 @@ import asyncio
 
 import glfw
 from wgpu.gui.glfw import WgpuCanvas  # WgpuCanvas wraps a glfw window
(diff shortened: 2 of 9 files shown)


almarklein:
I agree that an `init()` call is more user-friendly than an import. But the current way does have one big advantage: it avoids dynamic imports, and therefore allows tools like PyInstaller to work without further guidance. That said, only few users will freeze apps, so maybe we should indeed go for `init()`. One can still use the import approach when writing a freeze script. What do you think?

almarklein:
Also, we may call `init()` automatically if `requestadapter` gets called and no backend is selected yet. Another line less in user code.

Korijn:
> But the current way does have one big advantage: it avoids dynamic imports, and therefore allows tools like PyInstaller to work without further guidance. That said, only few users will freeze apps, so maybe we should indeed go for init(). One can still use the import approach when writing a freeze script. What do you think?

Well, I think that pyinstaller includes our entire package once `wgpu` is imported anywhere. I don't think it filters our modules, does it? Either way, we can contribute a hook to the pyinstaller repo.

> Also, we may call init() automatically if requestadapter gets called and no backend is selected yet. Another line less in user code.

Great idea! Just like matplotlib's `use(backend)` function, opt-in. :)

almarklein:
A similar discussion arguably goes for the gui backend.

I just experimented with this a bit. By explicitly importing `wgpu.qt`, it will included *that* module and its dependencies (Qt). When importing `wgpu.tk`, it will *that* module instead, and its dependencies (tk), no Qt libs in the frozen app. Same should go for the API backend (`rs` vs ..).

> we can always contribute a hook to the pyinstaller repo to make it work out of the box.

I don't think that will help, because we still need a way to specify *what* backend to include, so this will probably be the end-user juggling `includes` and `excludes`.

What about:
* We always support selecting gui/api backens by importing them. 
* We document somewhere that this is what ppl should do when using PyInstaller.
* We add a hook in `requestAdapter()` that imports the default api backend when nothing is selected when it's called.
* We don't need an `init()` function just yet.
* When/if we get more API backends, we can add the `init()` function, and keep all the above intact.

Korijn:
Sounds good.

Korijn:
I'll open a new branch to work on your suggestions.

almarklein:
We'd still need a hook in `requestAdapter()`, right?

Korijn:
That wouldn't be pyinstaller compatible though. I'm not sure how I feel about that... :/ it feels like setting the user up for a gotcha

almarklein:
We can document what's needed to freeze an app containing wgpu. And in that code inside `requestAdapter()` that does the import, we can catch `ImportError` and provide a useful error message. That said, you have a point that `wgpu` will behave differently when frozen, and this will cause some tears for someone at some point. Let's discuss this in person.

Issue #12, opened 2019-12-24
Tests ... we need tests


almarklein:
Let's start with flake checks, btw.

Korijn:
FYI latest flake8 gives 568 warnings, after setting max_line_length to 120 it's 227

almarklein:
Some of these types are actually flags or enums. The codegen simply does not resolve them further yet. Others are "descriptors" which we simply have to explain in the docs.

almarklein:
`max_line_length` can be 88, btw, because we use black

Korijn:
I propose ignoring the flake8 max line length warning and also providing type definitions so the type hints can be properly resolved. It's very hard to combine black and flake8 compliance with the codegen-script...

PR #19, opened 2019-12-24
Flake8

Korijn:
Partly addresses #12 

Note: I can't run the codegen-script since I don't have a binary yet. I tried adjusting it to produce noqa comments but I can't verify if it works properly. It probably fails to do so because it is used in conjunction with black, so we can't tell on which lines the violations will end up

Either way, this PR:

* Configures flake8 to be compatible with black (according to their docs)
* Fixes flake8 violations (a few)
* Adds specific error ignores to lines I couldn't/didn't know how to fix (a lot)
* Runs flake8 in CI
* Also runs black on python files in the root of the repo on CI
* Removed existing `# noqa` that didn't trigger violations

---

I'm unsure of the following changes:

```python
from py2spirv import python2spirv, i32, vec2, vec3, vec4
```
Is that how it's supposed to be done?


almarklein:
> I can't run the codegen-script since I don't have a binary yet.

It does not load the binary :)  Also, the files it generates are blackified before writing, so I'm curious what noqa's were needed.

Will have a proper look at this after xmas :)

Korijn:
> It does not load the binary :)

Really! Doh. I should have just tried it.

> Also, the files it generates are blackified before writing, so I'm curious what noqa's were needed.

Black doesn't fix comments that exceed max line length, for example. Also, the undefined type hints.

almarklein:
> Black doesn't fix comments that exceed max line length, for example. Also, the undefined type hints.

Maybe we should simply disable line length checking on flake8.

Korijn:
I would be ok with disabling that check!

PR #20, opened 2019-12-24
Add tk gui support

almarklein:
* Adds `gui/tk.py` implementing a `WgpuCanvas`.
* Add tk example.
* Refactor triangle examples so they all use the same code from a shared `triangle.py`.


almarklein:
I always found it disappointing that visvis/vispy could not render in Tk because it lacks suppor for GL. But here we are - GPU rendering in Python with zero dependencies!

PR #24, opened 2020-01-02
Consume upstream binaries

### LINKED ISSUE #11: Ship with binary libs in our wheels

Korijn:
Closes #11 

For now, we download a wgpu-native v0.5 development snapshot from my personal fork: https://github.com/Korijn/wgpu/releases/tag/v0.5dev

The script tracks the version currently in use, so it can be restored by CI in a DRY fashion.

```
$ python download-wgpu-native.py --help
usage: download-wgpu-native.py [-h] [--version VERSION] [--debug] [--os OS]
                               [--upstream UPSTREAM]

Download wgpu-native binaries and headers from github releases

optional arguments:
  -h, --help           show this help message and exit
  --version VERSION    Version to download (default: v0.5dev)
  --debug              Download debug build instead of release build
  --os OS              Platform to download for (default: linux)
  --upstream UPSTREAM  Upstream repository to download release from (default:
                       Korijn/wgpu)
```

I also changed the `rs` backend to load binaries for all supported platforms rather than just Windows.

The wheels now also include the binaries:

Screenshot from 2020-01-02 09-37-36

Try running it locally!

**NOTE** There are a few differences in wgpu.h and I'm not sure if we need to update anything accordingly. CI doesn't break but on the other hand that is probably due to lack of proper tests. :)

I also noticed that I broke codegen-script quite a bit with all the flake8 annotations. I tried to get it to to work again but then I hit another issue with the black subprocess and it's just very confusing. I guess you would probably be able to fix those issues more quickly than me.


download-wgpu-native.py
```diff
@@ -0,0 +1,101 @@
...
+from zipfile import ZipFile
+
+
+RESOURCE_DIR = os.path.join("wgpu", "resources")
+VERSION_FILE = os.path.join(RESOURCE_DIR, "wgpu_native-version")
+
+
+def get_current_version():
+    with open(VERSION_FILE, mode="r") as fh:
+        return fh.read()
```
almarklein: suggestion: add `.strip()` just to be sure there's no trailing whitespace.

download-wgpu-native.py
```diff
@@ -0,0 +1,101 @@
...
+                raise
+
+
+def get_os_string():
+    if sys.platform.startswith("win"):
+        return "windows"
+    elif sys.platform.startswith("darwin"):
+        return "macos"
+    elif sys.platform.startswith("linux"):
+        return "linux"
```
almarklein: `else`, sensible default, or raise an error?

download-wgpu-native.py
```diff
@@ -0,0 +1,101 @@
...
+        "libwgpu_native.dylib",
+        "wgpu_native.dll",
+        "commit-sha",
+    ]
+    print(f"Extracting {members} to {RESOURCE_DIR}")
+    extract_files(zip_filename, members, RESOURCE_DIR)
+    current_version = get_current_version()
+    if version != current_version:
+        print(f"Version changed, updating {VERSION_FILE}")
+        write_current_version(version)
```
almarklein: So, if I got this right, the file `wgpu_native-version` is in source control, and with it we specify what version we want. And if we use this script to update the version, we also update it as a marker to show what lib is currently in the resource dir. Can we have a description like this in a comment somewhere?

wgpu/resources/commit-sha
```diff
@@ -0,0 +1 @@
+v0.5dev
```
almarklein: This is going to be the git tag for the release that we've downloaded right? Can we add that in a comment too somewhere, plz?

almarklein:
> There are a few differences in wgpu.h and I'm not sure if we need to update anything accordingly.

No, these changes should not break any of our existing code.

> I also noticed that I broke codegen-script quite a bit with all the flake8 annotations.

Mmm, apparently I have not yet tried to run that script since your flake8 PR. I will have a look later :)

download-wgpu-native.py
Korijn: That's right. I would personally not edit it manually at all, and always use the script, but there is nothing stopping you from doing so. :) I'll add instructions to README.md.

wgpu/resources/commit-sha
Korijn: Well, there's two files.. `commit-sha` is extracted from the wgpu-native zipfile and we mostly include it to have full traceability. Our script works with the other file `wgpu_native-version` to track which release we downloaded. It's kind of a coincidence that they match. I'm not sure if we can assume that will always be the case?

download-wgpu-native.py
almarklein: Right, you can edit it by running the script ;)

wgpu/resources/commit-sha
almarklein: Makes sense!

Korijn:
I think we're ready to merge!

setup.py
```diff
@@ -18,11 +19,26 @@ def finalize_options(self):
...
+    "commit-sha",
+]
+if sys.platform.startswith("win"):
+    resources_globs.append("*.dll")
+elif sys.platform.startswith("linux"):
+    resources_globs.append("*.so")
+elif sys.platform.startswith("darwin"):
+    resources_globs.append("*.dylib")
+else:
+    raise RuntimeError(f"Platform '{sys.platform}' not supported")
```
almarklein: I'm not sure if we want to raise here too. We should raise in the download-script, because there simply aint any binaries for platforms other than win, linux and darwin. But ppl may want to install `wgpu-py` on e.g. a raspberry Pi and somehow provide their own `wgpu-native` libs.

Korijn: Awesome, I hadn't considered that

Korijn: It's besides the point, but on raspberry sys.platform apparently returns "linux" :)

I removed the raise from setup.py and rs.py

almarklein: mmm ... so we might need a more sophisticated solution eventually. Though for now this should be enough.

wgpu/backend/rs.py
```diff
@@ -44,7 +44,16 @@ def _get_wgpu_h():
...
     override_path = os.getenv("WGPU_LIB_PATH", "").strip()
-    embedded_path = get_resource_filename("wgpu_native.dll")
+
+    lib_filename = ""
+    if sys.platform.startswith("win"):
+        lib_filename = "wgpu_native.dll"
+    elif sys.platform.startswith("darwin"):
+        lib_filename = "libwgpu_native.dylib"
+    elif sys.platform.startswith("linux"):
+        lib_filename = "libwgpu_native.so"
```
almarklein: ah nice catch to remove it here too

almarklein:
Hoozah!

setup.py
Korijn: there's just tons of constraints that the current python package management solutions (pip, pypi, etc) can't handle. e.g. c-compiler used (musl, gcc) cpu architecture (ARM...) and many more are all factors that we can't specifically build wheels for today. I wouldn't try to solve that here locally, it's a much bigger problem.

PR #27, opened 2020-01-07
WIP: Make things work on all platforms

almarklein:
Misc:
* [x] Ensure `cffi` has high enough version
* [x] Qt canvas works with all libs (PySide2, PyQt5, Pyside, PyQt4).

GLFW canvas (incl. getting window id):
* [x] GLFW canvas on Windows (was already working)
* [x] GLFW canvas on Linux Xorg
* [x] GLFW canvas on Linux Wayland
* [x] GLFW canvas on OS X 

Get surface id from window id:
* [x] Windows (was already working)
* [x] Linux Xorg
* [x] Linux Wayland
* [x] OS X 

Actually tested drawing a triangle to a window:
* [x] Windows
* [ ] Linux Xorg
* [ ] Linux Wayland
* [ ] OS X

Notes:
* On Linux I was able to test that getting a surface id from a window id works, but I was not able to actually draw something, because Rust panicked on `requestDevice()`, probably because Vulkan does not work inside a VM?
* Same story for OS X, but in this case my old Macbook is too old to run High Sierra, which is needed to have Metal.
* Ahum ... converting a window id to a surface id works for any bullocks int that I give the function, so these tests are pretty useless.


almarklein:
My Macbook is too old to run High Sierra, which is required to have Metal :(  so I can't test the wgpu api for osx. I'll just add the code that *should* work, and test that it works on CI using a compute shader.

Korijn:
On linux there's a couple of problems to tackle. First is determining whether we're on X11 or Wayland, and then for each of these we need different variables. X11 requires a Display and a Window identifier, for example.

Korijn:
There's this, but I don't think the wheels on pypi include this module :( https://doc.qt.io/qt-5/qx11info.html

almarklein:
> First is determining whether we're on X11 or Wayland

For `wgpu-native` there's only `wgpu_create_surface_from_metal_layer`, `wgpu_create_surface_from_windows_hwnd` and `wgpu_create_surface_from_xlib`. Does this mean that it won't work (yet) on Wayland?

Korijn:
There's a `wgpu_create_surface_from_wayland` call too: https://github.com/gfx-rs/wgpu/blob/3a11fab693ab87142cc25725e5cd5e7cc46b30d1/wgpu-native/src/device.rs#L93

almarklein:
ah, then that was recently added :)

Korijn:
I just tried this on Linux:

```
korijn@swatcat:~/dev/wgpu-py (git:platforms) ✔ 
$ RUST_BACKTRACE=1 pipenv run python examples/triangle_glfw.py 
thread '<unnamed>' panicked at 'Unable to query surface formats: ERROR_INITIALIZATION_FAILED', src/libcore/result.rs:1165:5
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace.
fatal runtime error: failed to initiate panic, error 5
Aborted (core dumped)
```

Debugging does show that GLFW returns a `surface_id`:

```
> /media/ssd2/dev/wgpu-py/wgpu/backend/rs.py(845)_create_native_swapchain_if_needed()
-> self._internal = _lib.wgpu_device_create_swap_chain(
(Pdb) win_id
71303177
(Pdb) self._surface_id
4294967296
```

Exact location of crash:

```
> /media/ssd2/dev/wgpu-py/wgpu/backend/rs.py(846)_create_native_swapchain_if_needed()
-> self._device._internal, self._surface_id, struct
(Pdb) l
841  	        if self._surface_id is None:
842  	            win_id = self._canvas.getWindowId()
843  	            self._surface_id = get_surface_id_from_win_id(win_id)
844  	
845  	        self._internal = _lib.wgpu_device_create_swap_chain(
846  ->	            self._device._internal, self._surface_id, struct
847  	        )  # device-id
848  	
849  	    def getCurrentTextureView(self):
850  	        # todo: should we cache instances (on their id)?
851  	        # otherwise we have multiple instances mapping to same internal texture
(Pdb) s
thread '<unnamed>' panicked at 'Unable to query surface formats: ERROR_INITIALIZATION_FAILED', src/libcore/result.rs:1165:5
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace.
fatal runtime error: failed to initiate panic, error 5
Aborted (core dumped)
```

PR #30, opened 2020-01-10
Fix, improve, and rerun codegen script

almarklein:
* Remove `noqa` stuff from the classes that get code injected, because it prevents formatting signatures on a single line, which we need to parse the signatures.
* Add a fix for black still formatting long sigs on 3 lines even with huge line length.
* Black also add a comma at the end (previously it did not for this code?)
* The codegen script now also produces a report.


setup.cfg
```diff
@@ -1,4 +1,4 @@
 [flake8]
 max_line_length = 88
-extend-ignore = E203
+extend-ignore = E203, E501, F821
```
almarklein: * Adding long lines (E501) here (is enforced by black anyway).
* Adding undefined type annotations (F821) here until we fix these.

PR #33, opened 2020-01-13
Linux support

### LINKED ISSUE #28: Improved support for Linux (Xorg+tk, Wayland)
It seems like running `wgpu` on a VM is not possible. Or just hard? Please let us know if you know how to :D

In #27 we implemented support for all platforms, but we have not yet been able to verify this by actually running an example.

Checks:
* [ ] Xorg tk
* [x] Xorg qt
* [x] Xorg glfw
* [ ] Wayland tk
* [ ] Wayland qt
* [x] Wayland glfw (kinda: it works, but also crashes)

almarklein:
* Adds methods to get the display id on glfw, and a generic approach that uses the X11 lib via ctypes.
* Updates readme.
* Rename canvas classes and provide an alias with a common name. That way you can see from the class what GUI backend is wrapped.

Closes #28, except:
* Does not work for tk for some reason.
* On Wayland only glfw works, because I don't know (yet) how to get the display id from Wayland.

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,7 +24,7 @@ To get an idea of what this API looks like have a look at [triangle.py](https://
 *This is experimental, work in progress, you probably don't want to use this just yet!*
 
 * We have a few working examples!
-* These *should* work on all platforms but this is currently only tested on Windows (see #28 and #29), so they probably don't :)
+* Support for Windows and Linux. Support for OSX is underway.
 * We have not fully implemented the API yet.
 * The API may change. We're still figuring out what works best.
 * The API may change more. Until WebGPU settles as a standard, its specification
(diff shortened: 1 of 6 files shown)


almarklein:
@Korijn ken jij ff testuh?

Korijn:
```
$ echo $XDG_SESSION_TYPE
x11
```

Screenshot from 2020-01-13 13-24-13

Screenshot from 2020-01-13 13-24-26

Very nice!

Korijn:
I don't have a Wayland system at my disposal...

almarklein:
What OS did you test on? I tested on Ubuntu 19.10

Korijn:
Screenshot from 2020-01-13 14-32-41

PR #40, opened 2020-01-15
Release for tags only

Korijn:
* Trigger builds for master branch and tags that start with `v`
* Only run release job if branch name matches tag name (that's how we can tell the build was triggered by a tag)


almarklein:
So `Build.SourceBranchName` will be `v0.1.3` when I push that tag?

PR #43, opened 2020-01-20
Compute shader support

almarklein:
See commit messages.


almarklein:
Windows CI build fails because `git restore` does not exist. We need to fix that before a release, but for now I'm merging this regardless.

PR #44, opened 2020-01-21
Windows 32 bit builds

### LINKED ISSUE #37: wgpu-py does not work on Windows 32 bit
Because we only have a 64 bit version of wgpu-native. The unfortunate thing is that if you go to python.org on Windows, the default installer you'll be prompted with is for 32bit Python :( ... in 2020 :'(

Options:
* Detect win32 and provide a reasonable error message.
* There seems to be consensus to make python.org offer both versions, but someone needs to implement that. Maybe help?
* Somehow build a 32bit version of wgpu-native (if it's even possible). This looks similar to #36.

Korijn:
Closes #37 and the `git restore` issue

Pinned pip to <20.0.0 because it seems to be broken?

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -35,7 +35,7 @@ jobs:
       targetType: inline
       script: |
         set -ex
-        python -m pip install -U pip
+        python -m pip install -U pip<20.0.0
         pip install -U -r dev-requirements.txt
   - task: Bash@3
     displayName: Create source distribution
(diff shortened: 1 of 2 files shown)


almarklein:
Looking good!

almarklein:
Should we also add 32bit Linux while we're at it?

Korijn:
We can do it but that would be a little more work on CI. On Windows x64 OS, we can run x86 Python easily. That's not the case on Linux, and there is no x86 Linux agent available either.

We'd need to do it via a Docker image like i386/ubuntu and adjust the script to run that job differently from the others

almarklein:
Good point, I assumed the whole OS was 32bit; now I see why the name is `PY_ARCH` ;)

PR #45, opened 2020-01-21
Add utility to run compute shaders, and add tests for it 

almarklein:
* Add `compute_with_buffers()` utility function.
* Add tests for it. This'll be the first time we will be actually running wgpu on CI :)

After this makes it into a release, `python-shader` could use it to better validate the compiler.

~At this this point this is blocked by #47: the added test can't run on CI. If there is no quick solution for it, I propose~ we disable the test on CI, so that we can at least run it locally.

### DIFF
--- a/dev-requirements.txt
+++ b/dev-requirements.txt
@@ -6,5 +6,4 @@ setuptools
 wheel==0.31.1
 twine
 python-shader>=0.3.3
-numpy
 auditwheel; sys_platform == 'linux'
--- a/dev-requirements.txt
+++ b/dev-requirements.txt
@@ -5,4 +5,6 @@ requests
 setuptools
 wheel==0.31.1
 twine
+python-shader>=0.3.3
+numpy
 auditwheel; sys_platform == 'linux'
--- a/examples/compute_noop.py
+++ b/examples/compute_noop.py
@@ -1,17 +1,17 @@
 """
 Example compute shader that does ... nothing but copy a value from one
-buffer into another
+buffer into another.
 """
 
-import numpy as np
+import ctypes
+
 import wgpu
+import wgpu.backend.rs  # Select backend
+from wgpu.utils import compute_with_buffers  # Convenience function
 from python_shader import python2shader, i32, Array
 
-# Select backend
-import wgpu.backend.rs
 
-
-# %% Shader
+# %% Shader and data
 
 
 @python2shader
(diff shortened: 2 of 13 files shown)


wgpu/_compute.py
```diff
@@ -0,0 +1,152 @@
+"""
+Simple high-level utilities for doing compute on the GPU.
+
+The purpose of wgpu-py is to provide a Pythonic wrapper around wgpu-native.
+In principal, a higher-level API is not within the scope of the project.
+However, by providing a few un-opinionated utility functions, other projects
+(like python-shader) can use wgpu (e.g. in their tests) without having
+to keep track of changes in wgpu itself.
+"""
```
almarklein: This is something that needs thought. How far should we go here? I feel pretty confident about adding this function, since it is very "self contained", you give it some arrays and what you expect as output, it runs the shader, and returns the buffer contents. But should we have something similar for textures? Or support rendering something to an offscreen texture?

wgpu/_compute.py
```diff
@@ -0,0 +1,152 @@
...
+            The spec must be a tuple ``(*shape, dtype)``. A buffer is created
+            for each binding and its data is read and returned. The binding can be
+            one that is also used as an input array (in which case the GPU buffer
+            is the same).
+        shader (bytes, shader-object): Any compatible shader object representing
+            a SpirV compute shader.
+        n (int, tuple, optional): The number provided to dispatch. Can be an int
+            or a 3-tuple of ints to specify (x, y, z). If not given or None,
+            the first value of the first spec in output_arrays is used.
+    """
```
almarklein: @Korijn do you have any thoughts about how to represent these "output specs"? The idea is that we need to know the shape and dtype so that we a) know the size of the buffer, and b) can "cast" the data before returning it.

wgpu/_compute.py
```diff
@@ -0,0 +1,152 @@
...
+
+    # How can we get numpy arrays if numpy is not imported?
+    if "numpy" not in sys.modules:
+        raise RuntimeError("need numpy arrays.")
+    np = sys.modules["numpy"]
```
almarklein: This ain't pretty. Maybe it's a better idea to put utils like this in a submodule/subpackage that's not imported by default. But that depends a bit on the earlier comment about our intentions with this higher level API.

Korijn: Well, I think there's two concerns:

(1) Dependencies: if we want to make wgpu-py as broadly usable as possible, I think we should aim for being dependency free across the entire library, to avoid "gotcha" moments or even giving off the impression that we are tied to numpy and python-shader after all. So for example, instead of numpy, people could use the stdlib module `struct`, and instead of `python-shader`, people could `glsl2spirv` from the vulkan SDK. That should be the case throughout the library.

(2) Import time; Python imports are slow, and so we should try to keep importing wgpu-py fast. Probably it would be best to keep wgpu-py focused on a small core that is imported in one go, and keep other optional utilities separate, or opt-in.

wgpu/_compute.py
Korijn: I think what you've got going on now is pretty good! I just wonder if it needs to be a dict if the keys are incrementing integers? It could just be a list or tuple right?

wgpu/_compute.py
Korijn: I think that wgpu-py should be focused on a small core, we can provide some utilities in an optional "util" module or subpackage to keep import times fast, but also let's make sure we don't end up building half of visvis2 here. :)

Korijn:
For now, let's just add an option to skip the tests and merge so we can continue with development.

wgpu/_compute.py
almarklein: > but also let's make sure we don't end up building half of visvis2 here. :)

exactly

wgpu/_compute.py
almarklein: Because (I think) the numbers don't *have* to be incremental. And also to allow specifying the same index for an input and output.

Dynd has this nice feature that the dtype includes the shape, e.g. `10x10xfloat32`. That would have been nice here, but Numpy does not have such a thing, does it?

wgpu/_compute.py
almarklein: Fortunately, `python-shader` is not a dependency, you can pass in raw SPirV bytes too :)

Yeah, I think it would be better to not import this by default. I'm not sure about numpy, using struct or ctypes arrays could maybe work, but numpy is so much easier to work with. What if we did:
```py
try:
    import numpy as np
except ImportError:
    raise ImportError("wgpu does not need numpy, but this utility submodule does.")
```

Korijn: Well yes, numpy is convenient, but it's also a big dependency in terms of size on disk. Some people prefer to stick to things like struct to be able to ship lightweight software.

wgpu/_compute.py
Korijn: No, there is a project going on to add that to the numpy typeshed but I don't believe they ever got that to work... I don't think it's a huge deal though.

Korijn: Here it is:  https://github.com/numpy/numpy-stubs

almarklein:
Ready for final checkup. Changes since last review:
* The new module was moved to `utils` subpackage.
* We already had a utils module, so I renamed that to `_coreutils`.
* No more numpy, but ctypes arrays. An extra nice thing about this is that `output_arrays` must now be specified with just a ctypes array type, which includes the length :)
* A wrapper that works with numpy arrays is easily made, because numpy operates well with ctypes. (but that wrapper is not in our scope).
* The tests are skipped if wgpu cannot get a device (this is checked once in a test run).

examples/compute_noop.py
```diff
@@ -78,6 +100,10 @@ def compute_shader(input, buffer):
...
-data = buffer2.mapRead()
-# data = await buffer2.mapReadAsync()
-print(np.frombuffer(data, np.int32))
+# Read result
+result = buffer2.mapRead()
+result = IntArrayType.from_buffer(result)  # cast
+print(result[:])
+
+# With Numpy this would be:
+# print(np.frombuffer(buffer2.mapRead(), np.int32))
```
Korijn: Can you add such a comment for the short version too?

Korijn:
I'm not too familiar with ctypes but it looks good!

dev-requirements.txt
```diff
@@ -5,4 +5,6 @@ requests
 setuptools
 wheel==0.31.1
 twine
+python-shader>=0.3.3
+numpy
```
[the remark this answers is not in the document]
almarklein: No. I also removed numpy from this list; we don't use it in the tests at this point.

examples/compute_noop.py
almarklein: done

PR #48, opened 2020-01-21
Get rid of asyncio from the examples. Add one asyncio+glfw example.

### LINKED ISSUE #21: Don't force users to set up an event loop
I was just thinking about this; can wgpu be decoupled from asyncio, allowing users to stick to the native event loops of the various GUI toolkits, if they so desire?

It seems like it would have a few advantages; 
* Examples would have less boilerplate and there would be less concepts newcomers need to deal with in order to get started.
* It might also make it more feasible for existing GUI software projects to start using WGPU since they may already have complicated event loop configurations going on.

Taking Qt as an example, we can have three examples where one just uses the Qt event loop, one uses the naive integration (which we have currently) and another one that shows more advanced event loop integration.

Thoughts?

almarklein:
Closes #21. Makes @Korijn 😄.

### DIFF
--- a/README.md
+++ b/README.md
@@ -74,8 +74,25 @@ import wgpu.backend.rs
 
 ## GUI integration
 
-To render to the screen you can use any of the following GUI toolkits:
-`Tk` (included with Python), `glfw`, `PySide2`, `PyQt5`, `PySide`, `PyQt4`.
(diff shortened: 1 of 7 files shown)


Korijn:
Awesome! I think this will surely help in adoption.

We can tout this in the README.md as well, "opt-in async interface"

PR #49, opened 2020-01-22
Release only for builds that were initiated by a tag

### LINKED ISSUE #42: Fix triple quote issue on azure pipelines
It's unclear why but there are additional quotes at the end of the TAG variable... sometimes? It's causing the condition to fail even when it should succeed. :/

```
Evaluating: and(succeeded(), eq(variables['Build.SourceBranchName'], variables['TAG']))
Expanded: and(True, eq('merge', 'v0.1.2-2-g166d188'''))
                                                  ^^
```

Korijn:
Closes #42


azure-pipelines.yml
```diff
@@ -170,30 +170,35 @@ jobs:
...
-      versionSpec: '3.6'
   - task: Bash@3
-    displayName: Check git tag
+    displayName: git tag == git branch
     inputs:
       targetType: inline
       script: |
-        set -x
-        TAG=$(git describe --tags)
-        echo "##vso[task.setvariable variable=TAG]$TAG"
```
Korijn: in hindsight, maybe it was the trailing double quote here, I dunno. it's no longer an issue anymore :)

PR #51, opened 2020-01-27
docker builds for linux32

Korijn:


Korijn:
:tada:

Screenshot from 2020-01-27 09-47-03

almarklein:
So this basically replicates the build job for Docker images, currently only running 32bit Linux on it?

The platform suffixes look so inconsistent ... Is it the wheel format that says that it's `win32` vs `amd_64` for Windows and `i686` vs `x86_64` for Linux?

Korijn:
> So this basically replicates the build job for Docker images, currently only running 32bit Linux on it?

That's right.

> The platform suffixes look so inconsistent ... Is it the wheel format that says that it's `win32` vs `amd_64` for Windows and `i686` vs `x86_64` for Linux?

Yeah, that's what PyPA has established to be valid wheel platform tags.
Docker references 32-bit architectures with i386.

PR #56, opened 2020-02-11
Ensure correct lifetime of sub-structs that are referenced as pointers

### LINKED ISSUE #52: Triangle example: white window that closes after 1 second (Windows 10)
Running the triangle example with GLFW on Windows 10 (up to date, i.e. running the last Fall update) gives me a white window that exits shortly afterwards.
Using the tkinter canvas exits immediately without even showing a window.
I'm running a Surface Book 2 which has an integrated Intel GPU as well as a dedicated Nvidia GeForce 1050. My graphics drivers are up to date, both Intel and Nvidia. I can compile and run OpenGL and Vulkan applications just fine, so hardware support should not be the issue. A rough guess would be that wgpu-py fails to select the adapter because two are available, but I can't say for sure as I don't know the internals of wgpu.

almarklein:
fixes #52

### DIFF
--- a/wgpu/backend/rs.py
+++ b/wgpu/backend/rs.py
@@ -30,6 +30,7 @@ Developer notes and tips:
 import os
 import sys
 import ctypes
+from weakref import WeakKeyDictionary
 
 from cffi import FFI, __version_info__ as cffi_version_info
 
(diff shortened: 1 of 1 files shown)


almarklein:
* I added a call to `gc.collect()` right before the wgpu api call.
* Comment the added line to `new_struct()`
* Run test script 5x: fails each time
* Uncomment the added line
* Run test script 5x: success each time

Korijn:
LGTM

PR #60, opened 2020-02-12
Convert API to snake_case

### LINKED ISSUE #57: Adopt snake_case instead of camelCase
This is a Python lib.

almarklein:
Closes #57 

* flags: name is CamelCase, field names are UPPERCASE, values are int.
* enums: name is CamelCase, field names are snake_case, values are 'like-this'.
* structs: their fields are the function arguments, so we never refer them by name.
* function args: name is snake_case. 
* function names: name is snake_case.

### DIFF
--- a/examples/triangle.py
+++ b/examples/triangle.py
@@ -105,12 +105,12 @@ def _main(canvas, device):
         alphaToCoverageEnabled=False,
     )
 
-    swap_chain = canvas.configureSwapChain(
+    swap_chain = canvas.configure_swap_chain(
         device, wgpu.TextureFormat.bgra8unorm_srgb, wgpu.TextureUsage.OUTPUT_ATTACHMENT
     )
 
     def drawFrame():
-        current_texture_view = swap_chain.getCurrentTextureView()
+        current_texture_view = swap_chain.get_current_texture_view()
         command_encoder = device.createCommandEncoder()
 
         render_pass = command_encoder.beginRenderPass(
--- a/examples/triangle.py
+++ b/examples/triangle.py
@@ -47,90 +47,92 @@ def fragment_shader(
 def main(canvas):
     """ Regular function to setup a viz on the given canvas.
     """
-    adapter = wgpu.requestAdapter(powerPreference="high-performance")
-    device = adapter.requestDevice(extensions=[], limits=wgpu.GPULimits())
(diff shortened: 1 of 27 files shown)


almarklein:
I'm going to prepare PR's in python-shader and visvis2 to adjust to the new API. Then I'll merge them in quick succession.

wgpu/gui/qt.py
```diff
@@ -37,25 +37,25 @@ def __init__(self, *args, size=None, title=None, **kwargs):
...
         # https://doc.qt.io/qt-5/qt.html#WidgetAttribute-enum  WA_PaintOnScreen
         return None
 
-    def paintEvent(self, event):
-        self._drawFrameAndPresent()
+    def paintEvent(self, event):  # noqa: N802 - this is a Qt method
+        self._draw_frame_and_present()
 
-    def getSizeAndPixelRatio(self):
+    def get_size_and_pixel_ratio(self):
```
Korijn: do you think perhaps the entire Qt widget should have Qt style conventions? eh, it's just a little straining on the eyes :) wouldn't mind merging this as is

examples/triangle.py
```diff
@@ -47,90 +47,92 @@ def fragment_shader(
...
+        vertex_stage={"module": vshader, "entry_point": "main"},
+        fragment_stage={"module": fshader, "entry_point": "main"},
+        primitive_topology=wgpu.PrimitiveTopology.triangle_list,
+        rasterization_state={
+            "front_face": wgpu.FrontFace.ccw,
+            "cull_mode": wgpu.CullMode.none,
+            "depth_bias": 0,
+            "depth_bias_slope_scale": 0.0,
+            "depth_bias_clamp": 0.0,
         },
```
almarklein: Perhaps this illustrates the changes best: 
* the function name is snake_case
* the direct arguments are snake_case
* the "sub-arguments" (keys of args that are dicts) are snake_case
* enums and flags are unchanged.

Korijn:
LGTM!

wgpu/gui/qt.py
almarklein: mmm, that's a good question. I think that since the methods that we add here are part of the wgpu API (they come from an abstract base class), they should be snake_case. It actually makes it clear with what API you're interacting with. But yeah, it does not look qt-ish :/

PR #61, opened 2020-02-19
PyInstaller runtime hook and test

Korijn:
Don't merge until we have approval and alignment with @bjones1 at https://github.com/pyinstaller/pyinstaller/pull/4662

### DIFF
--- a/dev-requirements.txt
+++ b/dev-requirements.txt
@@ -1,4 +1,4 @@
-black
+black==19.10b0
 pytest
 flake8
 pep8-naming
@@ -8,3 +8,4 @@ wheel
 twine
 python-shader>=0.3.4
 auditwheel; sys_platform == 'linux'
+pyinstaller[hook_testing] @ https://github.com/bjones1/pyinstaller/archive/4232a.zip
--- a/.gitignore
+++ b/.gitignore
@@ -27,6 +27,7 @@ parts/
 sdist/
 var/
 wheels/
+wheelhouse/
 *.egg-info/
 .installed.cfg
 *.egg
--- a/setup.py
+++ b/setup.py
@@ -48,10 +48,10 @@ setup(
     url="https://github.com/almarklein/wgpu-py",
     cmdclass={"bdist_wheel": bdist_wheel},
     data_files=[("", ["LICENSE"])],
-    entry_points = {
-        'pyinstaller40': [
-            'hook-dirs = wgpu.__pyinstaller:get_hook_dirs',
-            'tests = wgpu.__pyinstaller:get_test_dirs',
+    entry_points={
+        "pyinstaller40": [
+            "hook-dirs = wgpu.__pyinstaller:get_hook_dirs",
+            "tests = wgpu.__pyinstaller:get_test_dirs",
         ],
     },
 )
--- a/setup.py
+++ b/setup.py
@@ -48,4 +48,10 @@ setup(
     url="https://github.com/almarklein/wgpu-py",
     cmdclass={"bdist_wheel": bdist_wheel},
     data_files=[("", ["LICENSE"])],
+    entry_points = {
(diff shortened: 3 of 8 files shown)


Korijn:
@almarklein do you have any idea why we are seeing this output?

```
/tmp/clockuh5m3lmu.c:2:1: warning: return type defaults to ‘int’ [-Wimplicit-int]
 main (int argc, char **argv) {
 ^
```

dev-requirements.txt
```diff
@@ -1,4 +1,4 @@
-black
+black==19.10b0
 pytest
```
almarklein: ?

almarklein:
Odd. I do see `FileNotFoundError: [Errno 2] No such file or directory: 'wgpu.h'`  too. So maybe it loads the library without a h-file *somehow*?

dev-requirements.txt
Korijn: I use pipenv to install from the dev-requirements.txt file, and because black is still in "pre-release" it will refuse to install it. I can force pipenv by being specific about the version that I want. I can revert this change if you want

Korijn:
> Odd. I do see `FileNotFoundError: [Errno 2] No such file or directory: 'wgpu.h'` too. So maybe it loads the library without a h-file _somehow_?

I check; `resource_filename("wgpu.h")` just returns `"wgpu.h"` which is very odd.

> Running locally I get `No module named 'wgpu.resources'`.

Well, it isn't a module, is it? I wouldn't expect it to be importable. It's just a folder with some data and binary files.

almarklein:
> Well, it isn't a module, is it?

Doesn't `pkg_resources` want it to be a subpackage? I think that is why we put an `__init__.py` in that dir.

Korijn:
Figured it out thanks to your comment :)

almarklein:
Hooza!

bjones1:
I've reviewed this from a PyInstaller perspective and it looks good to me! Thanks for working on this. Next steps I can think of:

1. How soon will you release this on PyPI? If it's soon, I'd like to update the PyInstaller requirements to install your package from PyPI then run the bundled tests it contains. If it's not soon, I can update the requirements to install from this github repo until it's released.
2. Until https://github.com/pyinstaller/pyinstaller/pull/4662 is merged, main-line PyInstaller won't use this. Would you prefer to:

    1. Tell wgpu users that want to package their installation to install from the PR branch instead?
    2. I could merge the current hook in PyInstaller, then revert that after the merge.

Also, any feedback on the process / refinements to the docs you could suggest?

Korijn:
> How soon will you release this on PyPI?

We are releasing every few weeks at the moment, so probably next week.

> Until pyinstaller/pyinstaller#4662 is merged, main-line PyInstaller won't use this

We can wait a while; we're still in early days (think alpha versions). How long do you think it will take?

> Also, any feedback on the process / refinements to the docs you could suggest?

I like the proposed setup where we define a test and then you also run it, so that both you and us verify our pyinstaller hook is working, but I missed a conceptual explanation about that. It took me a while to understand who is supposed to run tests. Turns out it was _both of us_. :)

bjones1:
Would you let me know when this code is released? I'll then add it to the packages PyInstaller runs tests on.

I'm guessing a month to get this merged in PyInstaller. If it takes longer, we can pursue another approach.

Thanks re: the docs. I've updated them to include a better test description at https://pyinstaller-sample-hook.readthedocs.io/en/latest/. Does this help?

Korijn:
Yes, that does help! Nice improvement.

We are fine waiting a month. :)

I will go ahead and merge this, will notify you on the other PR once a release is available!

bjones1:
Great! I appreciate your willingness to try this out and your feedback. I look forward to the release!

PR #62, opened 2020-02-19
WIP: Progress

almarklein:
* [x] Index buffers


wgpu/backend/rs.py
```diff
@@ -863,18 +863,25 @@ def draw(self, vertex_count, instance_count, first_vertex, first_instance):
         )
 
     # wgpu.help('Buffer', 'Size64', 'renderencoderbasedrawindirect', dev=True)
-    def draw_indirect(self, indirect_buffer, indirect_offset):
-        raise NotImplementedError()
+    # def draw_indirect(self, indirect_buffer, indirect_offset):
+    #     raise NotImplementedError()
```
Korijn: what's that? draw indirect?

almarklein: Something like [this](https://www.khronos.org/opengl/wiki/GLAPI/glDrawArraysIndirect) I assume.

PR #64, opened 2020-03-05
Rename backend -> backends

almarklein:
Seems to make more sense (as in https://github.com/almarklein/visvis2/issues/16).

I think the `gui` subpackage can remain as-is. `guis` does not make sense, and `gui_toolkits` is a bit long. @Korijn what do you think?


Korijn:
Sure :) I like consistency but "guis" seems odd.

PR #65, opened 2020-03-10
Add support for creating wgpu surface on macOS

### LINKED ISSUE #29: Support for macOS
In #27 we implemented support for all platforms, but we have not yet been able to verify this by actually running an example on macOS. Need at least High Sierra 10.13 to have Metal.

Checks:
* [ ] OS X tk
* [ ] OS X qt
* [ ] OS X glfw

berendkleinhaneveld:
Here is the work we did on calling objective-c methods with ctypes, to replicate the example code from wgpu-native. Works for Qt and GLFW.

Closes #29

### DIFF
--- a/wgpu/backends/rs.py
+++ b/wgpu/backends/rs.py
@@ -30,6 +30,7 @@ Developer notes and tips:
 import os
 import sys
 import ctypes
+import ctypes.util
 from weakref import WeakKeyDictionary
 
 from cffi import FFI, __version_info__ as cffi_version_info
@@ -138,7 +139,6 @@ def get_surface_id_from_canvas(canvas):
         return _lib.wgpu_create_surface_from_windows_hwnd(hinstance, hwnd)
     elif sys.platform.startswith("darwin"):
         # wgpu_create_surface_from_metal_layer(void *layer)
-        # todo: MacOS support
         # This is what the triangle example from wgpu-native does:
         # #if WGPU_TARGET == WGPU_TARGET_MACOS
         #     {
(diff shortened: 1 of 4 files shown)


Korijn:
Cffi is already in runtime deps, setup.py

berendkleinhaneveld:
Aha, this: `python setup.py develop` was what I was missing 😃

Korijn:
I guessed as much :) it's in the README under "development" , so I think we're fine.

almarklein:
But Berend's workflow of simply `pip install -r requirements.txt` and then using `PYTHONPATH` to have Python find the lib is fine too, and adding the dep there makes that easier.

wgpu/backends/rs.py
```diff
@@ -149,8 +149,47 @@ def get_surface_id_from_canvas(canvas):
...
+        ):
+            # NSWindow instances respond to contentView selector
+            content_view = objc.objc_msgSend(window, content_view_sel)
+        elif objc.objc_msgSend(window, responds_to_sel_sel, ctypes.c_void_p(layer_sel)):
+            # NSView instances respond to layer selector
+            # Let's assume that the given window pointer is actually the content view
+            content_view = window
+        else:
+            # Can't tell what kind of object window is
+            raise TypeError()
```
almarklein: Let's do something like `raise TypeError("bla bla fail")`

almarklein: Actually make it a `RuntimeError`; I think that's more appropriate from Python's point of view.

berendkleinhaneveld: Changed the TypeError to RuntimeError, and I've added a message. Do you think this is enough info?

Korijn:
Yes, but it also means we have to keep requirements in sync between the two files.

We could also add ` -e .` to dev-requirements.txt but that approach is not reversible (installing a package in editable mode is only reversible through setup.py develop, not through pip install -e :()

wgpu/backends/rs.py
almarklein: Perfect!

almarklein:
Ok, fair enough :)

almarklein:
@berendkleinhaneveld BTW: I confirmed that on glfw the paint event is indeed fired on resize.

Korijn:
Merge?

Issue #63, opened 2020-02-24
Canvas method to request a new draw


almarklein:
Having a (cross-gui-toolkit) method that can be used to request a new draw would be nice. When using qt `canvas.update()`  does the trick, but a more general solution would be favorable, plus we can add some niceties:

- calling `window.requestAnimationFrame` schedules a draw call indirectly (non-blocking)
- animation is paused when the canvas is not visible (scheduled calls are not executed until canvas becomes visible again)
- refresh rate of the GUI environment is matched if possible
- callback is passed the time delta since the last call
- you have to request another draw at the end of your animation function; this avoids scheduling too many calls if your drawing happens to be slower than the time budget allows for

Korijn:
I'm wondering if we should just leave this up to the application developer by providing an API that allows them to do it.

Korijn:
Nice work on the docs! The GUI API page's examples left me a little confused though; where are the calls to `request_draw`?

Looking at the Qt canvas a little better, it appears that Qt is making the calls, rather than the python code. I guess my expectations were somewhat different here.

I understand that Qt will trigger a draw when the window is resized for example, or opened for the first time. Does Qt also continuously trigger drawing when there is no interaction, or does the user then need to use `request_draw` to achieve that?

almarklein:
Good question. The GUI toolkit will indeed request a new draw when it *knows* that the visualization needs to be refreshed. Like when the window is resized, and om some systems when part of the window was hidden by another window and is now shown again. This was already the case before #78. What was added was the possibility for the user to request a draw. E.g. in cases that the visualization itself has changed.

This works somewhat different for Qt and glfw. In Qt the former already works automatically, and requesting a draw is done simply by calling `update()`. In both cases this will at some point result in a draw event.

For glfw you have to bring your own event loop. So this draw event ... we have to do that ourselves. The simplest way I could find to realize this is by adding a single function (`update_glfw_canvasses`) that you call on each iteration, and which draws any windows that have requested (manually or via e.g. a resize) to be redrawn.

Korijn:
Ok, great, that's perfect!

> Qt the former already works automatically, and requesting a draw is done simply by calling `update()`. In both cases this will at some point result in a draw event.

Just to make sure I fully understand the details here: so Qt does _not_ continuously call `update()` when only the visualization changes?

almarklein:
Haha, now I'm not sure if I understand your question correctly :)

[...]

PR #68, opened 2020-03-12
Hidpi support

### LINKED ISSUE #69: Drop Tk support?
* On Linux: seems impossible to make it work.
* On macOS: Dito. 
* On Windows: works, but painting occurs *after* the actual paint event, so there is flicker.
* No high-res support.

The only advantage is that Tk is always there (unless you have a mac). But `glfw` does a great job at providing a lean GUI that is very easy to install (binaries also ship on Linux since a few weeks).

Is it worth keeping the backend around only for a shitty solution on Windows?

almarklein:
## The problem

I've got a shiny new screen on my workplace, but now all examples fail :) Time to fix highres display support.

## Diagnosis

Some notes so far:
* If the size of the render texture does not exactly match the size of the widget, you get a black screen.
* In glfw, the widget height and weight are expressed in physical size. All good!
* In Qt, the widget height/weight are in logical pixels, but seem to be rounded. I have yet to figure out how to get the physical size. 
* In Tk ... *sigh* there is no concept of logical pixels. Everything works but is tiny on highres displays. 

Observations specific to Qt:
* The triangle is initially not shown. When resizing the window, it flickers. After a drag, its sometimes visible, sometimes not.
* Setting `canvas._swap_chain._surface_size = ()`, to invoke the creation of a new native swapchain, does not help.
* Using `canvas.grab().size()` you can get the size in physical pixels. This method is slow though!
* I can see a triangle and resize the window a real tiny bit such that the logical size (`canvas.size()`) does not change, but the widget becomes black. Surprisingly, the `canvas.grab().size()` is also not changed. (!!) There is also no resize event if the logical size does not change :(
* The example in visvis2/radiant that embeds a scene in a Qt app works fine.
* --> Fixed it by using a subwidget to provide the canvas.

## Solution

* Redesigned the API for getting logical/physical size and pixelratio for the canvas objects.
* Added Canvas API to set the size and close the window.
* Refactored qt gui code for hidpi support. Only limitation that the system pixelratio is rounded.
* Tweaks to glfw to improve hidpi support.
* Removed tk support, closes #69.


Korijn:
Getting hidpi to work with Qt on Windows is quite a trip. We figured it out for one of our internal Qt apps.

First, you need to let Windows know that your process is DPI aware:

```python
import ctypes
try:
    ctypes.windll.shcore.SetProcessDpiAwareness(True)
except:
    pass  # not windows
```

Then you can configure Qt:

```python
QApplication.setAttribute(Qt.AA_EnableHighDpiScaling, True)
QCoreApplication.setAttribute(Qt.AA_UseHighDpiPixmaps, True)
```

And we already merged the method to get the devicePixelRatio for the screen the window is currently on:

```python
class QtWgpuCanvas(BaseCanvas, QWidget):
    # ...
    def get_size_and_pixel_ratio(self):
        return self.width(), self.height(), self.window().devicePixelRatio()
```

almarklein:
@Korijn interesting that `devicePixelRatio` was already non-unit. Will look into it.

I applied the calls you mentioned earlier in Pyzo. I just learned that it should be `ctypes.windll.shcore.SetProcessDpiAwareness(2)`, see https://github.com/pyzo/pyzo/pull/696 for details. You may want to update your code as well.

almarklein:
> interesting that `devicePixelRatio` was already non-unit. 

This was because I was running from Pyzo (which is also Qt), and the interpreter process inherits some stuff, apparently.

Korijn:
> > interesting that `devicePixelRatio` was already non-unit.
> 
> This was because I was running from Pyzo (which is also Qt), and the interpreter process inherits some stuff, apparently.

Yeah. Processes are a tree.

almarklein:
Ready for review. What happened to CI? Did it stop running because this was a draft?

Korijn:
That's right. You should be able to trigger a build manually for this PR in azure though!

wgpu/gui/glfw.py
```diff
@@ -58,6 +58,15 @@
         )
 
 
+# Enable high-res displays
+# (otherwise glfw does not pick up being moved from one monitor to another)
+# todo: see if we can get this added to glfw
+try:
+    ctypes.windll.shcore.SetProcessDpiAwareness(2)
+except Exception:
+    pass  # fail on non-windows
```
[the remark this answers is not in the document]
almarklein: Also see https://github.com/FlorianRhiem/pyGLFW/pull/48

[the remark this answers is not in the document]
Korijn: Cool to see the quick response and cooperation!

I like how they put it inside an init() func instead of on import :)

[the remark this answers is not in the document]
almarklein: Code is removed. Instead we listen to the content-scale callback to implement smooth transitioning from one monitor to another.

almarklein:
Also tested on Linux.

wgpu/backends/rs.py
```diff
@@ -978,27 +978,27 @@ def __init__(self, device, canvas, format, usage):
...
+        # print(psize, self._canvas.get_logical_size(), self._canvas.get_pixel_ratio())
 
         struct = new_struct(
             "WGPUSwapChainDescriptor *",
             usage=self._usage,
             format=self._format,
-            width=cur_size[0],
-            height=cur_size[1],
+            width=max(1, psize[0]),
+            height=max(1, psize[1]),
```
Korijn: I could imagine that magically correcting might hide user errors, where the user might rather want to know that their application is setting ridiculous width/height values?

wgpu/gui/glfw.py
```diff
@@ -62,13 +34,17 @@ class GlfwWgpuCanvas(BaseCanvas):
...
         super().__init__()
+
+        # Handle inputs
         if size:
-            width, height = size
+            size = float(size[0]), float(size[1])
         else:
-            width, height = 256, 256
-        title = title or ""
+            size = 640, 480
```
Korijn: these aren't floats, so why do you explicitly cast to floats in line 44?

wgpu/gui/glfw.py
```diff
@@ -79,31 +55,71 @@ def __init__(self, *, size=None, title=None):
...
 
-        self._window = glfw.create_window(width, height, title, None, None)
-        glfw.set_window_refresh_callback(self._window, self._paint)
+        # Create the window (the initial size may not be in logical pixels)
+        self._window = glfw.create_window(int(size[0]), int(size[1]), title, None, None)
+
+        # Register callbacks
+        glfw.set_window_content_scale_callback(self._window, self._on_pixelratio_change)
+        glfw.set_framebuffer_size_callback(self._window, self._on_size_change)
+        glfw.set_window_refresh_callback(self._window, self._on_paint)
```
Korijn: I have some concerns here; you can't guarantee that the application can keep up with the desired refresh rate. But either way, it's probably fine for now

wgpu/gui/glfw.py
```diff
@@ -79,31 +55,71 @@ def __init__(self, *, size=None, title=None):
...
+        self._logical_size = self._get_logical_size()
+        self._on_paint()
+
+    def _on_paint(self, *args):
+        self._draw_frame_and_present()
+
+    # Helpers
+
+    def _get_logical_size(self):
+        # Because the value of get_window_size is ambiguous, we use the
```
Korijn: ambiguous how? can you add that to the comment?

wgpu/gui/qt.py
```diff
@@ -13,50 +14,109 @@
...
+# Enable high-res displays
+try:
+    ctypes.windll.shcore.SetProcessDpiAwareness(2)
+except Exception:
+    pass  # fail on non-windows
+try:
+    QtWidgets.QApplication.setAttribute(QtCore.Qt.AA_EnableHighDpiScaling, True)
+    QtCore.QCoreApplication.setAttribute(QtCore.Qt.AA_UseHighDpiPixmaps, True)
+except Exception:
+    pass  # fail on older Qt's
```
Korijn: I feel like this code should be in a method.. Qt app authors probably want control over these attributes themselves.

wgpu/gui/qt.py
```diff
@@ -13,50 +14,109 @@
...
+        lsize = self._subwidget.width(), self._subwidget.height()
+        return float(lsize[0]), float(lsize[1])
+
+    def get_physical_size(self):
+        # https://doc.qt.io/qt-5/qpaintdevice.html
+        # https://doc.qt.io/qt-5/highdpi.html
+        lsize = self._subwidget.width(), self._subwidget.height()
+        lsize = float(lsize[0]), float(lsize[1])
+        ratio = self._subwidget.devicePixelRatioF()
+        return int(lsize[0] * ratio + 0.4999), int(lsize[1] * ratio + 0.4999)
```
Korijn: is this a rounding trick?

wgpu/gui/qt.py
```diff
@@ -13,50 +14,109 @@
...
+
+    def close(self):
+        super().close()
+
+    def is_closed(self):
+        return not self.isVisible()
+
+
+class WgpuSubWidget(QtWidgets.QWidget):
+    """ The widget that actually prevents the surface to render to.
```
Korijn: provides*

wgpu/backends/rs.py
almarklein: The reason for adding this was to avoid wgpu errors when the window is resized to zero height/width. If we want to catch invalid user values (`<0`) we should catch these earlier.

almarklein: Added a check in the gui code :)

wgpu/gui/glfw.py
almarklein: yeah, there was a bit of historic stuff there. The cast to float functioned as a typecheck, but this is (in the current code) already done a bit below. I'm cleaning this up.

wgpu/gui/glfw.py
almarklein: Yeah, let's revisit when we work on #63

wgpu/gui/glfw.py
almarklein: updated.

wgpu/gui/qt.py
almarklein: Removed `AA_UseHighDpiPixmaps` as it is notr strictly needed for wgpu.

wgpu/gui/qt.py
almarklein: lol

wgpu/gui/qt.py
almarklein: Yeah. What was I thinking? Changed to use `round()`.

PR #70, opened 2020-03-16
Logging

### LINKED ISSUE #16: BaseCanvas._draw_frame_and_present swallowing errors
Currently the `except` in `BaseCanvas._drawFrameAndPresent` swallows errors and writes to stderr.

I think users will want to bring their own implementation for logging, and I also think some users will prefer to see errors bubble up to the top of the stack (where they may be logged or handled centrally for example) rather than be silently swallowed.

In short, the post-mortem debugging implementation probably has its use in development, but maybe it should be behind a flag?

almarklein:
Closes #16

Adds a logger with a custom formatter that enables PM debugging and shortens exceptions when they occur repeatedly (which happens easily with errors that occur in a draw). The logic to do that is removed from the draw function.

### DIFF
--- a/wgpu/_coreutils.py
+++ b/wgpu/_coreutils.py
@@ -3,8 +3,6 @@ Core utilities that are loaded into the root namespace or used internally.
 """
 
 import os
-import sys
-import logging
 import inspect
 from pkg_resources import resource_filename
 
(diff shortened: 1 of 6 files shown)


almarklein:
mmm, had to manually start Azure again. What up?

wgpu/_coreutils.py
```diff
@@ -222,3 +224,47 @@ def help(*searches, dev=False):
...
+            return super().format(record)
+
+
+def create_logger():
+    logger = logging.getLogger("wgpu")
+    logger.propagate = False
+    logger.setLevel(logging.WARNING)
+
+    handler = logging.StreamHandler()
+    logger.addHandler(handler)
```
Korijn: https://docs.python.org/3/howto/logging.html#configuring-logging-for-a-library

In particular the note is relevant here

almarklein: Removed configuration of the logger. I realized that the storing of tb info on `sys.last_xx` is not needed if the `exc_info` argument is used. The shortening of recurring messages can be done where it was done before.

wgpu/gui/base.py
```diff
@@ -34,24 +36,38 @@ def _draw_frame_and_present(self):
...
+    def _log_exception(self, kind, err):
+        """ Log the given exception instance, but only log a one-liner for
+        subsequent occurances of the same error to avoid spamming (which
+        can happen easily with errors in the drawing code).
+        """
+        msg = str(err)
+        msgh = hash(msg)
+        if msgh not in self._err_hashes:
+            # Provide the exception, so the default logger prints a stacktrace.
+            # IDE's can get the exception from the root logger for PM debugging.
```
Korijn: I'm happy to see that PM debugging still works as well :)

Korijn:
LGTM (except for the build status)

PR #71, opened 2020-03-21
WIP: Progress

almarklein:
* [x] API support for textures.
* [x] Tooling to run tests that render to a texture.
* [x] Tests that do storage with textures
* [x] Tests that render a textures geometry

### DIFF
--- a/wgpu/base.py
+++ b/wgpu/base.py
@@ -64,14 +64,6 @@ class GPUObject:
         return self._label
 
 
-class DictLike:
-    def __getitem__(self, name):
-        return self.__dict__[name]
-
-    def get(self, key, default=None):
-        return self.__dict__.get(key, default)
-
-
 class GPUAdapter:  # Not a GPUObject
     """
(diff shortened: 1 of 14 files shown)


wgpu/base.py
```diff
@@ -63,13 +63,9 @@ def label(self):
...
-
-class DictLike:
-    def __getitem__(self, name):
-        return self.__dict__[name]
-
-    def get(self, key, default=None):
-        return self.__dict__.get(key, default)
+    def __del__(self):
+        if hasattr(self, "destroy"):
+            self.destroy()
```
Korijn: I thought this was only invoked when using the `del` keyword, but apparently also by the garbage collector? That's good to know!

almarklein: Yes! The only thing is that there is no way to know *when* it gets called, though for variables in a function scope, not referenced from elsewhere, *when* is usually directly when the function returns.

almarklein:
Hoozah!

Issue #10, opened 2019-12-24
Implement the full API


almarklein:

almarklein:
Support for compute shaders added in #43

almarklein:
Closes by  #74

PR #74, opened 2020-03-28
Improve api coverage

### LINKED ISSUE #12: Tests ... we need tests

almarklein:
This brings us so close to full coverage that I consider this closing #10. Also closes #12. Now uncovered features are:

* debug markers (because I have no idea yet how these work)
* render bundles (because wgpu-native does not support them (yet))
* indirect drawing
* ~a few missing features inside render calls: depth_stencil_state, resolve_target, depth_stencil_attachment~ implemented and covered with tests!

todo:
* [x] test coverage!


Korijn:
I dont see debug markers in the webgpu spec... Can you provide a link/reference so I can read up on them?

Korijn:
Would it be much effort to add the indirect drawing API? It seems like it would be necessary to support instanced rendering?

almarklein:
> I dont see debug markers in the webgpu spec...

E.g. [here](https://gpuweb.github.io/gpuweb/#dom-gpucommandencoder-insertdebugmarker). Not much info though. More [here](https://www.khronos.org/opengl/wiki/Debugging_Tools).

> Would it be much effort to add the indirect drawing API?

I'll have a look. I initially planned to just implement all features that we might need in visvis2. But yeah ... we're pretty close to 100% now so maybe  just push through.

> It seems like it would be necessary to support instanced rendering?

I don't think so.

Korijn:
https://www.khronos.org/opengl/wiki/Debug_Output

This page describes use of debug markers in more detail; apparently you can annotate your GPU commands with "markers" and have the annotations show up in traces/logs, to help you identify/locate things.

Korijn:
> > It seems like it would be necessary to support instanced rendering?
> 
> I don't think so.

OK then. Nevermind :)

almarklein:
afbeelding

almarklein:
I'm merging this. I will have a go at debug markers and indirect drawing in new PR's.

PR #75, opened 2020-04-01
Implement indirect drawing

almarklein:
These new tests actually fail in the lib shipped with the library, because the `BufferUsage.INDIRECT` enum was [changed at some point](https://github.com/gfx-rs/wgpu/pull/468/files). It looks like
the .h file that we use does not precisely match the library. Eek!

Anyway, let's just merge this, and update to the latest wgpu-native.


Korijn:
LGTM

PR #77, opened 2020-04-02
Update to latest webgpu spec and last release of wgpu-native

almarklein:
It looks like the WebGPU spec is a bit ahead, maybe because the latest release from wgpu-native is quite old (feb 20). This results in a bit of asymmetry, but it's not too bad.


almarklein:
I've heard there should be a new release of wgpu-native shortly, so we'll probably have another PR like this soon, withe the API's more aligned.

PR #78, opened 2020-04-03
Implement request_draw

### LINKED ISSUE #63: Canvas method to request a new draw

almarklein:
Closes #63 

* For qt this is easy, because it's an alias for `update()`.
* For glfw it's more involved, because glfw does not provide an event loop. So I've made a function that one must call from your glfw mainloop, which will redraw all canvases that are invalidated.
* Updated the glfw tests (they check that multiple events or draw requests result in a single draw)
* The glfw example has a more sophisticated event loop that does not turn your laptop into a heater,

### DIFF
--- a/wgpu/gui/flexx.py
+++ b/wgpu/gui/flexx.py
@@ -24,5 +24,8 @@ class WgpuCanvas(flx.CanvasWidget):
         window.requestAnimationFrame(self._draw_frame_and_present)
         self.draw_frame()
 
+    def request_draw(self):
+        pass  # already drawing continuously
+
     def is_closed(self):
         return False
--- a/wgpu/gui/flexx.py
+++ b/wgpu/gui/flexx.py
@@ -2,6 +2,9 @@ from flexx import flx
 from pscript.stubs import window
 
 
+raise NotImplementedError()
+
+
 # Flexx cannot do multiple inheritance, so we consider BaseCanvas an interface :)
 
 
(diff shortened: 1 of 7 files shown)


wgpu/gui/qt.py
```diff
@@ -26,7 +26,9 @@ def enable_hidpi():
     """ Enable high-res displays.
     """
     try:
-        ctypes.windll.shcore.SetProcessDpiAwareness(2)
+        # See https://github.com/pyzo/pyzo/pull/700 why we seem to need both
+        ctypes.windll.shcore.SetProcessDpiAwareness(1)  # global dpi aware
+        ctypes.windll.shcore.SetProcessDpiAwareness(2)  # per-monitor dpi aware
```
almarklein, line 31: ping @Korijn

wgpu/gui/flexx.py
```diff
@@ -24,5 +24,8 @@ def _draw_frame_and_present(self):
         window.requestAnimationFrame(self._draw_frame_and_present)
         self.draw_frame()
 
+    def request_draw(self):
+        pass  # already drawing continuously
```
Korijn, line 28: shouldn't this be a call to window.requestAnimationFrame?

wgpu/gui/qt.py
Korijn: Interesting :D if it works... :)

Korijn:
LGTM

wgpu/gui/flexx.py
almarklein: Fixed

Issue #73, opened 2020-03-28
Support for grayscale textures


almarklein:
Currently, textures of format `r8unorm` and `r8uint` work, but `r16sint`, `r32sint` and `r32float` fail. I've not yet been able to find out why.

Korijn:
Are these formats known to work in wgpu-native upstream?

almarklein:
When I asked last week or so, they responded to not have trouble with them earlier. But then again, it does work with uint8 textures. At the time I did not have the tests to say what exactly worked and what not. Let's wait for their next release and if it then still fails I'll ask for help.

almarklein:
https://github.com/gfx-rs/wgpu/issues/575

almarklein:
I tweaked a Rust example to use R16Sint and R32Sint textures and that seemed to work. Should compare that example's wgpu calls, and compare the SpirV that comes off that fragment shader with our SpirV ...

PR #81, opened 2020-04-10
Update wgpu-native to v0.5.1

almarklein:


almarklein:
This is nice. The API's line-up so much better now :)

Also still
afbeelding

almarklein:
I secretly hoped this would fix #73. It does not.

PR #83, opened 2020-04-12
Allow passing canvas to request_adapter

almarklein:
The `request_adapter` in wgpu-native now supports passing the surface id, so that we can make sure that the returned adapter is actually compatible with that (or such type of) surface.

I considered a few options to express this in the API, and this seems to be the most straightforward, even though it makes the API deviate from WebGPU somewhat (because in WebGPU the browser knows the surface id, so the user does not need pass it). Made it a required argument, so that it's obvious when it's set to None.

It's quite liberating that in an API that's low-level, we can afford to be a bit verbose at times :)

### DIFF
--- a/codegen-script.py
+++ b/codegen-script.py
@@ -359,15 +359,13 @@ for fname in ("base.py", "backends/rs.py"):
                 py_args = [field.py_arg() for field in arg_struct.values()]
                 if py_args[0] == "label: str":
                     py_args[0] = 'label=""'
-                if "requestadapter" in func_id:
-                    py_args = ["*"] + py_args
-                else:
                     py_args = ["self", "*"] + py_args
             else:
                 py_args = ["self"] + argnames
 
             # Replace function signature
-            api_lines[i] = preamble + ", ".join(py_args) + "):"
+            if "requestadapter" not in func_id:
+                api_lines[i] = preamble + ", ".join(py_args) + "):"
 
             # Insert comments
             if fname == "base.py":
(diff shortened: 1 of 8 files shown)


Korijn:
Is it still possible for people to "roll their own" GUI layer? Meaning, without using one of the wgpu-py provided canvases?

almarklein:
Yes, but they'd have to implement some abstract methods of `WgpuCanvasBase`, so that we can query the physical size, and obtain the surface id. As far as I can tell these are `get_physical_size`, `get_window_id`, and `get_display_id` (on Linux).

We could document and test this requirement. Now, or when needed :)

Korijn:
I mean, without using WgpuCanvasBase (or anything else from wgpu.gui)?

almarklein:
Well, there is some interplay between wgpu and the gui. This occurs on 2 points:

* When requesting an adapter you can pass the surface id so that the adapter is compatible. We've changed that to passing the canvas. It can be any object as long as it implements `get_window_id` (and `get_display_id` on Linux).
* At some point you need a swap chain. And you need to call into wgpu-native for this. So here it would be far easier to follow our "canvas interface". But it's duck-typed, so I don't think that's too big a deal.

Our current `WgpuCanvasBase` is a mix of that canvas interface required to interact with the wgpu-api, and some convenience functions so that all canvases work the same. We should probably split these, to make it easier for ppl to hook into the API with minimal conformance to our GUI stuff.

PR #84, opened 2020-04-12
Debug markers

almarklein:
This adds support for debug groups and markers. I haven't been able to actually see them in RenderDoc, but that is likely because I am missing something else. The C-api of these calls don't leave much room to do something wrong. So we might need to update the guide on debugging when we know a bit more about this.


almarklein:
This would still be nice: https://github.com/Korijn/wgpu-bin/issues/6

almarklein:
>  I haven't been able to actually see them in RenderDoc, but that is likely because I am missing something else.

From the chat: wgpu-native and gfx-hal don't tell each-other about them yet.

PR #85, opened 2020-04-14
Fixes for Linux

almarklein:
* With these, I can run all tests on Linux, with my NVidia drivers.
* With Intel drivers, all tests work except the `test_compute.py`.
* With Intel drivers the 3 disabled texture render tests actually work!!

### DIFF
--- a/tests/test_gui_glfw.py
+++ b/tests/test_gui_glfw.py
@@ -4,6 +4,7 @@ like the swap chain.
 """
 
 import os
+import time
 
 from python_shader import python2shader, vec4, i32
 from python_shader import RES_INPUT, RES_OUTPUT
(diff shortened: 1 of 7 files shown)


Korijn:
> * With Intel drivers the 3 disabled texture render tests actually work!!

What :rofl:

tests/test_gui_glfw.py
```diff
@@ -185,6 +186,7 @@ def draw_frame():
     canvas.set_logical_size(300, 200)
     canvas.set_logical_size(400, 300)
     for i in range(5):
+        time.sleep(0.01)
```
Korijn, line 189: oh boy, that looks bad

almarklein: Yeah, well, this gives other threads/processes some slack. Apparently it's needed. It may be that `sleep(0)` (which introduces a thread-switch) might already have worked, but I did not try that.

PR #88, opened 2020-04-15
Support wgpu-native debug builds

Korijn:
To support https://github.com/Korijn/wgpu-bin/pull/8


Korijn:
Debug binaries are now available from v0.5.1 onward: https://github.com/Korijn/wgpu-bin/releases

PR #89, opened 2020-04-15
Fix grayscale texture render tests

### LINKED ISSUE #73: Support for grayscale textures

almarklein:
Closes #73 

So the problem appeared to be the `compare` arg in `create_sampler`. It apparently must be set to "undefined". Except ... there is no `wgpu.CompareFunction.undefined`. Fortunately, we can just set it to zero for now.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,6 +10,10 @@ __pycache__/
 *.py[cod]
 *$py.class
 
+# Renderdoc captures
+*.cap
+*.rdc
+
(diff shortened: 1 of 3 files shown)


Korijn:
Oh well :) LGTM

PR #91, opened 2020-04-15
More defaults

### LINKED ISSUE #90: Update API for optional/nullable values
In IDL, some values that are optional do not have a default value. We should probably express these in Python as having a default value of None.

This is partly the cause of #73

almarklein:
Closes #90 

* Changed the parsing of IDL to take care of default values. 
* Document the default values in `base.py`
* Implement the default values of sub-arguments (dicts) in `rs.py`
* The most notable api change is that we no longer have `texture.create_default_view()`. Instead you call `texture.create_view()` without arguments.


Korijn:
LGTM!

PR #93, opened 2020-04-16
query wgpu version

### LINKED ISSUE #82: Query version of wgpu-native?
We know the version of our own lib, but if someone uses their own lib, it would be nice to be validate that the versions match.

Korijn:
Closes #82


almarklein:
Idea: put these info in the `__init__.py` instead of tiny loose files, to make access this metadata more trivial.

Korijn:
But it only applies to the rs backend right ?

almarklein:
Good point, but can still be a .py file that can be imported.

Korijn:
True! I'll change it

almarklein:
Only if it's indeed easier! (I haven't given it much thought yet myself)

Korijn:
It's pretty simple and it'll avoid runtime IO so I'm convinced that it will be an improvement :)

Korijn:
Done!

wgpu/backends/rs.py
```diff
@@ -47,6 +47,11 @@
 
 logger = logging.getLogger("wgpu")  # noqa
 
+# wgpu-native version
+__version__ = "0.5.1"
+__commit_sha__ = "83d4981"
+version_info = tuple(map(int, __version__.split(".")))
```
almarklein, line 53: Nice, to put it here :)

download-wgpu-native.py
```diff
@@ -86,7 +95,14 @@ def main(version, os_string, arch, upstream, build):
...
     if version != current_version:
         print(f"Version changed, updating {VERSION_FILE}")
-        write_current_version(version)
+        filename = "commit-sha"
+        url = f"https://github.com/{upstream}/releases/download/v{version}/{filename}"
+        commit_sha_filename = os.path.join(tmp, filename)
+        print(f"Downloading {url} to {commit_sha_filename}")
+        download_file(url, commit_sha_filename)
+        with open(commit_sha_filename) as fh:
+            commit_sha = fh.read().strip()
```
almarklein, line 104: Or `commit_sha = requests.get(url).read().decode().strip()`

Korijn: Mmm shorter indeed. Will keep that in mind for next time :)

PR #96, opened 2020-04-23
Getting version and logging messages from wgpu-native 

almarklein:
Will need to wait for a new version/build of wgpu-native.

### DIFF
--- a/wgpu/backends/rs.py
+++ b/wgpu/backends/rs.py
@@ -41,7 +41,7 @@ from cffi import FFI, __version_info__ as cffi_version_info
 
 from .. import base, flags, _structs
 from .. import _register_backend
-from .._coreutils import get_resource_filename
+from .._coreutils import get_resource_filename, logger_set_level_callbacks
 from .._mappings import cstructfield2enum, enummap
 
 
@@ -55,6 +55,9 @@ version_info = tuple(map(int, __version__.split(".")))
 
 if cffi_version_info < (1, 10):  # no-cover
     raise ImportError(f"{__name__} needs cffi 1.10 or later.")
+
+
+# %% Load the lib and connect logging system
 
 
 def _get_wgpu_h():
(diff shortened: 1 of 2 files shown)


wgpu/backends/rs.py
```diff
@@ -115,6 +115,33 @@ def _get_wgpu_lib_path():
...
+
+    func = m.get(level, logger.warn)
+    func(msg)
+
+
+# Set log callback. Levels -> 1: error, 2: warn, 3: info, 4: debug
+_lib.wgpu_set_log_callback(_logger_callback)
+_lib.wgpu_set_log_level(2)
+logger.setLevel(logging.WARN)
+# todo: somehow keep log levels in sync. At least automate log-level mapping
```
Korijn, line 144: idea: a function that sets both levels at once? would also make it possible for users to set a different level.

Korijn:
Awesome!

wgpu/backends/rs.py
almarklein: Yes, I think that makes the most sense. We can do a separate function. Or we can subclass the logger class and re-implement `setLevel()` to also call `wgpu_set_log_level`. The latter is a bit more complex, but probably cleaner.

Korijn: When subclassing, users can still run `logger.getLogger("wgpu"); logger.setLevel(...)` and they wouldn't get the subclass, right? I'm not sure if there's an airtight solution?

almarklein: Yes, I think that `getLogger()` would return the subclass. So that would make it trivial to keep it in sync, right? 

> [...] what if you just set the rust logger to the lowest level [...] and only filter with the python logger level?

That would cause a lot of overhead. First for the Rust logging system, and then sending them over to Python (turning strings into zero-terminated char arrays) ...

Korijn: > Yes, I think that `getLogger()` would return the subclass. So that would make it trivial to keep it in sync, right?
> 
> > [...] what if you just set the rust logger to the lowest level [...] and only filter with the python logger level?
> 
> That would cause a lot of overhead. First for the Rust logging system, and then sending them over to Python (turning strings into zero-terminated char arrays) ...

Really now! Awesome.

almarklein:
```py
import logging

class MyLogger(logging.Logger):
    def setLevel(self, level):
        super().setLevel(level)
        print("Setting log level to ", level)

original_cls = logging.getLoggerClass()
logging.setLoggerClass(MyLogger)
logging.getLogger("wgpu")
logging.setLoggerClass(original_cls)

logger = logging.getLogger("wgpu")
logger.setLevel(logging.INFO)
```

```
>>> (executing cell "" (line 39 of "tmp.py"))
Setting log level to  20
```

Korijn: `setLoggerClass`, I never knew about that part of the logging API. will surely use this in the future

almarklein:
woppa!

PR #97, opened 2020-04-29
Prevent premature cleanup

almarklein:
I saw some odd behavior, which after some searching seemed to be due to the memory of ctypes object being reclaimed by Python while Rust was still accessing it.

There were some Rust panics that happened inconsistently, and with an error message comparing two values that I was pretty sure where correct. These seem to be gone with these changes.

It also makes the core a bit more explicit about the use of a pointer to a struct vs a "value struct".

### DIFF
--- a/wgpu/backends/rs.py
+++ b/wgpu/backends/rs.py
@@ -129,21 +129,15 @@ _cstructfield2enum_alt = {
 }
 
 
-def new_struct(ctype, **kwargs):
-    """ Create an ffi struct. Provides a flatter syntax and converts our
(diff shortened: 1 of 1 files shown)


Korijn:
LGTM!

almarklein:
> any idea how you would test this?

Yes, if we're using wgpu-py for a long time without getting unpredictable errors, under a wide range of circumstances than we're probably fine!

But seriously ... I try to come up with ideas, but the issue is unpredictable. The problem is not so much that memory is released, but that it gets overwritten, and you never know when that's going to happen.

PR #102, opened 2020-05-06
Run tests on CI with SwiftShader

### LINKED ISSUE #47: How to actually run wgpu on CI?
It would be very nice if we could run tests with wgpu on CI, also in python-shader, to better validate our code. Actually, I would be a bit disappointing if we can't.

But for that Vulkan/Metal needs to be available on the CI machines, which does not appear to be the case by default (see #45), as could have been expected. There are some [posts](https://github.com/FLAMEGPU/FLAMEGPU2_dev/issues/97) suggestion that other projects setup their own local server via e.g. Gitlab runner.

I briefly looked into GPU instances (e.g. at Scaleway), but these are aimed at really high performance work and therefore really expensive.

Strictly speaking we don't need a GPU, as long as Vulkan/Metal has something to run on (Intel graphics), but I'm not sure how that'd work in a VM.

Looking at [wgpu-native](https://github.com/gfx-rs/wgpu) and [gfx-hal](https://github.com/gfx-rs/gfx) it appears that CI ensures that the build succeeds, but the resulting binary seems not to be used (though maybe that appears elsewhere).

Korijn:
Closes #47 

Let's see how far we can get with this approach.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -101,7 +101,10 @@ jobs:
         set -ex
         git clone https://github.com/google/swiftshader.git
         sudo apt-get install -y cmake
-        cd swiftshader/build
+        cd swiftshader
+        # from https://chromium.googlesource.com/chromium/src/+/master/DEPS#197
+        git checkout baa10d7b1deef695c165f185be80fdb351a6d9a7
+        cd build
         cmake ..
         make --jobs=8
         ./vk-unittests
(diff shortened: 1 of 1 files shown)


Korijn:
Well, the build takes well over 30 minutes so we'll need pre-built binaries for CI.
But first I want to see if it works at all. :)

Korijn:
Build failures. I'm currently checking out SwiftShader master where I probably should be using a "release" or "tag", but I can't find any in their git repo...

Korijn:
Well this seems rather hopeless. I'll see if I can find out if OSMesa can also provide a Vulkan Loader ICD JSON file.

Korijn:
...at a later point in time.

PR #110, opened 2020-06-15
Moved to pygfx org, and use pyshader, make CI work again

almarklein:


almarklein:
Tagging works: https://github.com/imageio/wgpu-py/releases/tag/vtest2  I've setup a Pypi token on this repo. So when we push a tag here, we should also get a release on pypi.

PR #113, opened 2020-06-29
Writing to buffer and texture

### LINKED ISSUE #98: Consider removing ctypes as part of the public API
I think we only use ctypes stuff for mapped buffers. We could instead follow the approach of wgpu-rs, and use something like `create_buffer_with_data`. We can then accept both ctypes and numpy arrays, and more, in a generic way. The downside is that we deviate from WebGPU a bit. I think it might be worth it.

almarklein:
Closes #98 

* [x] Implement `write_buffer`.
* [x] Implement `write_texture`.
* [x] Unit tests.
* [x] Test that we don't need to hold on to the data reference.
* [x] Document the above.
* [x] No more need for ctypes (see #98).
* [x] Update changelog.
* [x] Somehow guard user from using mapped memory when buffer is unmapped. Look at wgpu-rs and WebGPU?
* [x] WebGPU removed `createBufferMapped` (but forgot to remove all references, which got me confused). Use `buffer.getMappedRange` instead.

### DIFF
--- a/docs/guide.rst
+++ b/docs/guide.rst
@@ -68,8 +68,7 @@ object. These objects provide a quite versatile view on ndarray data:
 .. code-block:: py
 
     # One could, for instance read the content of a buffer
-    m = buffer.map(wgpu.MapMode.READ)
-    buffer.unmap()
+    m = buffer.read_data()
     # Cast it to float32
     m = m.cast("f")
     # Index it
--- a/docs/guide.rst
+++ b/docs/guide.rst
@@ -60,7 +60,7 @@ Communicating array data
 The wgpu-py library makes no assumptions about how you store your data.
 In places where you provide data to the API, it can consume any data
 that supports the buffer protocol, which includes ``bytes``,
-``bytearray``, ctypes arrays, and numpy arrays.
+``bytearray``, ``memoryview``, ctypes arrays, and numpy arrays.
 
 In places where data is returned, the API returns a ``memoryview``
 object. These objects provide a quite versatile view on ndarray data:
--- a/docs/guide.rst
+++ b/docs/guide.rst
@@ -54,6 +54,37 @@ be between 0.0 and 1.0 inclusive. Vertices out of this range in NDC
 will not introduce any errors, but they will be clipped.
 
 
+Communicating array data
+------------------------
+
+The wgpu-py library makes no assumptions about how you store your data.
+In places where you provide data to the API, it can consume any data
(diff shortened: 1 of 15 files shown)


almarklein:
I am pleasantly surprised by the builtin `memoryview`; it's quite a nice way to represent typed ndarray data.

docs/guide.rst
```diff
@@ -54,6 +54,37 @@ be between 0.0 and 1.0 inclusive. Vertices out of this range in NDC
...
+    m[0]
+    # Show the content
+    print(m.tolist())
+
+Chances are that you prefer Numpy. Converting the ``memoryview`` to a
+numpy array (without copying the data) is easy:
+
+.. code-block:: py
+
+    array = np.frombuffer(m, np.float32)
```
almarklein, line 85: This summarizes this PR pretty well.

almarklein: And it feels really good.

Korijn: I don't filly understand why you need to map and then unmap but the rest of it looks great! Very simple.

docs/guide.rst
```diff
@@ -54,6 +54,37 @@ be between 0.0 and 1.0 inclusive. Vertices out of this range in NDC
 will not introduce any errors, but they will be clipped.
 
 
+Communicating array data
+------------------------
+
+The wgpu-py library makes no assumptions about how you store your data.
+In places where you provide data to the API, it can consume any data
+that supports the buffer protocol, which includes ``bytes``,
+``bytearray``, ctypes arrays, and numpy arrays.
```
Korijn, line 63: Does it also accept memoryviews as input?

almarklein: Yes. Will add that.

docs/guide.rst
almarklein: ehr ... I double-checked, and as soon as you `unmap` gets called, the memory is no longer valid. An oversight on my part. We need to tweak this API a bit, because this is way it is much to easy for users to shoot themselves in the foot ...

almarklein:
I've decided to, at least for now, completely ignore the buffer-mapping API. It's very hard to come up with a Pythonic solution, and in all, I find the API overly complex, for a benefit that seems small and rare. Instead, the buffer now had a `read_data()` and `write_data()` method, which is so much simpler. Even if we ever implement mapping, we might just keep these for convenience.

almarklein:
See #114 for more details, and to track this thing.

PR #116, opened 2020-07-01
Request draw

### LINKED ISSUE #101: request_draw callable argument
From https://github.com/almarklein/visvis2/blob/master/examples/geometry_cube.py:
```python
    # Request new frame
    canvas.request_draw()


if __name__ == "__main__":
    canvas.draw_frame = animate
    app.exec_()
```

Wouldn't it be nice if that last line were `canvas.request_draw(animate)` instead of `canvas.draw_frame = animate`? Just a thought I had reading the code. The example would then become:

```python
    # Request new frame
    canvas.request_draw(animate)


if __name__ == "__main__":
    canvas.request_draw(animate)
    app.exec_()
```

Feels elegant, because less API to remember for the user?

almarklein:
Closes #101 

The `canvas.request_draw()` method can now be given a callable, which will be used to draw the frame. This way there is no need to set `canvas.draw_frame`. Less API!

The argument is optional, which means it can be called once with the drawing function, and later it can be called from code that simply wants to schedule a new draw and may not be aware of the drawing function. Plus it makes things backwards compatible.

### DIFF
--- a/docs/reference_gui.rst
+++ b/docs/reference_gui.rst
@@ -53,9 +53,8 @@ library you are using by looking what module has been imported.
     # Instantiate the canvas
     canvas = WgpuCanvas(title="Example")
 
-    # Tell the canvas what function to draw to update itself.
-    # Alternatively you can create a subclass and implement draw_frame().
-    canvas.draw_frame = your_draw_function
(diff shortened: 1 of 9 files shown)


Korijn:
Awesome!

PR #124, opened 2020-08-07
fix that running download script twice fails

almarklein:


almarklein:
Unless this was intentional @Korijn ?

PR #125, opened 2020-08-20
Limit FPS for request_draw in Qt

almarklein:
GLFW has an inherent different loop, where you basically need to write the mainloop yourself, and this is where you'd implement FPS limiting. See the `triangle_glfw.py` example.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -2,8 +2,7 @@ import re
 import sys
 
 from setuptools import find_packages, setup
-from wheel.bdist_wheel import bdist_wheel as _bdist_wheel
-from wheel.pep425tags import get_platform
+from wheel.bdist_wheel import get_platform, bdist_wheel as _bdist_wheel
 
 
 NAME = "wgpu"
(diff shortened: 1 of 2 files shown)


setup.py
```diff
@@ -2,8 +2,7 @@
 import sys
 
 from setuptools import find_packages, setup
-from wheel.bdist_wheel import bdist_wheel as _bdist_wheel
-from wheel.pep425tags import get_platform
+from wheel.bdist_wheel import get_platform, bdist_wheel as _bdist_wheel
```
almarklein, line 5: Note: `get_platform` was previously also available in the `bdist_wheel` submodule. So this should still work on older versions of wheel too.

almarklein:
It seems like I forgot to merge this :P

setup.py
Korijn: I would recommend manually verifying (by locally building) that the wheel has not changed significantly, it's a bit sensitive

almarklein: I checked the source, and we're using the exact same functions, except it's no longer present in the place where we used to import it from.

PR #127, opened 2020-10-21
black

almarklein:


tests/test_compute.py
```diff
@@ -224,9 +229,27 @@ def compute_shader(
...
+        compute_with_buffers(
+            {0: in1},
+            {
+                0: (
+                    -1,
+                    "i",
+                )
+            },
+            compute_shader,
+        )
```
almarklein, line 252: How is this supposed to be more readable? :(

Korijn: Black does not make things more readable... ;) it just enforces rules consistently

almarklein: I just wish it would enforce rules that make it more readable ;)

almarklein: This is just silly

almarklein: They should include a sillyness factor in their algorithm :)

PR #135, opened 2021-03-24
Refactor code generation for public API

almarklein:
This is the first step in the codegen refactoring in #135. It changes how the API is created. The public API `base.py` is fully defined from the WebGPU spec (IDL), not from the C header. The backend implementations `rs.py` is defined from `base.py`. Any deviations from the "reference" API must be marked with decorators, making them explicit (and document-able).

I'll focus on handling of the C-header in later PR(s).


docs/reference_wgpu.rst
```diff
@@ -52,6 +52,20 @@ to use your own version of that library instead, set the ``WGPU_LIB_PATH``
...
+This API is derived from the WebGPU spec, but differs in a few places:
+
+
+.. autodata:: wgpu.base.apidiff
+    :annotation: Differences of base API:
+
+
+.. autodata:: wgpu.backends.rs.apidiff
+    :annotation: Differences of rs backend:
+
```
almarklein, line 67: I'm really happy about this part. The codegen enforces exactly the WebGPU API, except where methods are decorated. And all decorations are collected and end up here in the docs!

Korijn:
Looks great! I don't have time to go through it with a fine comb, sorry. I trust we will identify any issues soon enough anyway :)

almarklein:
Yeah, these changes are somehow hard to review. We could go over the code in a meeting once it has progressed further.

PR #137, opened 2021-03-26
Make qt module compatible with PyQt6 and PySide6

berendkleinhaneveld:
Tested with PyQt6 and PySide6 with Python 3.9.1.

I noticed `AA_EnableHighDpiScaling` in the `enable_hidpi` call, but that does not exist anymore in Qt 6. Instead Qt will now apply scaling automatically (https://doc.qt.io/qt-6/highdpi.html). I haven't tested this, as I don't have a device with high dpi scaling available (yet).

### DIFF
--- a/wgpu/gui/qt.py
+++ b/wgpu/gui/qt.py
@@ -11,11 +11,15 @@ import importlib
 from .base import WgpuCanvasBase
 
 # Select GUI toolkit
-for libname in ("PySide2", "PyQt5", "PySide", "PyQt4"):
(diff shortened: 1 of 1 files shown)


almarklein:
I can have a try tomorrow; I have a high-res display at the office.

berendkleinhaneveld:
Qt6 example HiDPI
Seems that the hires code is working properly!

almarklein:
When I move the window around, I get errors. Turns out this `round` to get the physical size must be `ceil`. I fixed it locally. Now to find out how I can push this to your branch ...

PR #138, opened 2021-03-29
Fix some leftover renaming master -> main and various CI fixes

almarklein:
Fixes CI.


Korijn:
Regarding the linux-32 build failure: our devtools (I think its Twine?) require cryptography, who recently made the switch to Rust, and they don't support linux-32. Therefore there is no prebuilt package available, and that's also why our build fails (pip falls back to a source distribution, tries to compile it, but fails). I propose dropping linux-32 support.

The black formatting errors we get... I can't reproduce them locally. I have the same version of black installed and it says everything is formatted properly.

almarklein:
Pfew, thanks for looking into that. I suppose that's kind of a sign that linus32 is not being very popular anymore. Is the reason that we had a Docker build also because there are no linux32 runners?

Korijn:
> Pfew, thanks for looking into that. I suppose that's kind of a sign that linus32 is not being very popular anymore. Is the reason that we had a Docker build also because there are no linux32 runners?

Yes...

CI should be green again now!

PR #144, opened 2021-04-02
Enable F821, and a few codegen improvements

almarklein:
The flake8 `F821 undefined name` was ignored, because function annotations and pyshader.

* [x] The IDL parser is now able to fully resolve types to either a Python builtin, or a GPU-class/flag/enum/struct.
* [x] The codegen can therefore now generate annotations that are "valid".
* [x] The `structs` module is therefore now also much cleaner.
* [x] Where we use pyshader (in tests and examples) we still ignore this error.
* [x] Improvement to codegen regarding ordering decorators and comments.


wgpu/base.py
```diff
@@ -327,17 +327,18 @@ def create_buffer_with_data(self, *, label="", data, usage: "GPUBufferUsageFlags
...
         self,
         *,
         label="",
-        size: "GPUExtent3D",
-        mip_level_count: "GPUIntegerCoordinate" = 1,
-        sample_count: "GPUSize32" = 1,
-        dimension: "GPUTextureDimension" = "2d",
-        format: "GPUTextureFormat",
-        usage: "GPUTextureUsageFlags",
+        size: "list(int) or structs.Extent3D",
```
Korijn, line 336: is this a valid python type expression?

almarklein: Good one. It won't run, but at least flake8 considers it good enough.  Maybe we should do the `List(int)` annotation.

almarklein: ... and is there a Uninion type annotation thing?

Korijn: see https://docs.python.org/3/library/typing.html :)

In this case you would write `Union[List[int], structs.Extent3D]` I think

There's been some changes in python 3.9 too regarding this subject, it's kind of a mess

Korijn: Flake8 only does partial parsing as well. If you really want to do type checking you need to use something like mypy or one of the more modern (read faster) alternatives, but it's all very immature and rabbit-holey... I wouldn't recommend it

almarklein: The important thing is that the annotation contains some more or less useful info for the user, and that flake8 accepts it :)

I'll implement the Union and List tho.

almarklein: Done. Still wrapped in a string, otherwise things get ugly in `structs.py`, but it looks very much like mypy code now :)

PR #145, opened 2021-04-02
Implement codegen for rs backend

almarklein:
* [x] Write `rs_mappings.py` with enum->int mappings.
* [x] Validate structs created in the code, and remove checking structs at runtime.
* [x] Compare flags and report differences.
* [x] Validate use of functions.
* [x] Test some of the new code.
* [x] Add tests that use the new codegen to verify that the code is up to date in CI.
* [x] Better structure all codegen functions and make their use more similar.
* [x] Since we now have more codegen-time validation, we can remove some runtime checks.
* [x] Remove old codegen file that was still kept as a reference.


almarklein:
Ready!

PR #156, opened 2021-04-16
Update to wgpu-native 0.7.0

### LINKED ISSUE #126: Adapt to wgpu-native 0.6.0!

almarklein:
Closes #126 

This PR:
* [x] Download new headerfile
* [x] Manually fix things, then more things until all examples work again.
* [x] Make tests work again.
* [x] Update destroy methods and see if they don't break now.
* [x] Add more tests to get full coverage of rs.py.
* [x] Self review.
* [x] Update release notes.
* [x] Update pygfx.

API changes:

* `GPUBuffer.read_data()` and `GPUBuffer.write_data()` are removed in favor of the methods on `GPUQueue`. The `queue.write_buffer()` is original WebGPU API. We add `read_buffer()` for symmetry. Having the methods the buffer class is not viable anymore, because stricter rules now prevent e.g. a buffer usage being both STORAGE and MAP_READ. A temporary buffer is thus needed to copy data over (and that makes having the method on the buffer class misleading).
* More restrictive rules for format enums and usage flags.
* When copying from/to textures, the `bytes_per_row` must now be a multiple of 256.
* Naga performs some shader validation, and for some reason `VertexId` produces an `UnsupportedBuiltIn` warning. This (I think) causes the swap chain status to be reported as suboptimal. 

Notes:
* Since wgpu-native is partly underway in following webgpu.h, some of their API is now camelCase, and some is snake_case.
* We should create a new release of wgpu-native soonish. The version number baked in the lib was not bumped :) Also ... `wgpu.h` is down to just a few lines since it now includes `webgpu.h`! Good news, I guess, but we'll need to update some things accordingly.
* There's a restriction on the texture row size (in bytes) when copying texture data between textures and buffers. The `queue.write_texture` does not have this restriction. Added `queue.read_texture` for symmetry.
* Let's enable wgsl shaders!


Korijn:
Any explanation as to _why_ "When copying from/to textures, the bytes_per_row must now be a multiple of 256."?

almarklein:
> Any explanation as to why "When copying from/to textures, the bytes_per_row must now be a multiple of 256."?

This is a restriction imposed by webgpu. Probably because some hardware has this limitation. It was mentioned in the WebGPU spec earlier, but now wgpu-native is actively requiring it.

.github/workflows/ci.yml
```diff
@@ -126,7 +126,7 @@ jobs:
       run: |
```
Korijn, line 126: might want to update the name of this step now :)

PR #159, opened 2021-04-23
Add support for WGSL

almarklein:
This allows consuming WGSL in addition to SpirV.

I'm having some trouble rewriting the examples though, and the error messages are not very helpful, e.g.:
```
Running script: "C:\dev\pylib\wgpu\examples\triangle_glfw.py"
|	[[stage(vertex)]]
>	fn vs_main([[builtin(vertex_index)]] index: u32) -> [[builtin(position)]] vec4<f32> {
|	             ^
thread '<unnamed>' panicked at 'Parsing', src\device.rs:694:5
```

I'm having more luck by using Naga to compile some wgsl to spirv and then using that. This suggests that we may just need a newer version of wgpu-native that has a newer naga. But even then, arrays seem problematic, or I am doing them wrong.

We can merge this as-is and continue when we've updated to a more recent wgpu-native.

### DIFF
--- a/wgpu/backends/rs.py
+++ b/wgpu/backends/rs.py
@@ -694,9 +694,11 @@ class GPUDevice(base.GPUDevice, GPUObjectBase):
 
         if isinstance(code, str):
             # WGSL
+            # H: chain: WGPUChainedStruct, source: const char
             source_struct = new_struct_p(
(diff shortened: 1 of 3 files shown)


Korijn:
LGTM!

PR #162, opened 2021-04-29
Update to latest wgpu-native 

almarklein:
Updates to (upcoming) wgpu-native v0.8.0.1

Wgpu-native has undergone massive changes. Instead of generating the C header from the Rust code, they now do it the other way around. The good thing is that its header file is an include of `webgpu.h` plus some extra niceties for desktop. The downside is that in this process some functionality has been lost. In other words, the transition is not quite finished yet. So I contributed the stuff that we need in wgpu-py. 

* [x] Adjust code for latest wgpu-native.
* [x] Turn our examples into using WGSL.
* [x] Make tests work.
* [x] Fix cube example.
* [x] Remove dev code in `rs.py`.
* [x] Check for other todo's that I may have left in the code.
* [x] Get rid of all pyshader dependencies -> use different way to compose uniform structs.
* [x] Revive destroy functions?
* [x] Release wgpu-native -> this should fix our CI.

Necessary contributions to wgpu-native:

* Implemented several functions that were removed during the recent big refactoring.
* Implemented missing pieces in bind groups (support anything but buffers was lost).
* Update to more recent wgpu-core and naga.

Regressions that are still open (we can fix them or wait for them to get fixed):

*  Wayland support fix/check.
* The adapter/device features.
* The adapter/device limits.
* Debug groups and markers.
* Preferred format from surface.
* Possibility to select backend, or make sure that Vulkan is selected over DX (it should but isn't always).


Korijn:
Was the jump from 7.0.0 to 0.7.0.2 so big?

almarklein:
The idea is for wgpu-native to follow the versioning of wgpu-core, plus a 4th index as a kind of build number. Currently, wgpu-native targets wgpu-core 0.7.0. The latest release is 0.7.0.2 because I first created a tag v0.7.0.1, but it wasn't targeting the latest commit, because I had done `git pull origin master` and not `git pull upstream master` ... sigh. 

But yeah, there is also extra cool stuff: a later version of Naga, so we can write shaders in wgsl!  The downside is that the shader validation is now on by default, and it won't accept the spirv generated by pyshader :'(  

We'll need to think about how to deal with that. Some options from the top of my head:
- A hard switch. Make a release of wgpu-py before this PR that we can use in pygfx for the time being. Then rewrite all pygfx shaders to wgsl. 
- Create our own build of wgpu-native that disables the validation (if possible).
- Fix pyshader to produce better SpirV. I'd rather not, but maybe it isn't so much work? Problem is that the error messages that Naga currently produces can be very cryptic ...

Korijn:
Only the hard switch to WGSL seems like it would be productive in the long run. Do we really have that many shaders to translate already? It can't be that bad, right?

Separately I would hope we can do something about cryptic error messages.. :/

almarklein:
> It can't be that bad, right?

Indeed, how hard can it be? :D   you're right that there are not that many shaders and they can be fixed one by one (no interconnections or something complex like that). My main concern, I think, is said cryptic error messages. Well, maybe I should give it a shot with one material first ...

Korijn:
Is there really no switch or option to enable better error messages? :(

almarklein:
> Is there really no switch or option to enable better error messages?

There may be ways to improve this. The error messages for the Spirv that PyShader generates are especially unhelpful, because they are about SpirV, using indices to objects to refer to stuff. The error messages that Naga generates for wgsl are usually more helpful, but I've also experienced that wgpu would panic, while letting naga check the wgsl from the command line showed no errors. Naga is moving fast, could be that things can be improved simply by having wgpu-native using a more recent version of naga.

Korijn:
It seems that our most economic option is to just make a hard switch to WGSL and wait for Naga to improve.

almarklein:
Interesting ... the Windows backend is now actually running all the tests using the DirectX backend.

almarklein:
Getting there ...

PR #170, opened 2021-06-10
Force backend by WGPU_BACKEND_TYPE env var

berendkleinhaneveld:
This makes sure that on other platforms than Windows any available backend is picked.

### DIFF
--- a/wgpu/backends/rs.py
+++ b/wgpu/backends/rs.py
@@ -223,24 +223,13 @@ class GPU(base.GPU):
         force_backend = enum_str2int["BackendType"]["Null"]
         if "WGPU_BACKEND_TYPE" in os.environ:
             try:
-                backend = int(os.environ["WGPU_BACKEND_TYPE"])
-                value = next(
-                    (
-                        key
-                        for key, value in enum_str2int["BackendType"].items()
-                        if value == backend
-                    ),
-                    None,
-                )
-                if not value:
-                    logger.warn(f"Invalid value for WGPU_BACKEND_TYPE: '{backend}'")
-                else:
-                    logger.warn(f"Forcing backend: {value} ({backend})")
-                    force_backend = backend
-            except ValueError:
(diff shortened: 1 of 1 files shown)


wgpu/backends/rs.py
```diff
@@ -217,7 +218,9 @@ def request_adapter(self, *, canvas, power_preference=None):
         # mature than Metal too, so let's just force that for now.
```
Korijn, line 218: Can you update the comment accordingly?

Korijn:
Is the Metal backend mature enough for usage yet?

almarklein:
> Is the Metal backend mature enough for usage yet?

I guess Berend will find out :)

Korijn:
As discussed: let's only force a backend if a certain envvar is given

wgpu/backends/rs.py
```diff
@@ -213,14 +213,35 @@ def request_adapter(self, *, canvas, power_preference=None):
...
+        # the NVidia control panel settings.
         # See https://github.com/gfx-rs/wgpu/issues/1416
-        # force_backend = lib.WGPUBackendType_Vulkan
-        force_backend = enum_str2int["BackendType"][
-            "Vulkan" if sys.platform.startswith("win") else "Null"
-        ]
+        force_backend = enum_str2int["BackendType"]["Null"]
+        if "WGPU_BACKEND_TYPE" in os.environ:
+            try:
+                backend = int(os.environ["WGPU_BACKEND_TYPE"])
```
Korijn, line 226: do you have to pick by integer? that seems... unfriendly?

berendkleinhaneveld: Good call!

Korijn:
LGTM!

almarklein:
I would expect this to fail, but I forgot that I had simply disabled all render-tests on CI. With this change, there are some tests running on CI (e.g. juggling with buffers and textures) which were previously skipped :)

PR #172, opened 2021-06-15
Force backend

almarklein:


wgpu/backends/rs.py
```diff
@@ -219,22 +219,23 @@ def request_adapter(self, *, canvas, power_preference=None):
...
-                logger.warn(f"Forcing backend: {backend} ({force_backend})")
+                backend = enum_str2int["BackendType"][force_backend]
             except KeyError:
                 logger.warn(
-                    f"Invalid value for WGPU_BACKEND_TYPE: '{backend}'.\n"
+                    f"Invalid value for WGPU_BACKEND_TYPE: '{force_backend}'.\n"
                     f"Valid values are: {list(enum_str2int['BackendType'].keys())}"
                 )
+            else:
+                logger.warn(f"Forcing backend: {force_backend} ({backend})")
```
Korijn, line 233: You could avoid the else by just putting this logger statement in the try block :)

PR #173, opened 2021-07-02
Implement support for off-screen canvases

almarklein:
This makes it possible to write Canvas classes that do not render to a screen, but instead get a texture on each draw, which can be processed  in whatever way makes sense.

This paves the way for e.g. a Jupyter widget.

Note that the WebGPU API for CanvasContext and GPUChain has changed (they are merged into a GPUPresentContext). We should update accordingly soon, because that might affect the API a bit.

### DIFF
--- a/wgpu/gui/base.py
+++ b/wgpu/gui/base.py
@@ -16,6 +16,9 @@ class WgpuCanvasInterface(base.GPUCanvasContext):
 
     # NOTE: It is not necessary to actually subclass this class.
 
+    # Whether to present to a surface or to a texture
+    _PRESENT_TO_SURFACE = True
+
     def __init__(self, *args, **kwargs):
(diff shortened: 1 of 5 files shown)


wgpu/gui/base.py
```diff
@@ -16,6 +16,9 @@ class WgpuCanvasInterface(base.GPUCanvasContext):
 
     # NOTE: It is not necessary to actually subclass this class.
 
+    # Whether to present to a surface or to a texture
+    _PRESENT_TO_SURFACE = True
```
Korijn, line 20: This could be a property that checks if _present() raises NotImplementedError. Would save a bit of bookkeeping :)

almarklein:
Merging now, so I can have a go at the next step. Feel free to still comment here, I can take things into account in upcoming work.

PR #174, opened 2021-07-05
Update to latest webgpu

almarklein:
* The swap-chain class has been removed. This object was used as a context manager.
  One should now use the present context object instead (see below). The
  drawn result is now automatically presented to the screen at the end of a draw event.
* The `canvas.configure_swap_chain()` method has been removed. Instead,
  `canvas.get_context()` should be used, to obtain a present context.
* The new `GPUPresentationContext` class can be `configure()`'d, and a texture can be obtained via
  `context.get_current_texture()`.
* Adds `WgpuOffscreenCanvas` to support off-screen rendering while using a canvas API.
* The `adapter.request_device()` method has its arguments `non_guaranteed_features` and `non_guaranteed_limits`
  replaced with `required_features` and `required_limits`.
* The enum field `StoreOp.clear` is now `StoreOp.discard`.
* Adds `adapter.is_software` property.

### DIFF
--- a/wgpu/gui/offscreen.py
+++ b/wgpu/gui/offscreen.py
@@ -1,0 +1,71 @@
+from .. import base, flags
+from ..gui.base import WgpuCanvasBase
+
+
+class WgpuOffscreenCanvas(WgpuCanvasBase):
+    """ Base class for off-screen canvases, providing a custom presentation
+    context that renders to a tetxure instead of a surface/screen.
+    The resulting texture view is passed to the ``present()`` method.
+    """
+
+    def get_window_id(self):
+        """ This canvas does not correspond to an on-screen window.
+        """
+        return None
+
+    def get_context(self, kind='gpupresent'):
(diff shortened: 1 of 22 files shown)


Korijn:
I'm happy to see the context manager disappear :)

almarklein:
> I'm happy to see the context manager disappear :)

Me too :)

wgpu/gui/offscreen.py
```diff
@@ -0,0 +1,69 @@
...
+
+class WgpuOffscreenCanvas(WgpuCanvasBase):
+    """Base class for off-screen canvases, providing a custom presentation
+    context that renders to a tetxure instead of a surface/screen.
+    The resulting texture view is passed to the ``present()`` method.
+    """
+
+    def get_window_id(self):
+        """This canvas does not correspond to an on-screen window."""
+        return None
```
almarklein, line 13: Look @Korijn, that `_PRESENT_TO_SURFACE` is also gone.

Korijn:
LGMT

PR #175, opened 2021-07-05
Updates for latest wgpu-native

almarklein:
* [x] (wait for) release wgpu-native
* [x] Bump to latest wgpu-native
* [x] Apply any necessary changes

The wgpu-native has bumped its version of wgpu-core and Naga. The only changes that matter to us are that Naga is a bit more strict with indexing (can't index into arrays created with `let`, but must use `var` instead).


almarklein:
This is now held back by https://github.com/gfx-rs/wgpu-native/issues/118

PR #178, opened 2021-08-06
Jupyter canvas

### LINKED ISSUE #176: Jupyter widget based on jupyter_rfb

almarklein:
Closes #176

I'm going to leave this open until I've updated pygfx, so I can make sure that it works well for pygfx too.

jupyter_rfb3

### DIFF
--- a/wgpu/gui/jupyter.py
+++ b/wgpu/gui/jupyter.py
@@ -1,0 +1,88 @@
+"""
+Support for rendering in a Jupyter widget. Provides a widget subclass that
+can be used as cell output, or embedded in a ipywidgets gui.
+"""
+
+from .offscreen import WgpuOffscreenCanvas
+
+import numpy as np
+from jupyter_rfb import RemoteFrameBuffer
+
+
+class JupyterWgpuCanvas(WgpuOffscreenCanvas, RemoteFrameBuffer):
+    """An ipywidgets widget providing a wgpu canvas. Needs the jupyter_rfb library."""
+
+    def __init__(self):
+        super().__init__()
+        self._pixel_ratio = 1
+        self._logical_size = 0, 0
+        self._is_closed = False
+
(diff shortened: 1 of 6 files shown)


wgpu/gui/jupyter.py
```diff
@@ -0,0 +1,88 @@
...
+        return int(self._logical_size[0] * self._pixel_ratio), int(
+            self._logical_size[1] * self._pixel_ratio
+        )
+
+    def set_logical_size(self, width, height):
+        self.css_width = f"{width}px"
+        self.css_height = f"{height}px"
+
+    def close(self):
+        RemoteFrameBuffer.close()
```
Korijn, line 56: Don't you need to pass `self` here?

Korijn:
Can you tell me a little bit about how this works if someone has multiple cells with rendered scenes in each cell?

wgpu/gui/jupyter.py
almarklein: oooh nice catch!

almarklein:
> Can you tell me a little bit about how this works if someone has multiple cells with rendered scenes in each cell?

Good question. Each cell would probably create a new canvas with a new renderer. So the device would be shared (we should probably make that so), but each canvas comes with its own pipeline(s). I should make a notebook with multiple cells to get a better feel. Will do when we can use pygfx.

almarklein:
image

almarklein:
BTW: you may see in the screenshots that the colors are a bit off. This has been fixed too by using the appopriate texture format.

Korijn:
What happens if you leave out the animation loop and just render once? Does that work?

almarklein:
Yes. Then a redraw is only requested when the canvas is resized (in the browser) like with normal windows.

almarklein:
Examples:  https://jupyter-rfb.readthedocs.io/en/latest/examples/

PR #179, opened 2021-08-11
Our Pyinstaller hook should now just work

### LINKED ISSUE #25: PyInstaller compatibility
To support freezing apps with wgpu-py, we need to take a couple of things into account:

* PyInstaller can't see dynamic imports, so static imports of specific backends are required; otherwise nested dependencies are not included in the frozen app. See https://github.com/almarklein/wgpu-py/pull/18#issuecomment-569410315 for the original discussion.
* Our package data (`wgpu/resources` folder) won't be included automatically either, unless we submit a hook to the pyinstaller repo or users write their own, containing a call to `collect_data_files('wgpu.resources')`. See https://github.com/pyinstaller/pyinstaller/blob/develop/PyInstaller/hooks/hook-certifi.py for an example

Progress:
* [x] Implement the new hook system in wgpu-py
* [x] Test it in CI using the PR branch of PyInstaller
* [x] Make a release of wgpu-py so PyInstaller can use it in their tests.
* [x] PyInstaller branch merged (https://github.com/pyinstaller/pyinstaller/pull/4582)
* [x] PyInstaller 4 release
* [ ] Update wgpu-py dev requirments to use `pyinstaller>=4` instead of the branch

almarklein:
I think this closes #25.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -78,8 +78,7 @@ jobs:
     - name: Test PyInstaller
       if: matrix.SKIPTEST != 1 && (matrix.pyversion == '3.6' || matrix.pyversion == '3.7')
       run: |
-          pip install psutil
-          pip install https://github.com/pyinstaller/pyinstaller/archive/develop.zip
+          pip install psutil pyinstaller>=4
           pytest -v wgpu/__pyinstaller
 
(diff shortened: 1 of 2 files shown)


Korijn:
LGTM!

PR #180, opened 2021-08-20
On Windows force Vulkan by default for now

almarklein:
I want to avoid that users need to set an env var in order to make things work.


Korijn:
Sure! Windows user can still force the regular behavior by setting the env var to an empty string.

almarklein:
> Windows users can still force the regular behavior by setting the env var to an empty string.

Well spotted ;)

PR #185, opened 2021-09-24
Update to wgpu-native

almarklein:
* Update wgpu-native:
  * [x] Update wgpu-native headers.
  * [x] Update wgpu-native to latest wgpu-core and naga.
  * [x] wgpu-native release.

* Update wgpu-py:
  * [x] Run `python download-wgpu-native.py --version xxxx`.
  * [x] Update to changes related to headers.
  * [x] Update to changes related to wgpu-core.
  * [x] Implement limits.
  * [ ] Implement drop methods. -> different PR


almarklein:
Ready! 🎆

download-wgpu-native.py
```diff
@@ -75,7 +76,17 @@ def get_os_string():
...
 
 def get_arch():
-    return "64" if sys.maxsize > 2 ** 32 else "32"  # True on 64-bit Python interpreters
+    # See e.g.: https://stackoverflow.com/questions/45124888
+    is_64_bit = sys.maxsize > 2 ** 32
+    machine = platform.machine()
+    if not is_64_bit:
+        return "i686"
+    elif machine.startswith(("arm", "aarch64")):
+        # Includes Raspberry Pi, MacOS M1, ...
```
almarklein, line 85: Maybe quick check with @berendkleinhaneveld that this indeed detects an M1 :)

wgpu/backends/rs_mappings.py
```diff
@@ -236,23 +251,25 @@
 enum_str2int = {
     "BackendType": {
         "Null": 0,
-        "D3D11": 1,
-        "D3D12": 2,
-        "Metal": 3,
-        "Vulkan": 4,
-        "OpenGL": 5,
-        "OpenGLES": 6,
+        "WebGPU": 1,
```
Korijn, line 254: Huh? How can webgpu be a webgpu backend?

almarklein: Good question. On WebGPU you can't select the backend, so its not that it leaked from that spec. Odd...

berendkleinhaneveld: The value is the same as D3D11, which is also suspect. 🤔

download-wgpu-native.py
berendkleinhaneveld: Will check in the morning

wgpu/backends/rs_mappings.py
berendkleinhaneveld: https://github.com/gfx-rs/wgpu/blob/76bcd9c494a4359853bc2caec8eebbaa990a34e3/wgpu-info/src/main.rs#L127 webgpu backend is used to indicate browser apparently

berendkleinhaneveld: Ah, int value should be 6: https://github.com/gfx-rs/wgpu/blob/1ee6036f440210e038098bcc372ddd738437918f/wgpu-types/src/lib.rs#L64

berendkleinhaneveld: OpenGLES seems to have been replaced?

download-wgpu-native.py
berendkleinhaneveld: `arm64`! So we're good to go.

wgpu/backends/rs_mappings.py
almarklein: > webgpu backend is used to indicate browser apparently

Right, maybe just to signal to wgpu-core that its running in a browser, maybe having some extra checks or constraints.

I can imagine that OpenGLES is dropped because it does not have compute shaders for instance. Not sure why it has ever been a backend.

almarklein:
Ready to go @Korijn ?

PR #189, opened 2021-10-05
rewrite of CAMetalLayer poiner extraction code to work on Apple aarch64

marcdownie:
…64 ABI

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -28,7 +28,7 @@ elif sys.platform.startswith("darwin"):
 else:
     pass  # don't include binaries; user will have to arrange for the lib
 
-runtime_deps = ["cffi>=1.10"]
+runtime_deps = ["cffi>=1.10", "rubicon-objc>=0.4.1"]
 
 
 setup(
--- a/wgpu/backends/rs_helpers.py
+++ b/wgpu/backends/rs_helpers.py
@@ -4,6 +4,8 @@ import ctypes
 
 from .rs_ffi import ffi, lib
 
+if sys.platform.startswith("darwin"):
+    from rubicon.objc.api import ObjCInstance, ObjCClass
 
 def get_memoryview_and_address(data):
     """Get a memoryview for the given data and its memory address.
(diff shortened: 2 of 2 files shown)


setup.py
```diff
@@ -28,7 +28,7 @@ def finalize_options(self):
 else:
     pass  # don't include binaries; user will have to arrange for the lib
 
-runtime_deps = ["cffi>=1.10"]
+runtime_deps = ["cffi>=1.10", "rubicon-objc>=0.4.1"]
```
almarklein, line 31: This should make the dependency for osx only.
```suggestion
runtime_deps = ["cffi>=1.10", "rubicon-objc>=0.4.1; platform_system=='Darwin'"]
```

Korijn: I'm pretty sure that should be `sys_platform == 'darwin'`

almarklein: I think both are fine :)
See https://docs.python.org/3/library/platform.html#platform.system and https://www.python.org/dev/peps/pep-0508/#environment-markers

almarklein: But if you've used `sys_platform` in another project, I would feel safer using that, haha.

Korijn: I am unsure!

almarklein: I'm pretty sure both should work. Should be easy enough to test.

almarklein: I just tested this line. If I run `pip install .` with exactly this line, then `rubicon-objc` is not installed. If I replace "Darwin" with "Windows" then it does.

Korijn: > I don't think we can make the installation happen with those flags via `install_requires` in our `setup.py`, or can we? And even if we could, I don't think our `setup.py` would be the correct place for such a fix.

That's correct, you would need to instruct users to install CFFI like that manually via the README. Sadly.

almarklein: @marcdownie if we add this change, than this PR is good to go, I think!

almarklein:
One question, has this also been tested with Rosetta, just to make sure that we don't break something else :)

Issue #136, opened 2021-03-24
Mac M1 support


anentropic:
-----

Overview (edited by AK):

* [x] wgpu-native must work with macos_arm64: https://github.com/gfx-rs/wgpu-native/pull/114
* [x] wgpu-native must build binaries for macos_arm64 in a release: https://github.com/gfx-rs/wgpu-native/pull/139
* [ ] CFFI must work on M1: #190
  * [ ] Convert cffi callback usage to new-style
  * [x] Resolve problems with cffi wheels for M1 #196
* [x] Make our `get_surface_id_from_canvas` work for M1. #195
* [x] We must update our CI to build wheels for macos_arm64: #194

-----

I was getting this error:

```
  • Installing wgpu (0.3.0): Failed

  EnvCommandError

  Command ['/Users/anentropic/Library/Caches/pypoetry/virtualenvs/hexblade-gi2zk5D5-py3.9/bin/pip', 'install', '--no-deps', 'file:///Users/anentropic/Library/Caches/pypoetry/artifacts/aa/f3/b2/8caef6980405490715f8040d2349bddb320b3962d15c24f0525b66d749/wgpu-0.3.0.tar.gz'] errored with the following return code 1, and output:
  Processing /Users/anentropic/Library/Caches/pypoetry/artifacts/aa/f3/b2/8caef6980405490715f8040d2349bddb320b3962d15c24f0525b66d749/wgpu-0.3.0.tar.gz
      ERROR: Command errored out with exit status 1:
       command: /Users/anentropic/Library/Caches/pypoetry/virtualenvs/hexblade-gi2zk5D5-py3.9/bin/python -c 'import sys, setuptools, tokenize; sys.argv[0] = '"'"'/private/var/folders/w1/_vgkxyln4c7bk8kr29s1y1k00000gn/T/pip-req-build-99e_f9gh/setup.py'"'"'; __file__='"'"'/private/var/folders/w1/_vgkxyln4c7bk8kr29s1y1k00000gn/T/pip-req-build-99e_f9gh/setup.py'"'"';f=getattr(tokenize, '"'"'open'"'"', open)(__file__);code=f.read().replace('"'"'\r\n'"'"', '"'"'\n'"'"');f.close();exec(compile(code, __file__, '"'"'exec'"'"'))' egg_info --egg-base /private/var/folders/w1/_vgkxyln4c7bk8kr29s1y1k00000gn/T/pip-pip-egg-info-23tylv6i
[...]

PR #194, opened 2021-10-07
Cibuildwheel

Korijn:
This PR leverages [cibuildwheel](https://github.com/pypa/cibuildwheel/) to handle creation of wheels, including macos arm64.

Checks 1 box in #136 as this works around the missing Apple Silicon runners on CI.

This PR makes the build configuration PEP517 and [pypa/build](https://github.com/pypa/build) compatible, allowing the `pip wheel` command to be used by cibuildwheel.

So to summarize, all the following commands are now legal:

- `python setup.py bdist_wheel`
- `pip wheel --no-deps .`
- `cibuildwheel --platform <plat>`

Build artifacts:

image

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -84,6 +84,13 @@ jobs:
       uses: actions/setup-python@v2
       with:
         python-version: ${{ matrix.pyversion }}
+    - name: Install llvmpipe and lavapipe for offscreen canvas
+      if: matrix.os == 'ubuntu-latest'
+      run: |
+        sudo apt-get update -y -qq
+        sudo add-apt-repository ppa:oibaf/graphics-drivers -y
+        sudo apt-get update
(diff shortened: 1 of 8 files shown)


Korijn:
> Nice!
> 
> One question: where is it defined what build are made exactly? E.g. why does it not try to build linux_arm64. Or phrased differently, what would we change when we have wgpu-native builds for linux_arm_64?

See pyproject.toml! I've used the `build` and `arch` options. Related docs: https://cibuildwheel.readthedocs.io/en/stable/options/#build-skip

Korijn:
> Ah, so now it's using the default, which is x86_64 and i686 for Linux?

Indeed!

Korijn:
Simplified the sdist-build job and fixed a bug in the publish job where it didn't depend on the sdist-build job. Should be good to go!

Issue #190, opened 2021-10-06
Making cffi work on Mac M1


almarklein:
See (the top post of) #136 for the bigger picture. I'm trying to collect what we know here (and may update).

It looks like we have two problems with CFFI.

### Problem

There are no binary wheels provided for macos_arm64, so it tries to use the x86 one (or something like this). Also see the corresponding issue: https://foss.heptapod.net/pypy/cffi/-/issues/506 - a workaround has been proposed, and it looks like they'll wait for someone else to fix it.

This error that you'd get:

> E ImportError: dlopen(/Users/xx/pypoetry/virtualenvs/wgpu-5bSf_T1V-py3.9/lib/python3.9/site-packages/_cffi_backend.cpython-39-darwin.so, 2): Symbol not found: _ffi_prep_closure
> E Referenced from: /Users/xx/pypoetry/virtualenvs/wgpu-5bSf_T1V-py3.9/lib/python3.9/site-packages/_cffi_backend.cpython-39-darwin.so
> E Expected in: flat namespace
> E in /Users/xx/pypoetry/virtualenvs/wgpu-5bSf_T1V-py3.9/lib/python3.9/site-packages/_cffi_backend.cpython-39-darwin.so

### Solutions

Ideally someone (not me) steps up and helps cffi build wheels for macos_arm64. No idea how hard this is.  

A workaround has been proposed:
```
pip install cffi --no-binary :all:
```

And a somewhat more complicated version:
```
LDFLAGS=-L$(brew --prefix libffi)/lib CFLAGS=-I$(brew --prefix libffi)/include pip install cffi --no-binary :all:
```

I've heard mixed results so far though.

It looks like installing via conda produces a working version.

If we can confirm that the above workaround works, we can add that in our installation instructions: "bla use conda or pip install cffi --no-binary :all:".

## 2. Callbacks are broken

Something to do with how callbacks work in cffi, and Mac nagging that the binary is unsigned. Read more here: https://cffi.readthedocs.io/en/latest/using.html#callbacks-old-style

The error that you'd get:

> MemoryError: Cannot allocate write+execute memory for ffi.callback(). You might be running on a system that prevents this. 

The cffi docs mention that the "new callback mechanism" does not have this problem. We only have 2 callbacks, so we might be able to fix this. I'll look into this.

cc @korijn @berendkleinhaneveld @marcdownie @SuperSimon81 @anentropic

berendkleinhaneveld:
[...]

PR #196, opened 2021-10-08
Update cffi dependency to 1.15.0rc2 to fix issue with macOS arm

berendkleinhaneveld:
Resolve problems with cffi wheels for M1 (see #136  and #190 ).

@marcdownie Could you check to see if you still get the MemoryError or not (https://github.com/pygfx/wgpu-py/issues/136#issuecomment-934485045)? I don't have a conda setup on my machine so I can't verify this.

### DIFF
--- a/dev-requirements.txt
+++ b/dev-requirements.txt
@@ -7,7 +7,6 @@ flake8
 flake8-black
 pep8-naming
 sphinx
-# cffi>=1.15.0rc2
 
 # Building wheels
 wheel


Korijn:
For some reason this breaks the pyinstaller tests? :/

berendkleinhaneveld:
I installed glfw with brew and after that I could run the glfw examples without getting the MemoryError, so I can't replicate it.

PR #202, opened 2021-10-26
Automatic gui selection

almarklein:
Relates to https://github.com/pygfx/pygfx/discussions/141 and https://github.com/pygfx/pygfx/pull/140

## Introduction

The is a proof of concept for how we can make it possible to write examples without explicitly selecting a GUI.

What this PR does:
* The glfw gui-backend provides an API for processing user events. Incidentally ;) its exactly the same API as jupyter_rfb.
* It also adds a `run()`  and `call_later()` function to the glfw and jupyter backends.
* It adds `gui/auto.py`, so you can `from wgpu.gui.auto import WgpuCanvas, run`. This is where the gui selection will take place.
* Updates two examples to show what it'd look like.

What we'd get is:
* Examples (and simple apps) in wgpu-py and pygfx run (unchanged) on desktop as well as in Jupyter.
* Less boilerplate for creating an application etc.
* Connecting a pygfx controller (camera interaction) becomes a oneliner.

## How we can take this forward

This is currently a draft with multiple open ending. I think there are 3 ways we can take this further:

1: Stick with just glfw and jupyter. Lobby/help glfw to use cibuildwheel, so glfw becomes available also on e.g. Mac M1. We can then rely on these two gui-backends for "simple stuff". The other backends (qt, wx) are meant for users who want to embed visualizations in an application. They will need to implement user interaction themselves, which is probably fine (or even preferable) when you create an actual application. Advantages are that its easier to maintain and that we can assume asyncio (and can ditch the `call_later()` function).

2: Force all gui backends to also provide the `handle_event`, `run()` and `call_later()` api. It's not actually *that* hard.

3: A middle ground: Allow gui backends to either support a minimal API or a fully featured one, making it eligible for selection via `wgpu.gui.auto`.

I don't think I'm in favor of 2, because it would make it harder to add and maintain a wide variety of backends. I think I might be leaning towards 1, but we can start with 3, and decide to move towards 1 later.

## Todo

* [x] Decide what direction to take.
* [x] Actually implement `is_jupyter()`.
* [x] Improve specification (and remove ambiguouties) of what the events should look like at jupyter_rfb.
* [x] Perhaps some upstream tweaks to jupyter_rfb.
* [x] Cleanup and self-review.
* [x] Warning upon import to install glfw.


almarklein:
Glfw is almost ready for M1. Not released yet, but its CI can build the wheels now:
image

The CI of glfw seems to build from source for Linux (x86, aarch64 and i686), and for Windows and Mac rely on the pre-compiled binaries from the GLFW project. These now include binaries for M1.

almarklein:
I'm ready here. This also resulted in some upstream changes to jupyter_rfb. These have yet to be released, but apart from an improvement to wheel event capturing, these changes relate to the definition of the event specification.

Korijn:
GLFW 2.4.0 appears to be available for M1 on pypi now as well!

Korijn:
I tested this on Windows, seems to work from command line (glfw) and jupyter notebook in chrome. Had some problems in VSCode, see https://github.com/vispy/jupyter_rfb/issues/53.

Korijn:
> I don't think I'm in favor of 2, because it would make it harder to add and maintain a wide variety of backends. I think I might be leaning towards 1, but we can start with 3, and decide to move towards 1 later.

Agree with 1. :)

I was wondering if you could describe a roadmap towards introducing automatic support for CI (executing the examples on CI should fall back to rendering to a texture I guess). Similar it would be nice to trigger this behaviour when building docs, so I guess an environment variable might be nice to toggle the alternative behaviour.

PR #208, opened 2021-12-09
Refactor GUI's

### LINKED ISSUE #142: Separate WgpuCanvas into toplevel and subwidget classes
A bit nitpicking, but ...

Right now we've implemented the GUI's to do:
```py
from wgpu.gui.whichever import WgpuCanvas
```
This makes the examples nicely consistent. Also, that class can be instantiated to work as a toplevel widget or a subwidget.  I guess we did it this way because it's trivial with Qt. For wx I had to use some `__new__` to make it work (#141). For other GUI backends it may similarly not be easy.

I propose something like:
```py
from wgpu.gui.whichever import WgpuCanvas
from wgpu.gui.whichever_support_subwidgets import WgpuCanvasWidget
```

almarklein:
Closes #142

* [x] For Qt, provide two classes instead of one: one for embedding, and one for a toplevel widget (as proposed in #142).
* [x] Dito for wx, also fixing/improving the wx backend a bit.
* [x] Clean up: remove stub flexx module and glfw specific examples.
* [x] Update docs and readme.


almarklein:
I verified that the wx backend now works with pygfx (with a minor change that I will push later)!

almarklein:
cc @Correct-Syntax

README.md
```diff
@@ -38,15 +38,16 @@ To get an idea of what this API looks like have a look at [triangle.py](https://
 pip install wgpu
 ```
 
-The library ships with Rust binaries for Windows, MacOS and Linux. If you want to use
+The wheels include the prebuild binaries. If you want to use
```
Korijn, line 41: prebuilt* :)

README.md
```diff
@@ -38,15 +38,16 @@ To get an idea of what this API looks like have a look at [triangle.py](https://
...
 a custom build instead, you can set the environment variable `WGPU_LIB_PATH`.
+You probably also want to install ``glwf`` (for desktop) and/or ``jupyter_rfb`` (for Jupyter).
```
Korijn, line 43: double backticks?

almarklein: I was editing rst right before editing this ;)

Korijn:
LGTM

PR #212, opened 2021-12-14
Use queue.write_texture instead of copy_to_texture with a temp buffer

almarklein:
This is cleaner and shorter, and avoids the (hopefully temporary) bug described in https://github.com/pygfx/pygfx/issues/219.


Korijn:
LGTM

PR #214, opened 2021-12-14
WX: max_fps (and fix request_draw on Windows)

### LINKED ISSUE #209: The wx gui only updates when mouse is down in Windows 11
cc @Korijn 

https://github.com/pygfx/pygfx/pull/215#issuecomment-991045696

> On Windows 11, cube.py (with glfw) and cube_qt.py (with pyside6) work. cube_wx.py acts weird; it shows up, renders the first frame, but only shows the rotation after I resize the window, but before I release the left mouse button.

almarklein:
Closes #209, and also adopts (configurable) fps limiting.

### DIFF
--- a/wgpu/gui/wx.py
+++ b/wgpu/gui/wx.py
@@ -3,6 +3,7 @@ Support for rendering in a wxPython window. Provides a widget that
 can be used as a standalone window or in a larger GUI.
 """
 
+import time
 import ctypes
 
 from .base import WgpuCanvasBase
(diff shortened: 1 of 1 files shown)


wgpu/gui/wx.py
```diff
@@ -28,37 +29,48 @@ def __init__(self, callback):
...
 
     def on_paint(self, event):
-        if self._draw_lock:
-            return
+        self._draw_time = time.perf_counter()
         dc = wx.PaintDC(self)  # needed for wx
-        self._draw_frame_and_present()
+        if not self._draw_lock:
+            self._draw_frame_and_present()
         del dc
```
Korijn, line 63: should this by in a try/finally maybe?

almarklein: The `_draw_frame_and_present` already does a try-except on the actual draw function:
https://github.com/pygfx/wgpu-py/blob/e838db358e68a95771fd1f6ce186b335c6722bf3/wgpu/gui/base.py#L99-L114

Korijn:
LGTM

PR #215, opened 2021-12-15
GLFW: max_fps

almarklein:
Add (configurable) FPS limiting to the GLFW backend.


Korijn:
Since I can't tell very easily from the code changes; it is also still possible to only render when requested explicitly by a user, right?

PR #216, opened 2021-12-15
QT: max_fps

almarklein:


Korijn:
Since I can't tell very easily from the code changes; it is also still possible to only render when requested explicitly by a user, right?

Otherwise, LGTM

almarklein:
Yes. The user application will use `canvas.request_draw` to request a new draw. If this happens occasionally, then there will occasionally be a draw. If the user code calls `canvas.request_draw` continuously (e.g. in the draw callback) then the limiting kicks in. There is not an explicit way to turn the limit off, but setting the `max_fps` to 999 should have the same effect.

PR #218, opened 2021-12-15
Jupyter: max_fps

almarklein:
Reaching 30 fps with the Jupyter widget would be impressive, but this makes the API consistent with the glfw canvas, and you can set the FPS to e.g. 5 or 10 if you want.

### DIFF
--- a/wgpu/gui/jupyter.py
+++ b/wgpu/gui/jupyter.py
@@ -3,6 +3,7 @@ Support for rendering in a Jupyter widget. Provides a widget subclass that
 can be used as cell output, or embedded in a ipywidgets gui.
 """
 
+import time
 import weakref
 import asyncio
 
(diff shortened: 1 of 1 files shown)


wgpu/gui/jupyter.py
```diff
@@ -69,7 +83,12 @@ def is_closed(self):
         return self._is_closed
 
     def _request_draw(self):
-        RemoteFrameBuffer.request_draw(self)
+        if not self._request_draw_timer_running:
+            now = time.perf_counter()
+            target_time = self._draw_time + 1.0 / self._max_fps
+            wait_time = max(0, target_time - now)
+            self._request_draw_timer_running = True
```
almarklein, line 90: The similarities in handling the max_fps are more similar than I anticipated. I've abstracted this away in the base class somewhat. Closing this PR in favor of a new one.

PR #219, opened 2021-12-15
Generic implementation of max fps

almarklein:
This puts most of the logic regarding limiting FPS into the base canvas class. The GUI implementations only have to take care of delaying the request for a new draw in the gui-specific way.

This PR also implements this for Jupyter.


Korijn:
LGTM! Tidies up nicely.

PR #221, opened 2021-12-17
Better glfw import error message

almarklein:
* Suppress the "during the handling of the above exception ...".
* Some more instructions for Linux.
* Deal with the fact that some glfw functions are not always available.


Korijn:
Does the python package for glfw not include the binaries?

almarklein:
I *think* that RPi is not compatible with the aarch64 platform. Anyway, when I pip installed glfw today it looked for a libglfw on my system. Also see https://github.com/FlorianRhiem/pyGLFW/pull/59

almarklein:
BTW: I also pushed some changes to deal with the fact that some libglfw implementations don't have all functions that we use.

PR #229, opened 2022-01-14
Qt auto gui

### LINKED ISSUE #226: Qt auto gui support

berendkleinhaneveld:
Add Qt as auto gui backend.

Closes #226.


wgpu/gui/auto.py
```diff
@@ -1,12 +1,13 @@
 """
 Automatic GUI backend selection.
 
-Right now we only chose between GLFW and Jupyter. We might add suport
-for e.g. Qt later. Or we might decide to stick with these two.
+Right now we only chose between GLFW, Qt and Jupyter. We might add suport
```
almarklein, line 4: My bad:
```suggestion
Right now we only chose between GLFW, Qt and Jupyter. We might add support
```

wgpu/gui/auto.py
```diff
@@ -27,9 +28,35 @@ def is_jupyter():
...
         from .glfw import WgpuCanvas, run, call_later  # noqa
-    except ImportError as err:
-        msg = str(err)
-        msg += "\n\n  Install glfw using e.g. ``pip install -U glfw``."
-        if sys.platform.startswith("linux"):
-            msg += "\n  You may also need to run the equivalent of ``apt install libglfw3``."
-        raise ImportError(msg) from None
+    except ImportError as glfw_err:
+        try:
+            for libname in ("PySide6", "PyQt6", "PySide2", "PyQt5", "PySide", "PyQt4"):
```
almarklein, line 33: I think we can drop PyQt4 and PySide because afaik they are not supported on py36. Or at least, I could not get them to install on CI the other day.

wgpu/gui/auto.py
```diff
@@ -27,9 +28,35 @@ def is_jupyter():
...
+            # application before any widget is created
+            app = QtWidgets.QApplication([])
+
+            def run():
+                app.exec() if hasattr(app, "exec") else app.exec_()
+
+            def call_later(delay, callback, *args):
+                QtCore.QTimer.singleShot(delay * 1000, lambda: callback(*args))
+
+        except ImportError as qt_err:
```
almarklein, line 52: Does this catch exceptions raised from `from .qt import ...`? I think we can move this part to the else clause of the try-except that uses `importlib.import_module(libname)`.

wgpu/gui/auto.py
```diff
@@ -27,9 +28,35 @@ def is_jupyter():
...
+
+            # When using Qt, there needs to be an
+            # application before any widget is created
+            app = QtWidgets.QApplication([])
+
+            def run():
+                app.exec() if hasattr(app, "exec") else app.exec_()
+
+            def call_later(delay, callback, *args):
+                QtCore.QTimer.singleShot(delay * 1000, lambda: callback(*args))
```
almarklein, line 50: I hope we can move `run()` and `call_later` in `qt.py`, so that people can do `from wgpu.gui.qt import WgpuCanvas, run`, e.g. to force using qt if glfw is available ;)

I think we need to do something smart with the instantiation of the application. Maybe, in the init of Qt's WGpuCanvas we can check whether an application exists, and we create it if not? Maybe there is a more elegant way?

BTW: Glfw has a similar issue, but people using glfw are not going to embed it into their glfw GUI app, so we can take control and initialize glfw.

wgpu/gui/qt.py
```diff
@@ -198,6 +260,175 @@ def close(self):
...
+        is a dict with at least the key event_type. For details, see
+        https://jupyter-rfb.readthedocs.io/en/latest/events.html
+        """
+        event_type = event.get("event_type")
+        for callback in self._event_handlers[event_type]:
+            callback(event)
+
+    def add_event_handler(self, *args):
+        """Register an event handler.
+
```
almarklein, line 289: We're tri-plicating this code now, let's put it in a base class ... in a next PR.

almarklein (commented):
Nice, looking good! A few minor comment and one somewhat harder issue wrt how/when to instantiate the QApplication.

wgpu/gui/auto.py
berendkleinhaneveld: Yes! I'll update this and the qt module as well.

wgpu/gui/auto.py
berendkleinhaneveld: Hmm, I see what you mean, but a check during instantiating seems a bit dirty.
What I like about using the `auto` module for instantiating the app is that there isn't too much cruft leaking into the `qt` module.
We could do something like: `from wgpu.gui.auto.qt import WgpuCanvas, run` for users that specifically would want to use Qt. That should make it explicit that the user wants to use the auto module with the qt framework and leave the `wgpu.gui.qt` module as is.

wgpu/gui/auto.py
berendkleinhaneveld: I didn't really get what you were after with this comment, but I think I managed to address your concern. I moved it to the else clause of the for loop instead though.

wgpu/gui/auto.py
berendkleinhaneveld: I've gone ahead and separated the auto gui functionality from the `wgpu.gui.qt` module into the `wgpu.gui.auto.qt` module. Hope you like it! We could also apply the same for glfw of course, but maybe in another PR, together with the de-tri-plication?

wgpu/gui/qt.py
berendkleinhaneveld: I've made #230 to track this!

examples/triangle_qt.py
```diff
@@ -2,18 +2,16 @@
...
 
 # For the sake of making this example Just Work, we try multiple QT libs
-try:
-    from PySide6 import QtWidgets
-except ModuleNotFoundError:
+for lib in ("PySide6", "PyQt6", "PySide2", "PyQt5"):
     try:
-        from PyQt6 import QtWidgets
+        QtWidgets = importlib.import_module(f"{lib}.QtWidgets")
+        break
```
almarklein, line 11: ❤️

wgpu/gui/auto/qt.py
```diff
@@ -0,0 +1,270 @@
...
+    int(QtCore.Qt.Key_Tab): "Tab",
+}
+
+
+class QAutoWgpuCanvas(WgpuCanvas):
+    def __init__(self, *args, **kwargs):
+        # When using Qt, there needs to be an
+        # application before any widget is created
+        global app
+        app = QtWidgets.QApplication.instance() or QtWidgets.QApplication([])
```
almarklein, line 91: A new application should only be created when `app` is still None.

almarklein: If you use this code, it should work also in `wgpu.gui.qt` ;) 

```py
global app
if QtWidgets.QApplication.instance() is None:
    QtWidgets.QApplication([])
if app is None:
    app = QtWidgets.QApplication.instance()
```

almarklein (commented):
> I didn't really get what you were after with this comment, but I think I managed to address your concern. I moved it to the else clause of the for loop instead though.

This is what I meant :)


TBH I'm not a fan of the `wgpu.gui.auto.qt` approach. I think it may make things complex (two places to get a Qt canvas).

berendkleinhaneveld:
> TBH I'm not a fan of the `wgpu.gui.auto.qt` approach. I think it may make things complex (two places to get a Qt canvas).

I was writing this big chunk of text about how it should be possible to have at least a basic widget available for users who want to incorporate a wgpu canvas in their application, but after reading the `wgpu.gui.qt` module a bit better, I believe that that is exactly what `QWgpuWidget` is for? So I guess I am not that opposed to the idea anymore of incorporating all these event handler methods right into `wgpu.gui.qt` :) I'll update accordingly!

wgpu/gui/auto.py
berendkleinhaneveld: Ok, reverted the creation of `wgpu.gui.auto.qt`!

wgpu/gui/auto/qt.py
berendkleinhaneveld: I've added the check for `app is None`, good call.

wgpu/gui/qt.py
```diff
@@ -147,12 +216,19 @@ class QWgpuCanvas(WgpuCanvasBase, QtWidgets.QWidget):
     # detect this. See https://github.com/pygfx/wgpu-py/pull/68
 
     def __init__(self, *, size=None, title=None, max_fps=30, **kwargs):
+        # When using Qt, there needs to be an
+        # application before any widget is created
+        global app
+        if app is None:
+            app = QtWidgets.QApplication.instance() or QtWidgets.QApplication([])
```
almarklein, line 223: We could also create a function:
```py
def get_app():
   return QtWidgets.QApplication.instance() or QtWidgets.QApplication([])
```

And call that here and in `run()`. That way we don't need that global variable, and `run()` would also work if no widgets are created yet (e.g. in a use case that dynamically creates canvases).

almarklein:
>  I believe that that is exactly what QWgpuWidget is for? 

Yes, that was recently split in two, to have a clearer distinction between using a toplevel canvas or a widget to embed.

almarklein (approved):
nice!

PR #231, opened 2022-01-19
Fix glfw window size on mac

almarklein:
The windows for GLFW are HUGE! Like 4x too big, which can be explained because we probably did a division where we should have done a multiplication. Also see the comment a few lines up.


almarklein:
@berendkleinhaneveld did you experience this too?

berendkleinhaneveld:
Yeah, had the same problem

PR #237, opened 2022-01-21
Use same lazy init approach in glfw as in qt

### LINKED ISSUE #232: Delay glfw.init, like we now do for qt

almarklein:
Closes #232


wgpu/gui/glfw.py
```diff
@@ -548,8 +549,11 @@ def _on_key(self, window, key, scancode, action, mods):
 WgpuCanvas = GlfwWgpuCanvas
 
 
-# We initialize glfw upon import
-glfw.init()
+def ensure_app():
+    # Is is safe to call init multiple times:
```
Korijn, line 553:
```suggestion
    # It is safe to call init multiple times:
```

Korijn:
LGTM!

PR #238, opened 2022-01-22
Run examples offscreen as part of test suite

### LINKED ISSUE #233: Run examples on CI and fail on error
- [ ] Adjust auto gui mechanism to switch to CI mode based on environment variable `CI=true` condition
- [ ] If so, use an OffscreenCanvas and adjust `run()` to render a single frame
- [ ] Ensure CI build fails on error running examples

### LINKED ISSUE #235: Screenshot verification of examples
- [x] Compare screenshot against "known good screenshot" on CI as additional acceptance criteria
- [x] Include command to regenerate the ground truth screenshots

Korijn:
Closes #233 
Closes #235

By setting the environment variable `WGPU_FORCE_OFFSCREEN=true` the auto gui mechanism will always pick `WgpuManualOffscreenCanvas`.

I've added minimal event loop support to the offscreen canvas so you any code expecting a GUI (our examples, notably) can still be executed according to the same abstraction.

I had to tweak the triangle example to support it.

Not sure if I'm happy about the mechanism that selects examples to include in the test suite, but it does work!

Double checked that CI also actually runs and passes the new tests:

image


wgpu/gui/_offscreen.py
```diff
@@ -8,6 +8,12 @@ class WgpuOffscreenCanvas(WgpuCanvasBase):
     The resulting texture view is passed to the ``present()`` method.
     """
 
+    def __init__(self, *args, **kwargs):
+        # we don't pass on constructor arguments; code may
+        # be assuming we are relying on e.g. qt or wx or jupyter
+        # but in the offscreen case we are definitely not
```
Korijn, line 14: I needed to do this because some examples pass e.g. `title="cube example"` which obviously isn't supported by `object.__init__()` which is the base class of offscreen canvases.

tests/test_examples.py
```diff
@@ -0,0 +1,48 @@
...
+
+if not can_use_wgpu_lib:
+    pytest.skip("Skipping tests that need the wgpu lib", allow_module_level=True)
+
+
+examples_dir = Path(__file__).parent.parent / "examples"
+examples_to_test = []
+for example_path in examples_dir.glob("*.py"):
+    example_code = example_path.read_text()
+    if "wgpu.gui.auto" in example_code and "canvas = WgpuCanvas(" in example_code:
```
Korijn, line 24: it's a little dumb but it works. any ideas to improve this (without blowing up the effort)?

ivoflipse: A fixture that's only true for those tests? If necessary using a separate test run with difference arguments

Korijn: Well I mean the detection mechanism. What is a good way to determine if an example can be tested or not? Not all of them leverage the auto gui mechanism.

Korijn: I'm thinking of adding a marker comment to exclude an example and just run every example that doesn't have it.

tests/test_gui_auto_offscreen.py
```diff
@@ -0,0 +1,49 @@
...
+    from wgpu.gui.auto import WgpuCanvas
+    from wgpu.gui.offscreen import WgpuManualOffscreenCanvas
+
+    assert WgpuCanvas is WgpuManualOffscreenCanvas
+
+
+def test_event_loop():
+    """Check that the event loop handles queued tasks and then returns.
+    TODO: This test may hang indefinitely in the failure case, can we
+    prevent that?"""
```
almarklein, line 37: Maybe something like this?
```py
def stopper():
    asyncio.get_running_loop().stop()

loop = asyncio.get_event_loop()
loop.call_later(5, stopper)
```

examples/triangle.py
```diff
@@ -143,11 +145,9 @@ def draw_frame():
     return device
 
 
-if __name__ == "__main__":
+canvas = WgpuCanvas(size=(640, 480), title="wgpu triangle")
+main(canvas)
```
almarklein, line 149: I think this may break the tringle_xx.py examples.

Korijn: Right... Will have another look.

tests/test_gui_auto_offscreen.py
Korijn: Then it's guaranteed to stop, but that's a property I want to test :) you see the problem here?

almarklein: Well, you could test that the loop has not stopped by the stopper, by making it set a flag and then asserting that the flag was not set, I think?

berendkleinhaneveld: We could use the decorator of the pytest-timeout plugin?

tests/test_examples.py
Korijn: Added an including flag.

tests/test_gui_auto_offscreen.py
Korijn: Decided to just leave it for now...

examples/triangle.py
Korijn: Reverted. Only the cube example is tested now.

Korijn:
This is ready as far as I am concerned. Added an explanation of how to work with this setup in `README.md`. Hope you like it.

.github/workflows/ci.yml
```diff
@@ -102,6 +102,15 @@ jobs:
     - name: Test on repo
       run: |
           pytest -v tests
+    - name: Regenerate screenshots for inspection
+      if: failure()
+      run: |
+          pytest -v tests -k test_examples --regenerate-screenshots
+    - uses: actions/upload-artifact@v2
+      if: always()
```
almarklein, line 110: This only needs to run if the above runs, right? So only on failure?

tests/test_examples.py
```diff
@@ -0,0 +1,74 @@
...
+    if pytestconfig.getoption("regenerate_screenshots"):
+        imageio.imwrite(screenshot_path, img)
+
+    # if a reference screenshot exists, assert it is equal
+    if screenshot_path.exists():
+        stored_img = imageio.imread(screenshot_path)
+        assert img.dtype == np.uint8
+        assert img.dtype == stored_img.dtype
+        # allow slight color deviations
+        assert np.allclose(stored_img, img, atol=1)
```
almarklein, line 74: Would be nice to also export diff images when it fails. We could do it in a later PR.

README.md
```diff
@@ -115,3 +115,23 @@ This code is distributed under the 2-clause BSD license.
...
+
+Include the comment `# test example: true` in an example to have pytest run it as part of the test suite.
+
+To support testing an example, ensure the following requirements are met:
+
+* The `WgpuCanvas` class is imported from the `wgpu.gui.auto` module.
+* The `canvas` instance is exposed as a global in the module.
+
+By default the test will simply verify that the example can be executed without an error, and that an
+image can be rendered using the canvas.
```
almarklein, line 129: And if a reference image is available it will be compared against that.

tests/test_examples.py
```diff
@@ -0,0 +1,74 @@
...
+    # assert something was rendered
+    assert img is not None and img.size > 0
+
+    # regenerate screenshot if requested
+    screenshot_path = screenshots_dir / f"{module}.png"
+    if pytestconfig.getoption("regenerate_screenshots"):
+        imageio.imwrite(screenshot_path, img)
+
+    # if a reference screenshot exists, assert it is equal
+    if screenshot_path.exists():
```
almarklein, line 69: Small note regarding pygfx: I'm a bit worried what this'll do for the repo size, because examples may change (e.g. tweak the camera pos). Though if we restrict this to the validation examples (and don't use auto-camera tricks) then it should be fine.

examples/cube.py
```diff
@@ -1,5 +1,7 @@
 """
 This example renders a simple textured rotating cube.
+
+# test example: true
```
almarklein, line 4: The comment is in the docstring, is this intended?

Nitpick, but `test_example` (with an underscore) feels more natural because it reads a bit like a variable.

Korijn:
Python 3.10:

```
tests/test_gui_auto_offscreen.py::test_event_loop
  /home/runner/work/wgpu-py/wgpu-py/wgpu/gui/offscreen.py:80: DeprecationWarning: There is no current event loop
    loop = asyncio.get_event_loop()
```

almarklein (commented):
Nice! Just a few comments.

Can we also test the triangle example, by making a new module that imports from `triangle.py` but which does have the canvas in the main namespace?

tests/test_examples.py
Korijn: Later PR please :)

README.md
Korijn: That's what it says in the next couple of lines right?

tests/test_examples.py
Korijn: Well, it's an opt-in mechanism. We can also lower the resolution. I'm not worried though,

examples/cube.py
Korijn: Not intended. I just need something and I don't really care what it looks like, so in this case I'll just defer to your judgement. :)

wgpu/gui/offscreen.py
```diff
@@ -62,3 +64,18 @@ def draw(self):
 
 
 WgpuCanvas = WgpuManualOffscreenCanvas
+
+
+def call_later(delay, callback, *args):
+    loop = asyncio.get_event_loop()
```
almarklein, line 70: They seem to want to get rid of `get_event_loop` in favor of `get_running_loop` and `run`.
```suggestion
    if hasattr(asyncio, "get_running_loop"):
        loop = asyncio.get_running_loop()
    else:  # py3.6 or lower
        loop = asyncio.get_event_loop()
```

Korijn: Actually the equivalent seems to be `asyncio.get_event_loop_policy().get_event_loop()`

wgpu/gui/offscreen.py
```diff
@@ -62,3 +64,18 @@ def draw(self):
...
+
+
+async def mainloop_iter():
+    pass  # no op
+
+
+def run():
+    """Handle all tasks scheduled with call_later and return."""
+    loop = asyncio.get_event_loop()
+    loop.run_until_complete(mainloop_iter())
```
almarklein, line 81:
```suggestion
    if hasattr(asyncio, "run"):
        asyncio.run(mainloop_iter())
    else:  # py36 and lower
        loop = asyncio.get_event_loop()
        loop.run_until_complete(mainloop_iter())
```

README.md
Korijn: That's sort of what I mean by "enable", but I'll try and reword it

> If you additionally want to enable screenshot verification, you can generate the reference screenshot
like so:

README.md
```diff
@@ -115,3 +115,23 @@ This code is distributed under the 2-clause BSD license.
 
 * Use the optional arguments to `python download-wgpu-native.py --help` to download a different version of the upstream wgpu-native binaries.
 * The file `wgpu/resources/wgpu_native-version` will be updated by the script to track which version we depend upon.
+
+### Example screenshot verification
+
+Include the comment `# test example: true` in an example to have pytest run it as part of the test suite.
```
almarklein, line 121:
```suggestion
Include the comment `# test_example: true` in an example to have pytest run it as part of the test suite.
```

PR #239, opened 2022-01-24
Implement WgpuAutoGui mixin to avoid duplicating code

### LINKED ISSUE #230: Fix tri-plication of event handler code in auto module

almarklein:
Closes #230

### DIFF
--- a/wgpu/gui/base.py
+++ b/wgpu/gui/base.py
@@ -178,8 +178,7 @@ class WgpuCanvasBase(WgpuCanvasInterface):
 
 
 class WgpuAutoGui:
-    """ Mixin class for canvases implementing autogui.
-    """
+    """Mixin class for canvases implementing autogui."""
 
     def handle_event(self, event):
         """Handle an incoming event.
--- a/wgpu/gui/base.py
+++ b/wgpu/gui/base.py
@@ -175,3 +175,73 @@ class WgpuCanvasBase(WgpuCanvasInterface):
         the FPS is limited to avoid draining CPU and power.
         """
         raise NotImplementedError()
+
+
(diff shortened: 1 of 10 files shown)


wgpu/gui/base.py
```diff
@@ -175,3 +175,72 @@ def _request_draw(self):
...
+    def handle_event(self, event):
+        """Handle an incoming event.
+
+        Subclasses can overload this method. Events include widget
+        resize, mouse/touch interaction, key events, and more. An event
+        is a dict with at least the key event_type. For details, see
+        https://jupyter-rfb.readthedocs.io/en/latest/events.html
+        """
+        event_type = event.get("event_type")
+        for callback in self._event_handlers[event_type]:
```
Korijn, line 192: Are you ok with assuming `self._event_handlers` exists?

wgpu/gui/offscreen.py
```diff
@@ -3,9 +3,10 @@
 import numpy as np
 
 from ._offscreen import WgpuOffscreenCanvas
+from .base import WgpuAutoGui
 
 
-class WgpuManualOffscreenCanvas(WgpuOffscreenCanvas):
+class WgpuManualOffscreenCanvas(WgpuAutoGui, WgpuOffscreenCanvas):
```
Korijn, line 9: I don't think this one has `self._event_handlers` now

wgpu/gui/base.py
almarklein: No, I'm not ;)

wgpu/gui/offscreen.py
almarklein: Good catch.

PR #241, opened 2022-02-02
Glfw mouse position fix

almarklein:
This is (again) due to the fact that the meaning of screen space depends on the platform. We had a trick to detect the meaning, and in this PR we use the result to fix the pos in mouse events.

~Would be good if someone on Windows/Linux with a hidpi screen could check that the positioning still works well there. E.g using pygfx's picking_points.py example.~ Done :)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -118,6 +118,7 @@ jobs:
       if: matrix.PYINSTALLER == 1
       run: |
           pip install psutil pyinstaller>=4
+          pyinstaller --version
           pytest -v wgpu/__pyinstaller
 
   # The release builds are done for the platforms that we want to build wheels for.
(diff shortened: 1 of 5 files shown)


Korijn:
Jaraco problem looks like a recent change in setuptools actually: https://github.com/pypa/setuptools/blob/main/CHANGES.rst

almarklein:
I posted an issue: https://github.com/pyinstaller/pyinstaller/issues/6564  Let's merge this as-is.

Korijn:
You can mark the test with xfail. That way CI will pass and we will know when the outcome changes again later

almarklein:
I set the GH actions `continue-on-error`. Or did you mean the pytest xfail? :P   I think either way the error is hidden, right?

Korijn:
Let's go with pytest; we don't want to ignore _all_ errors!

almarklein:
>  we don't want to ignore all errors!

No, I only applied it to the pyinstaller step.

almarklein:
Looks like a fix is underway in PyInstaller. Let's hold on for a bit.

PR #244, opened 2022-03-14
Event system

berendkleinhaneveld:
Event system with the following concepts:

* Event (KeyboardEvent, PointerEvent, etc.)
* EventTarget
* EventDispatcher

(will update description later)

- [ ] tests
- [ ] docs (strings)
- [ ] example

### DIFF
--- a/wgpu/gui/events.py
+++ b/wgpu/gui/events.py
@@ -1,0 +1,74 @@
+from collections import defaultdict
+
+
+class EventTarget:
+    """Mixin class for canvases implementing autogui."""
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self._event_handlers = defaultdict(set)
+
+    def handle_event(self, event):
+        """Handle an incoming event.
+
+        Subclasses can overload this method. Events include widget
+        resize, mouse/touch interaction, key events, and more. An event
+        is a dict with at least the key event_type. For details, see
+        https://jupyter-rfb.readthedocs.io/en/latest/events.html
+        """
+        event_type = event.get("event_type")
+        for callback in self._event_handlers[event_type]:
+            callback(event)
+
+    def add_event_handler(self, *args):
+        """Register an event handler.
+
+        Arguments:
+            callback (callable): The event handler. Must accept a
+                single event argument.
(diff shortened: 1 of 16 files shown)


wgpu/gui/events.py
```diff
@@ -0,0 +1,236 @@
...
+                    # Apparently the set_pointer_capture was called for this
+                    # event.pointer_id, so return immediately
+                    return
+            if not event.bubbles:
+                break
+            target = getattr(target, "parent", None)
+
+        # The EventDispatcher itself does not have to be part
+        # of the hierarchy so we'll handle the event separately
+        super().handle_event(event)
```
almarklein, line 236: Does this need to be wrapped in an `if not event.bubbles`?

wgpu/gui/jupyter.py
```diff
@@ -48,7 +67,8 @@ def handle_event(self, event):
             self._pixel_ratio = event["pixel_ratio"]
             self._logical_size = event["width"], event["height"]
 
-        super().handle_event(event)
+        ev = EVENT_TYPE_MAP(event_type)(event_type, **event)
```
almarklein, line 70: Interesting how little change is needed here :)

If we go down this path, we should probably update the jupyter_rfb spec to say that events may be dict-like.

wgpu/gui/events.py
```diff
@@ -0,0 +1,236 @@
...
+                return
+            else:
+                target.handle_event(event)
+                if pointer_id and pointer_id in EventTarget.pointer_captures:
+                    # Apparently the set_pointer_capture was called for this
+                    # event.pointer_id, so return immediately
+                    return
+            if not event.bubbles:
+                break
+            target = getattr(target, "parent", None)
```
almarklein, line 232: This looks pygfx-agnostic at first glance, but it does assume the name 'parent' ... is this a sign that we're too opinionated here?

almarklein (commented):
I made some minor comments. Looks like its already in a good shape technically.

From a high-level perspective, this system feels quite specific, even though wgpu-py offers this as tool. I wonder if we can keep things more generic here, and move the specifics to pygfx. This is more of a guts feeling - I say that without having thought what that could look like ...

PR #247, opened 2022-03-23
Apply rate limits for events

almarklein:
Closes https://github.com/pygfx/pygfx/issues/264

The `WgpuAutoGui` implements a method `_handle_event_rate limited()` - as opposed to `handle_event()` - that can be used to rate limit events.

This also refactors the exception logging mechanism to make it generally usable, as a context manager. It was previously only used for the draw handler. Now it's used for events, and we could use it in pygfx.

### DIFF
--- a/wgpu/gui/base.py
+++ b/wgpu/gui/base.py
@@ -198,12 +198,18 @@ class WgpuAutoGui:
         self._event_handlers = defaultdict(set)
 
     def _get_event_wait_time(self):
+        """Calculate the time to wait for the next event handling (for rate-limited events)."""
         rate = 75  # events per second
         now = time.perf_counter()
         target_time = self._last_event_time + 1.0 / rate
         return max(0, target_time - now)
 
     def _handle_event_rate_limited(self, ev, call_later_func, match_keys, accum_keys):
+        """Alternative `to handle_event()` for events that must be rate-limted.
+        If any of the `match_keys` keys of the new event differ from the currently
+        pending event, the old event is dispatched now. The `accum_keys` keys of
+        the current and new event are added together (e.g. to accumulate wheel delta).
+        """
         event_type = ev["event_type"]
         # We may need to emit the old event. Otherwise, we need to update the new one.
         old = self._pending_events.get(event_type, None)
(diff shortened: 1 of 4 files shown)


wgpu/gui/base.py
```diff
@@ -184,8 +193,38 @@ class WgpuAutoGui:
 
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
+        self._last_event_time = 0
+        self._pending_events = {}
         self._event_handlers = defaultdict(set)
 
+    def _get_event_wait_time(self):
+        """Calculate the time to wait for the next event handling (for rate-limited events)."""
+        rate = 75  # events per second
```
Korijn, line 202: Can we make this configurable?

wgpu/gui/jupyter.py
```diff
@@ -48,6 +48,8 @@ def handle_event(self, event):
             self._pixel_ratio = event["pixel_ratio"]
             self._logical_size = event["width"], event["height"]
 
+        # No need to rate-limit the pointer_move and wheel events;
+        # they're already rate limited by jupyter_rfb in the client.
```
Korijn, line 52: Really? To the same rate?

almarklein: No, not the same. Probably less, and may depend on the connection too.

Korijn:
Whew, looks pretty complicated. Do I understand correctly that you may be forced to emit events if pointer down/up events are among the stream of move events?

almarklein:
> Do I understand correctly that you may be forced to emit events if pointer down/up events are among the stream of move events?

Yes. If another event is in the stream, then the pending event is emitted, to make sure that the original order of events is preserved.

almarklein:
In other words, pending rate-limited events are dispatch (i.e. `_dispatch_pending_events` is called) in the following cases:
* When the timer runs out.
* When a non-rate-limited event is dispatched.
* When a rate-limited event of the same time is scheduled that has certain keys different (e.g. modifiers changes).

Korijn:
> In other words, pending rate-limited events are dispatch (i.e. `_dispatch_pending_events` is called) in the following cases:
> 
> * When the timer runs out.
> * When a non-rate-limited event is dispatched.
> * When a rate-limited event of the same time is scheduled that has certain keys different (e.g. modifiers changes).
> 
> Maybe I should add this in the comments.

If you can do that, I think we can merge :)

PR #252, opened 2022-04-08
Fix: "Unable to present swap chain: AlreadyAcquired"

panxinmiao:


almarklein:
Under what circumstances do you get this error? I have not seen it before.

panxinmiao:
> Under what circumstances do you get this error? I have not seen it before.

When running the "cube. py" example, this error occurs when I minimize the window:

`Requested size 1x1 is outside of the supported range: Extent3d { width: 0, height: 0, depth_or_array_layers: 1 }..=Extent3d { width: 0, height: 0, depth_or_array_layers: 1 }
thread '<unnamed>' panicked at 'called `Option::unwrap()` on a `None` value', src\device.rs:696:10`

 I added a check to the beginning of the `draw_frame` method, like this:

      w, h = canvas.get_logical_size()
      if w == 0 or h == 0:
          return

Run again, and this error will occur when the window is minimized

or:
Call `draw_frame()` directly instead of use `canvas.request_draw(draw_frame)`,  this error also occurs.

almarklein:
The CI is complaining because of formatting issues. Could you please run  `black .`? Other than that this looks good to me!

panxinmiao:
> The CI is complaining because of formatting issues. Could you please run `black .`? Other than that this looks good to me!

Sure.

Korijn:
Just wondering (don't mean to block this PR): does this mean the renderer continues to draw even when the window is minimized?

almarklein:
No, the GUI toolkit *should* not invoke new draws until the widget is actually visible. I'm pretty sure Qt does this correctly, but might be worth checking for glfw.

PR #253, opened 2022-04-08
Fix: "RuntimeError: Set changed size during iteration"

panxinmiao:


wgpu/gui/base.py
```diff
@@ -256,7 +256,7 @@ def _dispatch_pending_events(self):
     def _dispatch_event(self, event):
         """Dispatch event to the event handlers."""
         event_type = event.get("event_type")
-        for callback in self._event_handlers[event_type]:
+        for callback in list(self._event_handlers[event_type]):
```
Korijn, line 259: I would suggest creating a shallow copy using `.copy()` instead, that way the original collection type is preserved
```suggestion
        for callback in self._event_handlers[event_type].copy():
```

almarklein:
Thanks!

PR #254, opened 2022-04-08
'QWgpuWidget' adds support for events, and delegate events of QWgpuCanvas' to subwidgets

panxinmiao:
So, can use "OrbitControls" , etc. on "QWgpuWidget"

### DIFF
--- a/wgpu/gui/qt.py
+++ b/wgpu/gui/qt.py
@@ -114,7 +114,7 @@ def enable_hidpi():
 enable_hidpi()
 
 
-class QWgpuWidget(WgpuCanvasBase, QtWidgets.QWidget):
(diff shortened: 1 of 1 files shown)


Korijn:
Looks like all you did was move things around and add a subclass, correct? Or did you also change the implementation of some methods?

panxinmiao:
> Looks like all you did was move things around and add a subclass, correct? Or did you also change the implementation of some methods?

Yes, there is no change in any method implementation, only moved, so "QWgpuWidget" can also use "AutoGui" events

Korijn:
In that case I think this is OK, @almarklein or @berendkleinhaneveld would you like to test this before we merge, perhaps?

berendkleinhaneveld:
LGTM

almarklein:
The diff is a bit hard to read (not your fault) but looks like a simple change indeed. LGTM! Thanks!

PR #258, opened 2022-04-12
More example-tests

almarklein:
* Add a testable variant of the triangle example.
* Tweak the compute example so it will fail on wrong output.


almarklein:
Ready!

Issue #261, opened 2022-04-17
examples need update, not working


pieper:
I am able to pip install wgpu, but the examples are not working because they use the `@` syntax rather than the `[[ ]]` syntax.  After changing to the `[[ ]]` in the `compute-noop.py` example it runs, but then gives these errors and then aborts.

```
Unexpected resource type: Array { base: [1], size: Dynamic, stride: 4 }
Unexpected resource type: Array { base: [1], size: Dynamic, stride: 4 }
```
this is with:
```
>>> import wgpu
>>> wgpu.__version__
'0.7.7'
```

Korijn:
Try grabbing the examples from git tag v0.7.7.

Note that in general we can't guarantee examples from the tip of the master branch will work with the last release. For example, right now we are transitioning to a newer version of wgpu, so the shader syntax is being changed.

If you really want to work with the master branch, then you should not install from pip, and instead check out the "Developers" section from the README.

pieper:
Ah, great, makes sense that does work.

Thanks for the quick response!

PR #262, opened 2022-04-17
Add note about example version in repo vs pypi

pieper:
Add a note so that people don't get confused (like [I did](https://github.com/pygfx/wgpu-py/issues/261)) by trying to use the current main branch examples with the pip installable release version.


almarklein:
Maybe even better is to add this to a (new) README.md in the examples dir. Then it's visible no matter how you got there, and we can say something like "you may want to select the appropriate release at the top".

Another option would be to include the examples in the docs. Then they match the version of the docs that you're watching. This is more work though.

almarklein:
Thanks!

PR #271, opened 2022-04-29
Fix that handle_event cannot be overloaded to handle move and wheel events

### LINKED ISSUE #269: "pointer_move" event seems not working
The "pointer_move" and "wheel" events work in the previous versions (<= 0.7.6), but they seem not working anymore in the new version 0.8.0. For example,  in the events.py example, changing the code
```py
class MyCanvas(WgpuCanvas):
    def handle_event(self, event):
        if event["event_type"] == "pointer_move":
            print(event)
```
gives no result. 

But it seems working when using
```py
canvas.add_event_handler(mouse_event, "pointer_move", "wheel")
```
There are some issues by setting events via subclass.

almarklein:
Fixes #269

### DIFF
--- a/wgpu/gui/base.py
+++ b/wgpu/gui/base.py
@@ -247,11 +247,12 @@ class WgpuAutoGui:
 
     def _dispatch_pending_events(self):
         """Handle any pending rate-limited events."""
-        events = self._pending_events.values()
-        self._last_event_time = time.perf_counter()
-        self._pending_events = {}
-        for ev in events:
-            self._dispatch_event(ev)
+        if self._pending_events:
+            events = self._pending_events.values()
+            self._last_event_time = time.perf_counter()
+            self._pending_events = {}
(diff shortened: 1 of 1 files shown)


wgpu/gui/base.py
```diff
@@ -247,11 +247,12 @@ def handle_event(self, event):
...
-        self._last_event_time = time.perf_counter()
-        self._pending_events = {}
-        for ev in events:
-            self._dispatch_event(ev)
+        if self._pending_events:
+            events = self._pending_events.values()
+            self._last_event_time = time.perf_counter()
+            self._pending_events = {}
+            for ev in events:
+                self.handle_event(ev)
```
almarklein, line 255: Two changes really:
* Wrap this up in a `if self._pending_events:`
* Call `self.handle_event(ev)` instead of `self._dispatch_event(ev)`

PR #273, opened 2022-05-02
Ability to turn vsync off

### LINKED ISSUE #256: FPS is always limited
For some reason it's impossible to push FPS over screen refresh rate (commonly 30 or 60).

See pygfx/pygfx#278 and pygfx/pygfx#254

almarklein:
Closes #256.

Turn vsync off when max_fps is high. This could also be a separate prop, but I suppose they sort of go hand in hand.


almarklein:
Maybe allow `max_fps=0/False`

Korijn:
Why can't the settings be independent?

almarklein:
Yeah ... better make it separate.

Korijn (approved):

Issue #171, opened 2021-06-14
How not to panic


marcdownie:
I've asked a very similar question before — here #108 . This is rightfully closed: error messages are _vastly_ better than they were a year ago. However, stepping back a bit, I see the wgpu still happily delivers these messages while issuing a Rust-level panic — that is, printing something and swiftly killing the whole process. 

Has this become the long term plan? or is there a way to recover from these panics? is the intention that our wgpu-py renderers are separate, easily re-initializable disposable processes? Is this what `triangle_subprocess.py` is aiming towards? Is wgpu[-py] for cattle rather than pets?

I'm asking this question here and not upstream because, while this behavior _very_ Rust, it's really not very Python at all — imagine writing wgpu-py from the REPL, or live-coding something built on wgpu-py from Jupyter kernel etc. Now that we have a WSL --- a shader language that we can pass along to wgpu as strings (without recourse to an external shader compiler) --- we dare not really use it dynamically, because a typo in a shader quits our process with:

```
Failed to parse WGSL code for Some(""): unknown type: `vec`
thread '<unnamed>' panicked at 'Parsing', src/device.rs:112:5
```
 
It speaks to what long-term applications of wgpu-py apps that are viable inside the python ecosystem. And it's also completely different set of rules from, say both OpenGL and WebGL --- which are perfectly recoverable from an error state.

almarklein:
Good point! For the time being, things should be pretty safe from the moment that all your shaders are correct (and you don't dun out of GPU memory). For dynamic use these panics are indeed not productive at all.

Eventually wgpu will be used from the browser, which is also very dynamic - and I'm pretty sure they won't kill the tab whenever there is a shader error :)  So in the long run I think there won't be any panics ever. For the time being we'll have to live with them, and during that time wgpu-py will be less suited for interactively writing shaders.

almarklein:
I've asked upstream to get some more info: https://github.com/gfx-rs/wgpu-native/issues/113

Korijn:
So to clearly answer this question in particular:

> Has this become the long term plan?

It seems very very unlikely!

However, full disclosure, we, the maintainers of wgpu-py, have to delegate your question upstream, so please keep an eye on the thread @almarklein mentioned.

Vipitis:
I would love to know the progress on this. From the linked issue it looks like there has been some progress.
A python `try:` can't catch these panics so crashes instead. But the error messages look really helpful for my use case, so if there is a way to get these into python as exception for example.
[...]

PR #278, opened 2022-05-31
Improve error reporting on WGSL

panxinmiao:
The current wgsl error log is not very convenient to read, especially when the shader code is long and complex, the current error prompt is difficult to locate the problem in a short time.

This PR improves the WGSL error prompt, and visually displays the error code and the line number and position of the code.

Looks something like this:
```
Shader Error: label:  Some("")
error: invalid field accessor `texcoord1`


     ┌─ wgsl:34:51
     │
  34 │     let value = textureSample(r_tex, r_sampler, in.texcoord1).r;
     │                                                    ^^^^^^^^^ invalid accessor



thread '<unnamed>' panicked at 'assertion failed: `(left == right)`
  left: `0`,
 right: `1`: ShaderModule[0] is no longer alive', C:\Users\runneradmin\.cargo\git\checkouts\wgpu-53e70f8674b08dd4\988990c\wgpu-core\src\hub.rs:169:9
```

It shows the position of the error code is line 34, column 51, and is underlined. the cause is `invalid field accessor texcoord1`

### DIFF
--- a/wgpu/backends/rs.py
+++ b/wgpu/backends/rs.py
@@ -33,6 +33,7 @@ import os
 import sys
 import ctypes
 import logging
+import re
 import ctypes.util
 from weakref import WeakKeyDictionary
 from typing import List, Dict
@@ -563,6 +564,15 @@ class GPUAdapter(base.GPUAdapter):
 
 
 class GPUDevice(base.GPUDevice, GPUObjectBase):
+    __shader_error_tmpl = re.compile(
+        r'Parsing\(ShaderError { source: "(.*)", label:(.*), inner: ParseError (.*) }\)',
+        re.M | re.S,
+    )
+    __inner_error_tmpl = re.compile(
+        r'{ message: "(.*)", labels: \[\((\d+)\.\.(\d+), "(.*)"\)\], notes: \[(.*)\] }',
+        re.M | re.S,
+    )
+
     def __init__(self, label, internal, adapter, features, limits, queue):
         super().__init__(label, internal, adapter, features, limits, queue)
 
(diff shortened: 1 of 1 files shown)


almarklein:
This is awesome! A year or so ago the application would just segfault if there was an error in the shader, so I was already happy to see *some* context, but this just pinpoints to the exact spot!

I also did a quick test and it worked great for what I tried.

Could you please refactor this into a function (e.g. `parse_wgpu_error(message) -> better_message`) and put it in rs_helpers? That way the rs.py does not become much larger than it already is, and we can write tests for that function (because the formatting that the regexpresses depend on might change).

panxinmiao:
> and we can write tests for that function (because the formatting that the regexpresses depend on might change).
 
I am not very familiar with `pytest`. I am not sure whether it is possible to do unit test by directly calling the `device.create_shader_module()` method and capturing the logs in the error output stream.

I tried. It seems that the test will be interrupted due to a fatal Python error.

almarklein:
> I am not sure whether it is possible to do unit test by directly calling the `device.create_shader_module()` method and capturing the logs in the error output stream.

My point is that now we can test `parse_wgpu_shader_error()` directly without going through wgpu :) Could you add a quick test for that? Maybe just with a fake message, but that at least makes it clear what the code expects from the error message (in terms of what words being present etc.)

almarklein:
Nice work!

I'm not sure whether we can fix #171. This solves one case, but there might be other cases where we still cause a panic.

Vipitis:
hey, I have chased a panic caused by incorrect shadercode (using a variable or function that doesn't exist). To this line just above the error handling https://github.com/pygfx/wgpu-py/blob/main/wgpu/backends/rs.py#L1265 
Other errors get handled as expect like syntax errors. I haven't built a minimal test cases yet and dug into cffi myself. As I have been using the Shadertoy class, the broken test case is in here: https://huggingface.co/spaces/Vipitis/shadermatch/blob/main/tests.py

Do you perhaps know a quick reason or solution?

almarklein:
It seems to fail at the same point where normally the textual error is generated (if the shader code is faulty). Although they are getting more rare over time, there are still cases where wgpu-core panics instead of producing an error. 

If you're familiar with compiling a Rust project, you could use Naga (https://github.com/gfx-rs/naga) to check the shader, and hopefully it produces something more friendly.

PR #279, opened 2022-06-02
Improve error reporting on WGSL again

panxinmiao:
Related: #278 

This PR fixes the problem that the error position indication is not accurate when the shader code contains a tab or \t symbol.

Add possible error notes

Add ANSI colors to make the results more readable.

Now, it looks something like this:
image

The test case has also been updated.

### DIFF
--- a/tests/test_rs_basics.py
+++ b/tests/test_rs_basics.py
@@ -5,10 +5,10 @@ import random
 import ctypes
 import sys
 import tempfile
+import re
 
 import wgpu.utils
 import wgpu.backends.rs
-import wgpu.backends.rs_helpers
 import numpy as np
 
 from testutils import run_tests, can_use_wgpu_lib, is_ci, iters_equal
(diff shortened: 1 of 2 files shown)


tests/test_rs_basics.py
```diff
@@ -659,7 +659,13 @@ def test_write_texture2():
 
 def test_parse_shader_error(caplog):
 
-    # invalid attribute access
+    device = wgpu.utils.get_default_device()
+
```
Korijn, line 663: I would recommend to use the following function to determine if the terminal you are outputting to supports colors. Many terminals don't!

```python
def terminal_supports_colors():
    term_isatty = sys.stdout.isatty()
    term_smart = os.environ.get("TERM", "") not in {"", "dumb", "unknown"}
    return term_isatty and term_smart
```

Korijn:
> Uh, don't know why CI failed.

Looks like a github outage. I've restarted the failed jobs.

tests/test_rs_basics.py
panxinmiao: Learned something new. :)
I will update it after a little while.

Korijn (approved):
Looks great!

PR #281, opened 2022-06-11
wgsl error tips supports validation error

panxinmiao:
Related #278 #279 

Currently, wgsl error tips only support 'Parsing error'.

This PR adds support for "Validation error".

Like the following:

image


almarklein (approved):
LGTM!

PR #288, opened 2022-10-06
Use srgb for offscreen canvases

almarklein:
See e.g. https://almarklein.org/gamma.html and https://github.com/pygfx/pygfx/issues/329

### DIFF
--- a/wgpu/gui/_offscreen.py
+++ b/wgpu/gui/_offscreen.py
@@ -36,12 +36,12 @@ class WgpuOffscreenCanvas(WgpuCanvasBase):
     def get_preferred_format(self):
         """Get the preferred format for this canvas. This method can
         be overloaded to control the used texture format. The default
-        is "rgba8unorm" (not including srgb colormapping).
+        is "rgba8unorm-srgb".
         """
         # Use rgba because that order is more common for processing and storage.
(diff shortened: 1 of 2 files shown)


almarklein:
This does make the assumption that the shader outputs colors in physical (linear-srgb) space. But this is basically also the assumption for normal canvases.

almarklein:
> The new PNGs look washed out/wrong :/

The triangle is actually better. You can see how the gradient was not smooth, but it is now.

As for the cube ... nothing is technically correct/incorrect because there is no reference 🤷   But I agree that the example would be better if its input were considered srgb. That sets a better example. So I changed that. I also changed the background to white, because the clear color is always physical, which might be confusing.

almarklein:
Better?

almarklein:
Wait, maybe I'll tweak that triangle example some more.

almarklein:
Some more context. Before this PR, when rendering to a canvas, an srgb texture was used. But when rendering to an offscreen buffer, e.g. to make the screenshots, it was a non-srgb texture.

Now it's always an srgb texture (by default), meaning that the colors that the shaders produce should be considered physical colors. The shaders are also adjusted to do a gamma correction, so that we can consider input colors to be the usual srgb.

Korijn (approved):

Issue #291, opened 2022-11-02
Remove debug builds from pypi


FirefoxMetzger:
This is a suggestion rather than an issue. 

I'm currently setting up wgpu-py and pygfx on my Windows workstation and I've noticed that wgpu-py ships with both the debug and the release library (.dll) of libwgpu when installed from pypi. I wonder if it would make sense to only include the release build on pypi, since most users that `pip install wgpu` would expect to get the release version of wgpu. This would bring down the package size in pypi and might have positive performance implications, too. It would also ease CI, as we would only compile the release build instead of compiling both.

Korijn:
I seem to recall that it's helpful in development and debugging; you won't get any stacktraces from wgpu-native with the release build for example. I'm sensitive to the argument of bringing down package size.

> I wonder if it would make sense to only include the release build on pypi, since most users that pip install wgpu would expect to get the release version of wgpu.

Technically, you get both. The release build is used by default. The debug build is only used if you set the environment variable `WGPU_DEBUG=1` when importing wgpu.

> This would bring down the package size in pypi

Yes! Unzipped I think it accounts for about a little over 50% of filesize, but zipped the ratio is less significant. I would assume the two binaries have a lot of commonalities so they compress well. Anyway; less data transferred, means less impact on environment, so I'm sensitive to this one!

> and might have positive performance implications, too

Not at runtime, but it would benefit download and install time for users, and build time for us.

> It would also ease CI, as we would only compile the release build instead of compiling both.

Well, we actually don't compile anything. We maintain the build process on the upstream [wgpu-native](https://github.com/gfx-rs/wgpu-native) repo, and download binaries from their github releases in our workflows. See `download-wgpu-native.py`.

So our CI would not be affected if we drop the debug builds.

almarklein:
Like a year or two ago, the debug builds were useful for when wgpu-native was used in the wrong way to get some more info on what went wrong. Over time, the error reporting of wgpu-native has improved a lot, via API mechanisms. So right now I'd say the debug builds are for devs that want to debug wgpu-py/wgpu-native particularly.  Actually, I think I have not used the debug build for over a year :)

So I agree with @Korijn's proposal - let's remove them from the wheels.

FirefoxMetzger:
I see. If we are using the release build by default then there is indeed only package size.

[...]

PR #299, opened 2022-11-02
package only the release binary

FirefoxMetzger:
Closes: https://github.com/pygfx/wgpu-py/issues/291

This PR adds a small improvement to `setup.py` which makes it only include the release binary instead of including both the debug and the release version. The main benefit of doing so is to reducing package size.

I've also changed the glob (e.g., `*.dll`) into an explicit filename. My angle here is security. We know what the filename will be in advance and this way we can avoid collecting other binaries that we didn't expect to find. If you prefer the glob, I'm happy to revert :)

@Korijn I've also changed the `sys.platform` call to `platform.system`. `platform` is the currently recommended way to deal with platform-specific logic in python so I figured I would use the chance to update. That said, I don't feel very strongly about this and if you'd rather have the old style I'm happy to revert.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -1,5 +1,5 @@
 import re
-import sys
+import platform
 
 from setuptools import find_packages, setup
 from wheel.bdist_wheel import get_platform, bdist_wheel as _bdist_wheel
@@ -19,12 +19,12 @@ class bdist_wheel(_bdist_wheel):  # noqa: N801
 
 
 resources_globs = ["*.h", "*.idl"]
-if sys.platform.startswith("win"):
-    resources_globs.append("*.dll")
-elif sys.platform.startswith("linux"):
-    resources_globs.append("*.so")
-elif sys.platform.startswith("darwin"):
-    resources_globs.append("*.dylib")
+if platform.system() == "Windows":
+    resources_globs.append("libwgpu-release.dll")
+elif platform.system() == "Linux":
+    resources_globs.append("libwgpu-release.so")
+elif platform.system() == "Darwin":
+    resources_globs.append("libwgpu-release.dylib")
 else:
     pass  # don't include binaries; user will have to arrange for the lib
 


almarklein (approved):
The platform triage indeed looks better this way. Explicit filenames sound like a good idea!

FirefoxMetzger:
RTR :) 

I checked the CI artifact and it only includes the release version. For a dev install we would download the binaries independently so this should not be affected by changes in `setup.py`. (though I didn't confirm this explicitly)

Korijn (approved):

Korijn:
> This PR adds a small improvement to `setup.py` which makes it only include the release binary instead of including both the debug and the release version. The main benefit of doing so is to reducing package size.

Great!

> I've also changed the glob (e.g., `*.dll`) into an explicit filename. My angle here is security. We know what the filename will be in advance and this way we can avoid collecting other binaries that we didn't expect to find. If you prefer the glob, I'm happy to revert :)

Nice improvement.

> @Korijn I've also changed the `sys.platform` call to `platform.system`. `platform` is the currently recommended way to deal with platform-specific logic in python so I figured I would use the chance to update. That said, I don't feel very strongly about this and if you'd rather have the old style I'm happy to revert.

No problem, can you refer me to the docs that make this recommendation? I'd like to just read up on that. Thanks!

FirefoxMetzger:
> No problem, can you refer me to the docs that make this recommendation? I'd like to just read up on that. Thanks!

I don't think there is an explicit recommendation in the python docs. However,

- [`sys.platform`](https://docs.python.org/3/library/sys.html#sys.platform) suggests to check out the `platform` module for "detailed checks for the system’s identity".
- `platform.system()` has a more consistent output (Windows, Linux, Darwin, Java, None) than `sys.platform` (`"win32"`, `"linux2"`, `"linux3"`, `"freebsd"`...). As a result you end up with `sys.platform().startswith(...)` rather than `platform.system() == "..."`
- The value of `sys.platform` is set at compile time (see the docs linked above) whereas `platform.system()` is determined at runtime ([it calls uname](https://github.com/python/cpython/blob/46ccb35a400b1abe8c29ad2d7f17295610d768c7/Lib/platform.py#L910-L917)).

I think this is a nice summary: https://stackoverflow.com/a/11674977/6753182 and so is https://stackoverflow.com/a/58071295/6753182 . Both are SO answers though, so also not authoritative.

PR #310, opened 2022-12-08
Execute examples in the test process

Korijn:
This avoids many types of overhead, like process creation, python interpreter startup, module importing, and wgpu device acquisition.

We use the runpy module here (and not exec) since it facilitates the use case quite well but also because it executes the code in a temporary module namespace.

With this change the full test suite runtime is improved as follows:

* Locally: 10.39s to 3.79s
* CI: 10.85s to 4.24s (but likely just a lucky shot, since software rendering is the bottleneck there, rather than initialization)

(Note: exactly the same change as https://github.com/pygfx/pygfx/pull/362)

### DIFF
--- a/examples/tests/test_examples.py
+++ b/examples/tests/test_examples.py
@@ -2,7 +2,6 @@
 Test that the examples run without error.
 """
 
-import asyncio
 import os
 import importlib
 import runpy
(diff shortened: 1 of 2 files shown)


almarklein:
I was wondering whether perhaps this is a problem that should be solved at the source:

https://github.com/pygfx/wgpu-py/blob/ebbc6fc3bfa4cc753bc471e530cd73e038fc57a8/wgpu/gui/offscreen.py#L80-L87

What if we'd just clear all pending tasks after that last line?

Korijn:
Seems to work!

almarklein (approved):
Nice!

PR #311, opened 2022-12-19
Improve the error prompt when wgsl code is multi line error

panxinmiao:
The previous error prompt did not consider the situation when the error code is multiple lines.
For example:

image

It indicates that the error is at line 45, position -390, It is still difficult to quickly locate the wrong position.

This PR improves this, When the error is multiple lines, the code of error part and reason will be output directly, help locate quickly.

image

### DIFF
--- a/tests/test_rs_basics.py
+++ b/tests/test_rs_basics.py
@@ -751,7 +751,12 @@ Parsing error: unknown scalar type: 'f3'
 """.strip()  # noqa
     )
 
-    # test4: Validation error, mat4x4 * vec3
+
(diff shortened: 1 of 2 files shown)


almarklein (changes requested):
Awesome! Could you please add a unit test for this case to avoid regressions?

panxinmiao:
> Awesome! Could you please add a unit test for this case to avoid regressions?

OK, unit test has been added.

Korijn (approved):

almarklein (approved):
Thanks!

PR #312, opened 2022-12-19
Shadertoy and some examples

panxinmiao:
This is the `Shadertoy` implementation of the `wgpu-py` version, It provides a screen pixel shader programming interface similar to "[shadertoy](https://www.shadertoy.com/)".

This PR also provides several examples (most migrated or adapted from shadertoy).

See:  https://github.com/pygfx/pygfx/pull/410


wgpu/utils/__init__.py
```diff
@@ -15,3 +15,4 @@
 
 from ._device import get_default_device  # noqa: F401
 from ._compute import compute_with_buffers  # noqa: F401
+from ._shadertoy import Shadertoy  # noqa: F401
```
Korijn, line 18: `_shadertoy.py` imports `wgpu.backends.rs`, which is something the user should do (see the "Usage" section of the README).

I would recommend to just remove this import here in to avoid the problem :) it's an optional util, people can import the ShaderToy class if they want to use it

panxinmiao: Has been modified. 😄

Korijn:
Finally, I think you should document this class in `docs/reference_utils.rst`!

wgpu/utils/shadertoy.py
```diff
@@ -0,0 +1,311 @@
...
+        self._uniform_data = np.zeros((), dtype=uniform_dtype)
+
+        self.shader_code = shader_code
+        self.resolution = resolution
+
+        self._prepare_render()
+        self._bind_events()
+
+    @property
+    def resolution(self):
```
almarklein, line 141: This public prop needs a docstring :)

wgpu/utils/shadertoy.py
```diff
@@ -0,0 +1,311 @@
...
+        return tuple(self._uniform_data["resolution"])
+
+    @resolution.setter
+    def resolution(self, value):
+        self._uniform_data["resolution"] = value
```
almarklein, line 146: From what I understand, this prop should not have a public setter.

wgpu/utils/shadertoy.py
```diff
@@ -0,0 +1,311 @@
...
+        def on_mouse_move(event):
+            xy = event["x"], event["y"]
+            if event["button"] == 1 or 1 in event["buttons"]:
+                self._uniform_data["mouse"] = xy
+
+        self._canvas.add_event_handler(on_resize, "resize")
+        self._canvas.add_event_handler(on_mouse_move, "pointer_move", "pointer_down")
+
+    def _update(self):
+        now = time.time()
```
almarklein, line 242: Only relative time matters, so let's use a more precise variant:
```suggestion
        now = time.perf_counter()
```

panxinmiao:
> Finally, I think you should document this class in `docs/reference_utils.rst`! Afterwards I'm OK with merging this :)

It seems that CI will imports `shadertoy` module when building document.
Therefore, I put the import code into `try...exceprt...`

I'm not sure if this is in line with the specification. 😅

wgpu/utils/shadertoy.py
```diff
@@ -0,0 +1,311 @@
...
+
+    # todo: add more built-in variables
+    # todo: support input textures
+    # todo: support multiple render passes (`i_channel0`, `i_channel1`, etc.)
+
+    def __init__(self, shader_code, resolution=(800, 450)) -> None:
+
+        self._uniform_data = np.zeros((), dtype=uniform_dtype)
+
+        self.shader_code = shader_code
```
almarklein, line 134: Let's store as `self._shader_code` and add a public `@property` getter. That way it's also listed in the documentation.

docs/reference_utils.rst
```diff
@@ -7,3 +7,14 @@ The wgpu library provides a few utilities. Note that the functions below need to
 .. autofunction:: wgpu.utils.get_default_device
 
 .. autofunction:: wgpu.utils.compute_with_buffers
+
+
+Shadertoy
+--------------------
+
+The Shadertoy class provides a "screen pixel shader programming interface" similar to `shadertoy <https://www.shadertoy.com/>`_,
+Helps you research and quickly build or test shaders using WGSL via WGPU.
```
almarklein, line 16: Bit of a nitpick, but could you put this explanation in the class docstring instead?

wgpu/utils/shadertoy.py
```diff
@@ -0,0 +1,311 @@
+import wgpu
+import time
+
+try:
+    import numpy as np
+    import wgpu.backends.rs
+    from wgpu.gui.auto import WgpuCanvas, run
+except ImportError:
+    pass
```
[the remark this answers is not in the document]
almarklein: Ah, saw the discussion just now :)

panxinmiao:
> 
> Ah, I see. Instead of the try/except, please add the following to `docs/conf.py` on line 18 (prior to any `wgpu` imports):
> 
> ```python
> os.environ["WGPU_FORCE_OFFSCREEN"] = "true"
> ```

`wgpu. backends. rs` still cannot be imported in CI (the `libwgpu.so` library is required). 

I have moved this import code to `_prepare_render()` method, It works. 
Is it OK to modify it like this?

panxinmiao:
> This is great! A few small comments.

Have modified as suggested. 😅

Korijn (approved):

Korijn:
Nice work!

PR #316, opened 2022-12-22
Improve wgsl error reporting

almarklein:
I ran into a case where the line info could not be found, and it printed the whole source. I fixed that, and refactored the code a bit along the way. I put the logic for parsing shader errorrs in a class to hold the different parts together. I also refactored the tests a bit, because the dedented code blocks made the tests very hard to read.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -19,7 +19,8 @@ __pycache__/
 *.so
 
 # Distribution / packaging
-.Python
+.DS_Store
+Python
 build/
 develop-eggs/
 dist/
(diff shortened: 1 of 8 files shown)


Korijn:
> I put the logic for parsing shader errorrs in a class to hold the different parts together.

I have to say I'm not a fan of this practice: classes signal there is some statefulness. When there isn't any, a module should be sufficient to hold the pieces together, and it doesn't give off the confusing impression that there is statefulness.

almarklein:
I chose this approach because this module is a large collection of stuff already. But I agree about it giving the false impression of statefulness. I'll refactor and just use comments and prefixes to keep things grouped.

Korijn (approved):

PR #317, opened 2022-12-23
Fix offscreen size arg

almarklein:
We use `size` in all the GUI backends, but for some reason we used `width` and `height`  for the offscreen canvas 🤷 

This means that the validation examples in PyGfx have the default size instead of the custom one (if given). In particular the new validation example in https://github.com/pygfx/pygfx/pull/404 does not fit on the image :)


Korijn:
> In particular the new validation example in [pygfx/pygfx#404](https://github.com/pygfx/pygfx/pull/404) does not fit on the image :)

I've been wondering about that for a while now! 😂

Korijn (approved):

PR #319, opened 2022-12-23
Make canvases close when they get deleted

### LINKED ISSUE #305: Ensure canvas cleanup
Add a `__del__` method to automatically call `close()` on Canvas base class

almarklein:
Closes #305.

* [x] Implement `WgpuCanvasBase.__del__` to call `self.close()`.
* [x] Avoid refs holding onto the canvas in Qt.
* [x] Avoid refs holding onto the canvas in glfw. Plus tests.
* [x] Avoid refs holding onto the canvas in wx.
* [x] Avoid refs holding onto the canvas in offscreen.
* [x] Avoid refs holding onto the canvas in Jupyter.
* [x] Improved support for interactive sessions with glfw (when asyncio is already running).

### DIFF
--- a/wgpu/gui/base.py
+++ b/wgpu/gui/base.py
@@ -1,6 +1,7 @@
 import os
 import sys
 import time
+import weakref
 import logging
 from contextlib import contextmanager
 import ctypes.util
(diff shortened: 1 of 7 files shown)


Korijn:
Can you describe the scenario where this weakbind construction is important? It seems really complicated...

almarklein:
Registered callbacks hold a reference to the canvas instance, preventing it from being cleaned up. So when a user `del`'s a canvas, it sticks around.

A bound method is essentially a wrapper for a function (from the class) and an instance. The `weakbind` takes these apart and stores the instance in a weakref. So the `proxy` function that it returns has the same capability as the bound method, except not holding onto the instance.

Korijn (dismissed):

wgpu/gui/base.py
```diff
@@ -41,6 +42,21 @@ def log_exception(kind):
             logger.error(msg + f" ({count})")
 
 
+def weakbind(method):
+    """Wrap a method so that the instance is held in a weakref."""
```
almarklein, line 46: More precise docstring.
```suggestion
    """Replace a bound method with a callable object that stores the `self` using a weakref."""
```

Korijn (approved):

PR #327, opened 2023-01-12
Update IDL

almarklein:
Update to the latest webgpu.idl, which may affect the API.

The eventual API changes are minimal this time.
- Texture has new props weight, height, depth_or_array_layers  
- Buffer has new prop map_state  

The IDL does move `get_preferred_format` to the global GPU object, but we need to have it on the canvas context, because we have different types of canvases. So in the end this did not result in a change 🤷


Korijn (approved):
LGTM!

almarklein:
I'll merge this once I've made sure that PyGfx works with the update too.

PR #343, opened 2023-02-17
Support GLSL in shadertoy

### LINKED ISSUE #339: Support GLSL in shadertoy
wgpu now supports GLSL. Making the shadertoy support it too, would mean users could almost copy paste the shaders from the internet.

I imagine @panxinmiao might want to have a look at this? No rush though!

panxinmiao:
close #339 

The shadertoy tool now supports GLSL code.

When using GLSL, the built-in variables and entry functions have an additional alias (camel-case), which is consistent with the shadertoy website. In this way, we can use the code directly copied from the website almost without any modification.

I also added several examples of using GLSL.

### DIFF
--- a/tests/test_util_shadertoy.py
+++ b/tests/test_util_shadertoy.py
@@ -19,7 +19,7 @@ def force_offscreen():
 
 
 def test_shadertoy_wgsl():
-    # Import here, because it imports the wgou.gui.auto
+    # Import here, because it imports the wgpu.gui.auto
     from wgpu.utils.shadertoy import Shadertoy  # noqa
 
     shader_code = """
@@ -44,7 +44,7 @@ def test_shadertoy_wgsl():
 
 
 def test_shadertoy_glsl():
-    # Import here, because it imports the wgou.gui.auto
+    # Import here, because it imports the wgpu.gui.auto
     from wgpu.utils.shadertoy import Shadertoy  # noqa
 
     shader_code = """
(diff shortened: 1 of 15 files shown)


tests/test_util_shadertoy.py
```diff
@@ -38,5 +38,31 @@ def test_shadertoy():
     shader = Shadertoy(shader_code, resolution=(800, 450))
     assert shader.resolution == (800, 450)
     assert shader.shader_code == shader_code
+    assert shader.shader_type == "wgsl"
+
+    shader._draw_frame()
+
+
+def test_shadertoy_glsl():
+    # Import here, because it imports the wgou.gui.auto
```
almarklein, line 47: typo
```suggestion
    # Import here, because it imports the wgpu.gui.auto
```

tests/test_util_shadertoy.py
```diff
@@ -18,15 +18,15 @@ def force_offscreen():
         del os.environ["WGPU_FORCE_OFFSCREEN"]
 
 
-def test_shadertoy():
+def test_shadertoy_wgsl():
     # Import here, because it imports the wgou.gui.auto
```
almarklein, line 22: typo (was there earlier)
```suggestion
    # Import here, because it imports the wgpu.gui.auto
```

almarklein (changes requested):
Nice! Just two typo's.

almarklein (approved):

Korijn (approved):

PR #346, opened 2023-02-17
Avoid unexpected numpy imports

almarklein:
In #345 I ran into the doc build env not having installed numpy, and therefore failing the build. Numpy is not a dependency. So let's fix that.

* [x] Remove numpy usage from `gui/offscreen.py`. Easy, because a memory object is easy enough to turn into a numpy array by the caller. 
* [x] Remove numpy usage from `utils/shadertoy.py`. Had to refactor the uniform array stuff a bit.
* [x] Added a CI testbuild to avoid regressions. It simply imports wgpu and some of its subpackages in a minimal env.
* [x] Refactoring the CI code a bit, moving the codegen and pyinstaller tests out of the matrix-test-build. This makes the yaml easier to read and the tests run faster.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -13,14 +13,14 @@ on:
 jobs:
 
   lint-build:
-    name: Linting
+    name: Test Linting
     runs-on: ubuntu-latest
     strategy:
       fail-fast: false
     steps:
-    - uses: actions/checkout@v2
+    - uses: actions/checkout@v3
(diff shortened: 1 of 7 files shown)


Korijn:
Because you renamed some jobs, you have to go into the branch protection settings and changed the required jobs.

almarklein:
Ah, I though GH was having caching issues. Fixed!

Korijn (approved):

PR #365, opened 2023-07-24
Don't enable features by default 

almarklein:
(should do that in pygfx)

Let's wait with merging until we've updated pygfx accordingly.


Korijn (approved):

almarklein:
The tests fail because we turned off the feature 😆

almarklein:
Fixed!

Korijn (approved):

almarklein:
In theory, someone using a bleeding edge version of wgpu-py and a release-version of pygfx may get into trouble if we merge this. Not a very likely situation but I suppose there is no harm in keeping this open for a bit.

PR #369, opened 2023-10-03
Update API to latest IDL

almarklein:
This updates wgpu-py to the latest WebGPU spec (represented in IDL). There are just a few minor changes and these mostly affect API that's not really "active" yet.

### DIFF
--- a/wgpu/backends/rs.py
+++ b/wgpu/backends/rs.py
@@ -287,7 +287,8 @@ class GPU(base.GPU):
 
         # For now, Rust will call the callback immediately
         # todo: when wgpu gets an event loop -> while run wgpu event loop or something
-        assert adapter_id is not None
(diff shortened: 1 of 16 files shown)


wgpu/backends/rs.py
```diff
@@ -339,7 +347,11 @@ def to_py_str(key):
         # WebGPU features
         features = set()
         for f in sorted(enums.FeatureName):
-            i = enummap[f"FeatureName.{f}"]
+            key = f"FeatureName.{f}"
+            if key not in enummap:
+                assert f in KNOWN_MISSING_FEATURES
```
Korijn, line 352: I strongly advise against using `assert` (especially without a message) in user code (anything outside a `tests` folder)

almarklein: Fixed!

Korijn (approved):

PR #370, opened 2023-10-03
Update to wgpu-native 0.17.0.2

### LINKED ISSUE #371: Error handling
### In short
In the update of #370, wgpu-native has become more permissive of errors. This means that with the new code, code/examples that have an error, will generate error messages, but they do not fail.

### Details
The wgpu-native lib notifies us of errors using a callback, i.e. errors are logged asynchronously. In the callback (`_on_error()`) we `logger.error()` the message, which displays it to the terminal. A second mechanism by which errors were communicated is that ids for new objects failed to set. We raised an error when this happened.

So previously, if the wgsl failed to compile, you'd first get an error being logged, and then an exception for the shader module not being created.

<details>
<summary>example shader error</summary>
<samp>
Running script: "/Users/almar/dev/py/wgpu-py/examples/triangle.py"

Shader error: label:  Some("")
Parsing error: unknown attribute: 'vertexx'

   ┌─ wgsl:10:1
   │
10 │ @vertexx
   │  ^^^^^^^ unknown attribute
   │
   = note:

  File "/Users/almar/dev/py/wgpu-py/examples/triangle.py", line 159, in <module>
    main(canvas)
  File "/Users/almar/dev/py/wgpu-py/examples/triangle.py", line 69, in main
    return _main(canvas, device)
  File "/Users/almar/dev/py/wgpu-py/examples/triangle.py", line 82, in _main
    shader = device.create_shader_module(code=shader_source)
  File "/Users/almar/dev/py/wgpu-py/wgpu/backends/rs.py", line 1287, in create_shader_module
    raise RuntimeError("Shader module creation failed")
RuntimeError: Shader module creation failed
</samp>
</details>

With the new version there is the logged error, but the module is still created, so later more errors get logged as the code tries to use an broken shader module. In effect, our examples-test considers the example to be ok.

### Possible solutions

1. In the example tests, detect whether errors have been logged. 
2. Detect whether errors have been logged in a few tactical places, e.g. when a command encoder is created.
3. Check for errors on every API call that makes calls into the lib, and raise them as Python exceptions.
4. Check for errors on every lib call.

I feel quite strongly that one of the last two options is the best way to go. If you try to create a shader-module with an incorrect shader, you simply want an exception asap. This means, however, that we'd need to include error-handling in nearly every method in `rs.py`. Maybe we can use a decorator or meta class to do that efficiently. 

@Korijn thoughts?

almarklein:
Closes #371 

* [x] Update the header files for wgpu-natove 0.17.0.2
* [x] Run codegen, fix parsing, and handle the fixme's.
* [x] The standard error messages are improved, we can ditch our smart shader error handling! 
* [x] The `xxxDrop` methods are now `xxRelease`.
* [x] ~Re-activate two release methods that were previously not used because they segfaulted.~  nope, still panics.
* [x] ~Maybe we can call all release methods directly rather than via a delay?~ nope, still hangs.
* [x] Update wgsl in examples.
* [x] ~Fix that the glsl shadertoy examples hang. -> looks like a race condition related to asyncio.~ No longer broken for some reason.
* [x] Update manylinux target.
* [x] Test pygfx on it.
* [x] Update release notes.


wgpu/gui/offscreen.py
```diff
@@ -76,11 +76,6 @@ def draw(self):
 
 def call_later(delay, callback, *args):
     loop = asyncio.get_event_loop_policy().get_event_loop()
-    # for the offscreen canvas, we prevent new frames and callbacks
-    # from being queued while the loop is running. this avoids
-    # callbacks from one visualization leaking into the next.
-    if loop.is_running():
-        return
     loop.call_later(delay, callback, *args)
```
almarklein, line 79: Oh, this un-does #322, so this is probably not a good idea ...  

I was under the impression that this was a leftover and technically not needed because we cancel all tasks below. But these are times, not tasks.

CHANGELOG.md
```diff
@@ -25,6 +25,15 @@ Added:
 
 * New `wgpu.wgsl_language_features` property, which for now always returns an empty set.
 * The `GPUShaderModule.compilation_info` property (and its async version) are replaced with a `get_compilation_info()` method.
+* The WebGPU features "bgra8unorm-storage" and "float32-filterable" are now available.
+
+Changed:
+
+* The binary wheels are now based on manylinux 2.29, and the 32bit Linux wheels are no longer build.
```
Korijn, line 32: in ci.yml you're using 2.28

CHANGELOG.md
```diff
@@ -25,6 +25,15 @@ Added:
...
+* In WGSL: toplevel constantas must be defined using `const`, using `let` will now fail.
+* In WGSL: it is no longer possible to re-declrare an existing variable name.
+* Error messages may look a bit different, since wgpu-native now produces nice messages replacing our custom ones.
+* Errors produced by a call into a wgpu-native function now produce a Python exception (no more async logging of errors).
```
Korijn, line 36:
```suggestion
* The binary wheels are now based on manylinux 2.28, and the 32bit Linux wheels are no longer built.
* In WGSL: toplevel constants must be defined using `const`, using `let` will now fail.
* In WGSL: it is no longer possible to re-declare an existing variable name.
* Error messages may look a bit different, since wgpu-native now produces nice messages replacing our custom ones.
* Errors produced by a call into a wgpu-native function now produce a Python exception (no more async logging of errors).
```

Korijn (approved):

CHANGELOG.md
almarklein: nice catch - this was indeed a typo

Korijn (approved):

PR #377, opened 2023-10-15
Buffer mapping API

### LINKED ISSUE #114: Making buffer mapping part of the public API?
Also see:
* https://github.com/gpuweb/gpuweb/issues/605
* https://github.com/gpuweb/gpuweb/issues/897
* https://github.com/python/cpython/pull/105290
* https://gist.github.com/almarklein/a7b59d5cf18ae979f79602115807b0b2  - ways in which invalidating views does not work
* 
## Intro

We're deviating from the WebGPU API with respect to how to read/write buffer data, because its hard to reproduce the API in Python in a way that does not make it very easy for the user to access released memory and thus cause a segfault.

## The WebGPU API
 
The WebGPU API for synchronizing data between a GPUBuffer and the CPU makes use of "mapping". The API works as follows:
* You request the buffer to map its data (`mapAsync()`), using a specific range.
* Then you obtain an `ArrayBuffer` using `getMappedRange()`, again using a subrange (within the total subrange).
* You copy to/from that array-buffer (via a typed array view).
* You `unmap` the buffer.

This API offers appealing advantages:

* Allows reading and writing data in a buffer.
* Allows doing that with subranges.
* Allows doing that with multiple subranges in one go, i.e. without mapping/unmapping multiple times.
* Without unnecessary data copies.

## The problem

It is challenging to find a Pythonic API to replicate this behavior. I think that any solution that we implement should make it *impossible* for the user to access the memory after it is unmapped. (I mean impossible, unless the users is deliberately using ffi or something to do so.)

However, this appears to be oddly hard to do with the way buffers, arrays and views interact in Python.

As an example. one can invalidate a `memorview` object by calling its `release()` method. But if another memoryview or numpy array has been mapped to the same memory, these continue to work.

## What we have now

The solution so far has been to implement a much simpler API using `map_read()` and `map_write(data)`, without the option to read/write a subrange of the buffer. So basically only the first bulletpoint.

## What we need

At the very least the first two bullet points. I very much hope to also include the third bullet-point. If we can't avoid data-copies (the fourth bullet point), that's unfortunate, but not problematic as Python is not super-duper-fast anyway.

### Stick to read_data and write_data

What we have now, but can include args to specify a range.

### Chunked writing / reading

I think this would cover most use-cases, without the need to expose the mapping stuff:

```py
def write_chunks(sequence, offset=0, size=0):
    # With sequence an iterable (or even a generator) providing tuples (offset, data)
    ...
```

### Mapping, but via a custom class so that we can restrict access

```py
class BufferMapping:

    def __init__(self, mem):
        self._never_touch_this_mem = mem  # a memoryview
        self._ismapped = True  # The buffer will set this to False when it's unmapped

    def cast(self, format, shape=None):
        if not self._ismapped:
            raise RuntimeError("Cannot use a buffer mapping after it's unmapped.")
        self._never_touch_this_mem = self._never_touch_this_mem.cast(format, shape)
        return self

    def __getitem__(self, index):
        if not self._ismapped:
            raise RuntimeError("Cannot use a buffer mapping after it's unmapped.")
        res = self._never_touch_this_mem.__getitem__(index)
        if isinstance(res, memoryview):
            raise IndexError("Cannot get a sub-view")  # or also wrap in a BufferMapping?
        return res

    def __setitem__(self, index, value):
        if not self._ismapped:
            raise RuntimeError("Cannot use a buffer mapping after it's unmapped.")
        self._never_touch_this_mem.__setitem__(index, value)

    def to_memoryview(self):
       # Make a copy
        new_obj = (ctypes.c_uint8 * self._never_touch_this_mem.nbytes)()
        new_mem = memoryview(new_obj)
        new_mem[:] = self._never_touch_this_mem
        return new_mem
```

The thing is ... when would you use this? To map the data and then setting data elements one by one? That would be slow because of the overhead that we introduce. In batches then? Well, in that case you could call `write_data(subdata, offset)` a few times ...

The use-cases where a mapped API has an advantage (in Python) seem flaky, and the API is much more complex. Therefore we don't currently expose this API in wgpu-py.

However ... I could miss a use-case. And I could miss a possibly elegant solution.

almarklein:
This closes #114, also updated that issue's top comment to explain the issue better.

Repeating the advantages of the WebGPU mapping API for clarity:
* Allows reading and writing data in a buffer.
* Allows doing that with subranges.
* Allows doing that with multiple subranges in one go, i.e. without mapping/unmapping multiple times.
* Without unnecessary data copies.

I've had another look at disabling the base array so that the views don't work anymore, but without success.

Therefore this proposal does not expose a mapped array to read from / write to. Instead it has two methods for this: `read_mapped()` and `write_mapped()`. The former has a kwarg `copy` that can be set to False to get the actually mapped data. Users who know what they're doing can then benefit from more performance.

With this new API we support all the four points above. It's also a lot closer to the WebGPU spec: we have `map()` and `unmap()` in between which the data can be read or written multiple times (fast!). But instead of obtaining a piece of the mapped memory (`getMappedRange()`), we have `read_mapped()` and `write_mapped()`, so we don't have to expose the mapped memory (except whith  `read_mapped(..., copy=False)`.


almarklein:
I moved the test file with cases that *do not* work to a Gist: https://gist.github.com/almarklein/a7b59d5cf18ae979f79602115807b0b2

Korijn:
This may be ignorant but how bad is it really that you could have a view on released memory? It could just be a documented risk that users of wgpu-py need to be aware of.

almarklein:
Well we could document it carefully, but it's so easy to forget, also because this is not something people are used to having to care about in Python. What makes it worse is that these views can easily end up in a different place unrelated to wgpu. That, plus the fact that segfaults are sometimes not reproducible (sometimes it will crash, sometimes you get bogus data), can result in some nasty bug tracking ...

I also realized just now that the data-copy downside of only an issue when reading from the buffer. When writing it's not because Python/numpy makes use of views so much.

almarklein:
Just pushed a proposal for an API. The buffer can be mapped and unmapped like the WebGPU spec, but instead of requesting an array-like object (that represents the mapped memory) to read from or write to, the buffer has methods `read_mapped()` and `write_mapped()`.

Still have to adjust examples and tests.

wgpu/backends/rs.py
```diff
@@ -1535,76 +1533,87 @@ def callback(status_, user_data_p):
...
+
+        # Get mapped memoryview.
         # H: void * f(WGPUBuffer buffer, size_t offset, size_t size)
         src_ptr = libf.wgpuBufferGetMappedRange(self._internal, 0, size)
         src_address = int(ffi.cast("intptr_t", src_ptr))
         src_m = get_memoryview_from_address(src_address, size)
+
+        # Copy the data. The memoryview created above becomes invalid when the buffer
+        # is unmapped, so we don't want to pass that memory to the user.
+        data = memoryview((ctypes.c_uint8 * size)()).cast("B")
```
Korijn, line 1580: you could add a kwarg (I'm not sure if that would conflict with the IDL) like `copy=True` to allow users to opt in to the dangerous behaviour?

almarklein: That sounds reasonable! At least in that case its a very deliberate choice.

almarklein: Now we check all 4 of the boxes 🚀

almarklein:
Ran into an interesting edge-case. Filed an issue upstream: https://github.com/gfx-rs/wgpu-native/issues/305  For now I fixed it here.

almarklein:
This change does not actually require changes in PyGfx - it looks like we only used the parts of the API that have not changed.

Korijn (approved):

almarklein:
> This change does not actually require changes in PyGfx - it looks like we only used the parts of the API that have not changed.

Woops, using it in the picking logic. Will prepare a pr for pygfx shortly.

PR #382, opened 2023-10-19
Stop forcing Vulkan on Windows

almarklein:
Until now we forced using the Vulkan adapter backend on Windows, because DX12 was less well support in the early days of wgpu-core.

* [x] Remove code that explicitly selects Vulkan for Windows.
* [x] Avoid raising exceptions in ffi callbacks.
* [x] Add ability to select fallback adapter in `wgpu.request_adapter()`.

## Motivation for selecting Vulkan

* Without drivers installed, Vulkan cannot be used.
* I found that without drivers installed, not forcing the Vukan adapter runs (using D3D12), but not without flaws. At least better than Vulkan.
* At wgpu-core D3D12 is listed as having first class support, but ...
* The D3D12 backend (also with drivers installed) still has flaws, but that's ok, because ...
* I found that *with* drivers installed, auto-selection picks Vulkan.
* Eventually wgpu can also use D3D11 on older Windows versions that don't support Vulkan or Dx12.
* The situation around honoring the Nvidia panel settings are the same for all backends.

In summary, I think its fine to let wgpu select the backend on Windows from hereon.

## Experiments on a clean Windows install

* With a clean Windows install (no Nvidia drivers installed yet):
       * Vulkan (forced) crashes when it tries to display a window.
       * Did not test D3D12 (missed that opportunity)
* With Nvidia drivers uninstalled:
       * Vulkan reports "no available adapter".
       * D3d12 kinda works. It runs some/most examples, but fails the 3D tests in test_rs_compute_tex.py, produces an error for the Naga-generated DX shader code on the volume rendering example, and buffer mapping (e.g. in pygfx picking) seems iffy too.
       * Tried D312 with both Integrated Graphics and GPU, seems the same errors come up, which is in (extra) indication that the problem is in the part where wgpu is converted to D3D12.
* With NVidia drivers:
    * When letting wgpu select a backend, it picks Vulkan.
    * Vulkan work as expected, passing all tests (unit and examples) in wgpu-py and pygfx.
    * Forcing D3D12 creates the same glitches as mentioned above.
    * Forcing D3D11 reports no available adapter.
    * Forcing the fallback adapter (software rendering, with D3D12) is not flawless, but seems to have less issues than D3D12 on hardware.

## Experiments on reacting to Nvidia control panel

* Both with Vulkan and D3D12 (and also with auto-select, i.e. Vulkan), the Nvidia control panel setting is ignored.
* It simply selects the Nvidia GPU or Integrated Graphics only based on the value of `power_reference` in `wgpu.request_adapter()`.

### DIFF
--- a/wgpu/backends/rs.py
+++ b/wgpu/backends/rs.py
@@ -183,7 +183,9 @@ libf = SafeLibCalls(lib, error_handler)
 
 @_register_backend
 class GPU(base.GPU):
-    def request_adapter(self, *, canvas, power_preference=None):
+    def request_adapter(
+        self, *, canvas, power_preference=None, force_fallback_adapter=False
+    ):
         """Create a `GPUAdapter`, the object that represents an abstract wgpu
         implementation, from which one can request a `GPUDevice`.
 
@@ -194,7 +196,9 @@ class GPU(base.GPU):
                 render to (to create a swap chain for, to be precise). Can be None
                 if you're not rendering to screen (or if you're confident that the
                 returned adapter will work just fine).
-            powerPreference(PowerPreference): "high-performance" or "low-power"
+            power_preference(PowerPreference): "high-performance" or "low-power".
+            force_fallback_adapter (bool): whether to use a (probably CPU-based)
+                fallback adapter.
         """
 
         # ----- Surface ID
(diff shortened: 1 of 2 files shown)


Korijn (approved):

Korijn:
How about the graphics preference in Windows Control Panel? As far as I am aware that has superseded the Nvidia settings thing.

almarklein:
> How about the graphics preference in Windows Control Panel? 

This one seems also to be ignored 🤷

Issue #360, opened 2023-05-18
Issues/Improvements to Shadertoy support


Vipitis:
hey,

I have found this project and it looks really promising. I am working on a project that looks at language models generating shadercode.
Following the examples, I got it working for myself. 
However I had several issues.
 - there is `shader.show()` but to stop there is `shader._canvas.close()` (in a Jupyter notebook).
 - to get the actual frames you can call `shader._canvas.get_frame()`
 - Shadertoy always uses the auto canvas, so you can't get it on a fake Canvas for example.
 - I have two instances of the Shadertoy class to then compare ground truth and generation, the 2nd shader is a frame ahead

Now my questions:
- Is there a way to render a sinlge frame (perhaps given a time)?
- How are erros in the shadercode handled? right now I seem to get a Jupyter kernal crash. (might be GPU dependant?)
- can you access the parser/compiler to figure out where in the  source code a specific function is defined?

My suggestion:
 - add methods to the Shadertoy class to access the information inside the `._canvas` directly
 - expose more arguments to the `__init__()` to set canvas, framerate, ... (keep defaults).

Will read through more docs and look at the source, but the low level stuff is too advanced for me, might find some time to contribute myself.

almarklein:
Hi!

General remarks regarding the shadertoy utility: 
* The current shadertoy is designed to be really plug-n-play, and is not (yet) very configurable. 
* That said, I'm sure it could be improved, perhaps by adding optional arguments? 
* The shadertoy was contributed by a user, and since it is not our priority, we are unlikely to improve it ourselves. Though we'd accept contributions to improve it 👍 

> Is there a way to render a sinlge frame (perhaps given a time)

Not at the moment. Perhaps an option could be added e.g. `offscreen` that will then use an offscreen canvas and return the screenshot?

> How are erros in the shadercode handled? right now I seem to get a Jupyter kernal crash. (might be GPU dependant?)

In most cases the errors should be caught when the GPU shader module is created and a clear error should be printed. A kernel crash should be considered a bug (though it might be one that we cannot fix, because it's cause lies upstream in wgpu-native).

> can you access the parser/compiler to figure out where in the source code a specific function is defined?

No, it's opaque. We just call a function that we pass the wgsl, and it produces a GPU module.

> expose more arguments to the __init__() to set canvas, framerate, ... (keep defaults).

This sounds like the most feasible approach to make this utility more flexible 👍

Vipitis:
Thank you got the answers.

[...]

PR #401, opened 2023-10-28
Improve Shadertoy Utility

Vipitis:
picking up from https://github.com/pygfx/wgpu-py/issues/360

Several steps to this PR (will be updated)
- [x] add offscreen support
- [x] add snapshot method
- [x] ~~validate shadercode (with naga?) before creating shader module~~ (addressed by https://github.com/gfx-rs/wgpu-native/pull/313)
- [x] add tests for new features
- [x] fix style
- [ ] rebase

I welcome any feedback to my additions. I don't want to break any existing functionality.


wgpu/utils/shadertoy.py
```diff
@@ -1,9 +1,14 @@
 import time
 import ctypes
+import numpy as np
+from PIL import Image
```
almarklein, line 4: This gets a bit tricky in terms of dependencies. I'd prefer to make `snapshot()` return a memoryview object, that can be easily converted to a numpy array (can mention in the docs that ppl can use `np.asarray()`.

Vipitis: definitely, I haven't even looked up what dependencies are included, and looks like numpy is just in the dev requirements. 
Will shorten the method and the copy over the docstring

Vipitis:
short update, I had encountered more panics and found a ways to catch them by doing more translations.
And now that glsl errors get handled properly, it really shouldn't be here anymore.

Would it be okay to take all of that validation out (I will keep it in own project instead) and just add the offscreen and snapshot features with this PR for now?

wgpu/utils/shadertoy.py
```diff
@@ -463,8 +474,67 @@ def _draw_frame(self):
...
         self._canvas.request_draw(self._draw_frame)
-        run()
+        if self._offscreen:
+            run_offscreen()
+        else:
+            run()
+
+    def snapshot(self, time_float: float = 0.0, mouse_pos: tuple = (0, 0, 0, 0)):
+        """
+        Returns a PIL Image of the specified time. (Only available when ``offscreen=True``)
```
almarklein, line 484:
```suggestion
        Returns an image of the specified time. (Only available when ``offscreen=True``)
```

almarklein:
> Would it be okay to take all of that validation out (I will keep it in own project instead) and just add the offscreen and snapshot features with this PR for now?

Yes, that sounds good! The glsl validation are present in the latest wgpu-native release, so they will be in wgpu-py in the next update.

almarklein:
Thanks!

PR #402, opened 2023-10-30
Fixup compute_noop.py numpy example

claydugo:
I found this example to be very helpful. 

It looks like the shader source didnt get changed to the new variable in some past refactor. 

This got it running for me and printing like the other 2 examples.


Korijn (approved):

almarklein:
Thanks!

Issue #391, opened 2023-10-23
First class async support


almarklein:
### Introduction

WebGPU defines certain functions and props as async, so that in an async setting, there is less wait time. Would make a lot of sense to  also support that in wgpu-py.

Would be great if we can make this work without forcing asyncio, but also supporting e.g. trio.

At the moment the async API of `wgpu-native` is a bit crude (there is a function to poll its loop). Last I heard was that this API will change at some point. So might be worth waiting for that.

In our code, all async methods have an `_async`  suffix, and they all have a sync version of the function (that simply waits). 

### Aync parts in WebGPU spec

*Updated on 25-09-2024*

Methods that are async-only in WebGPU:

* `gpu.request_adapter_async()`
* `gpu.enumerate_adapters_async()` 
* `adapter.request_device_async()` 
* `buffer.map_async()` 
* `device.lost_async`
* `shadermodule.get_compilation_info_async()`
*  `queue.on_submitted_work_done_async()`

Method that have both a sync and async variant:

* `create_compute_pipeline()`
* `create_render_pipeline()`

### Considerations

* Would be nice to be able to use the API synchronously, e.g. make a script that does compute. 
* wgpu (and derived code) should work well in an interactive session.
* Also support e.g. Trio.
* Not all GUI event loops support async/await, e.g. Qt.
* For WebGPU: no way to make the async functions synchronous (I think).
* How much is performance afected by using sync versions of `createRenderPipelineAsync` and `mapAsync`?
* Maybe prefix the sync variants with `_sync` to indicate that an async variant is available?

### Thoughts

I think I'd like to try making wgpu-py async according to the spec, dropping the sync versions. Let's see how far we can take that. In render engines, one could run "draw iteration" on asyncio, if necessary with `loop.run_until_complete()`. Also when inside Qt. 

If we can make that work, we're more compliant, avoid busy waiting in user code, and remove the biggest hurlde to move to the  browser.

### Async in wgpu-native

*01-10-2024*

At the time of writing, this is still somewhat in flux. 

References:
* https://github.com/webgpu-native/webgpu-headers/issues/199
* https://docs.google.com/document/d/1qJRTJRY318ZEqhK6ryw4P-91FumYQfOnNP6LpANYy6A/edit?usp=sharing
* https://webgpu-native.github.io/webgpu-headers/Asynchronous-Operations.html

State:
* In wgpu-native we have `wgpuDevicePoll()` (in `wgpu.h`).
[...]

PR #403, opened 2023-11-01
Add tests for object releasing

### LINKED ISSUE #275: Rust panic after creating hundreds of renderers/canvases
I still need to figure out how to reliably reproduce this, but in a script I was batch running yesterday, my computer would actually lock up* after a rust panic. The problem went away after I modified the script to reuse the same canvas and renderer instance, instead of creating new ones for every image I was drawing.

\* Lock up as in, no response to ctrl+c, windows task manager would not respond, ctrl+alt+del does nothing, shutting down/rebooting gets stuck in a black screen... only way out was to hold the power button for 30 seconds.

### LINKED ISSUE #353: Investigate memoryleaks
In #306 we explored the memoryleak of Lavapipe, and found out that Device objects leak memory on Windows too. To be fair, in practice you'd likely only use one device per process, but there are use cases for using multiple devices. Let's also do some testing on other objects, and report/fix errors upstream.

The purpose of this issue is to track known leaks. We want wgpu-py to be a reliable system without leaks.

### Status

* Adapter
  * [ ] All systems: Memory does grow, albeit very slowly. Could be a small leak?
* Device
  * [ ] LavaPipe with Vulkan: Leaks significantly. 
  * [ ] Windows with Vulkan/D3D12: Leaks significantly. Crashes at the 64th iter! 
  * [ ] MacOS with Metal: Does not seem to leak.
* Buffer
* Texture
* ...

----

Code to track (updated):
```py
import gc
import psutil

import wgpu.backends.rs
import wgpu


p = psutil.Process()


def print_mem_usage(i):
    megs = p.memory_info().rss / 1024**2
    print(f"memory usage (round: {i}): {megs:.3f} MB")


if __name__ == "__main__":
    print_mem_usage(0)
    for i in range(500):
        adapter = wgpu.request_adapter(canvas=None, power_preference="high-performance")
        device = adapter.request_device()
        device.destroy()
        del device
        wgpu.backends.rs.delayed_releaser.release_all_pending()
        gc.collect()
        print_mem_usage(i + 1)

    wgpu.print_report()
```

almarklein:
Closes #353

This adds tests, in a new folder so it won't interfere with the unit tests. The idea is basically, doing for each kind of object:

* Get how many objects we have of each kind.
* Create a bunch of objects of a specific kind.
* Make sure the counts are indeed up.
* Delete the objects. 
* Check that the counts are down.

Todo:

* [x] Generic logic that is the same for each unit test.
* [x] Implement some meta-tests to make sure that the tests indeed fail when they should. 
* [x] Implement a test function for every object.
* [x] Can we *now* remove that `delayed_releaser`? Yes!
* [x] Can we *now* enable some release calls that previously panicked? Yes!
* [x] ~Measure process memory consumption to detect leaks?~ --> #406
* [x] Do we want our `BindGroupLayout` object to cache instances like wgpu-core does?
* [ ] Does this close #275?
* [x] Investigate fails:
    * [ ] Device (simply does not clean up in native).
    * [ ] Queue (related to the above)
    * No luck yet.
    * Following up in #405

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -198,9 +198,12 @@ jobs:
           pip install -U -r dev-requirements.txt
           python download-wgpu-native.py
           pip install -e .
-    - name: Test on repo
+    - name: Unit tests
       run: |
           pytest -v tests
+    - name: Memory tests
+      run: |
+          pytest -v tests_mem
 
   # The release builds are done for the platforms that we want to build wheels for.
   # We build wheels, test them, and then upload the wheel as an artifact.
--- a/tests_mem/test_mem.py
+++ b/tests_mem/test_mem.py
@@ -1,10 +1,10 @@
 import gc
+import asyncio
 
 import wgpu.backends.rs
-import wgpu.gui.offscreen
 
 import pytest
-from testutils import can_use_glfw, can_use_wgpu_lib
+from testutils import can_use_glfw, can_use_wgpu_lib, can_use_pyside6
 
 
 if not can_use_wgpu_lib:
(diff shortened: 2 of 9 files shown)


Korijn:
Can you explain what polling the device does conceptually?

almarklein:
> Can you explain what polling the device does conceptually?

From what I understand, wgpu has an an event-loop of sorts, so it can perform asynchronous tasks. Polling the device flushes the device's queue. We also use poll in the functions that are asyn in nature, like buffer mapping. Related to #391.

almarklein:
I think this is good to go. I'll have a look (probably after the weekend) what the state of #275 is. But in the mean time this is ready for review.

almarklein:
I created 2 new issues for follow-up work.

Korijn (approved):

PR #404, opened 2023-11-03
Remove asyncio from offscreen gui

almarklein:
When working on a unit test for canvases in #403, I realized that the behavior in `gui/offscreen.py` was a bit weird. We use asyncio for the event loop, but we also make `run()` behave as if there is no loop. We've had to change this code one or two times because of lingering refs due to the `call_later`.

This PR removes the event loop entirely, and fakes `call_later` a bit to make examples that use it still behave more or less as can be expected.

### DIFF
--- a/wgpu/gui/offscreen.py
+++ b/wgpu/gui/offscreen.py
@@ -1,4 +1,4 @@
-import asyncio
+import time
 
 from ._offscreen import WgpuOffscreenCanvas
 from .base import WgpuAutoGui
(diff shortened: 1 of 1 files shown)


wgpu/gui/offscreen.py
```diff
@@ -38,10 +38,11 @@ def is_closed(self):
         return self._closed
 
     def _request_draw(self):
-        call_later(0, self.draw)
+        # Deliberately a no-op, because people use .draw() instead.
+        pass
 
     def present(self, texture_view):
-        # This gets called at the end of a draw pass via GPUCanvasContextOffline
+        # This gets called at the end of a draw pass via _offscreen.GPUCanvasContext
```
almarklein, line 45: Oh, this change is actually from #403, but I guess that's fine since its only a comment.

Korijn (approved):
It's not like we are making `async` functions synchronous here, so it's not really changing the API at all. Asyncio is an implementation detail. So I'm okay with this change!

almarklein:
> so it's not really changing the API at all

Correct. Even the behavior should be the same. The advantage is that the code is simpler, and that it also works in an application that already has asyncio running (which before this pr was not working well).

PR #408, opened 2023-11-08
Refactor backends

### LINKED ISSUE #393: Renaming rs backend to native
I called the wgpu-native backend `rs` because it was written in Rust, and perhaps the project was just called "wgpu" at that point?

Anyway, the `rs` is a bit weird now, so I propose to rename it. 

Implication wise, it means a lot of people will have to rewrite their import to `import wgpu.backends.native`, but this is easy to deprecate in a friendly way, and we've backwards incompatible changes that are more work to adjust t

almarklein:
In this PR:
* [x] Introduce an `auto` backend, that we can later use to select the most appropriate backend.
* [x] No more need to explicitly import the backend, because `request_adapter()` loads it automatically. 
* [x] Rename the `rs`  backend  to `wgpu_native` (closes #393). 
* [x] When a backend is loaded, it only replaces the `wgpu.gpu` object. 
* [x] The `request_adapter()` function is now a method of `wgpu.gpu`. 
* [x] Backend-specific logic is now only available from the backend's subpackage, as a functional API (`enumerate_adapters()`, `request_device_tracing()`). 

Todo:
* [x] Fix up codegen.
* [x] Check download script.
* [x] Move `request_device_tracing()`
* [x] Update docs on needing to import the backend.
* [x] Same for readme, examples, docstrings maybe.
* [x] Merge #412 in here.
* [x] Make `canvas` arg in `request_adapter()` optional? -> No, this arg has been removed already in next webgpu spec.
* [x] Check / review the pyinstaller hook (also wrt glfw).
* [x] Deprecate `wgpu.request_adapter()` in favor of `wgpu.gpu.request_adapter()`?
* [x] Update release notes.


almarklein:
@Korijn what dou you think about this? Just `import wgpu` and get going.

wgpu/__pyinstaller/hook-wgpu.py
```diff
@@ -1,17 +1,35 @@
...
-# Include backends that we want to Just Work
+# Include the resources subpackage and its contents.
+hiddenimports += ["wgpu.resources"]
+datas += collect_data_files("wgpu", subdir="resources")
+binaries += collect_dynamic_libs("wgpu")
+
+# Include backends. Make sure all our backend code is present,
+# and let PyInstaller resolve imports/dependencies for some.
+datas += collect_data_files(
+    "wgpu", subdir="backends", include_py_files=True, excludes=["__pycache__"]
```
Korijn, line 16: include_py_files=True ? That seems unnecessary?

almarklein: This way, all `.py` files are included, so that wgpy is at least complete, as a library. E.g. if a downstream lib dynamically imports a gui backend, it'll still work (provided the dependencies are fulfilled).

Korijn: I understand that but it's not beneficial to launch time of the distributable. Pyinstaller packages all the python code in compiled form in a binary format that can be loaded in a single disk operation. Including python files as data files is not going to benefit from that.

Application builders will have to statically import gui backends (this covers majority of cases I think) or add a hiddenimports line to their pyinstaller spec file. This is standard practise for pyinstaller, so I don't think we should deviate from that.

almarklein: Ok, I guess I was trying to hard. I will simplify the hook accordingly.

almarklein:
@korijn when you review, can you have a look at the current pyinstaller hook?

Korijn (approved):
The hook looks good to me (and so does everything else!). This was a big one! 🐋

Issue #410, opened 2023-11-09
Make non-spec functionality stand out better


almarklein:
In addition to implementing the WebGPU spec, we also provide additional functionality that wgpu-native offers. It would be a waste not to do this, because it provides more power for desktop use-cases. However, when we add other backends, it is possible that (some of) this functionality is not available in a certain backend, making code that uses these features less portable. And that's sad, because the point of wgpu is to have portable code.

Note that this is not the quite the same as the API differences here: https://wgpu-py.readthedocs.io/en/stable/wgpu.html#differences-from-webgpu. Most these differences don't need extra functionality other than the outlined in the spec.

The cases that I know of that can be a problem are:
* `GPUAdapter.request_device_tracing()`.
* `GPU.enumerate_adapters()`.
* ~The `force_fallback_adapter` argument in `request_adapter()`.~
* Certain features (i.e. extensions) are specific to wgpu-native.

Some ideas to help with this problem:
* Document clearly that this is functionality specific to desktop use.
* Prefix these functions and arguments with something like "ext_".
* Only allow using such functions by enabling it as a kind of "extension". Perhaps this can be merged with the features mechanism already in webgpu? 
* Putting the functionality in the backend module, i.e. not putting it into the main namespace.
* Using backend-specific behavior can only be done in a specific context manager.

In deciding about this, we should also take into account how these functionalities end up in the docs.

Korijn:
This is an appropriate use case for the `NotImplementedError` when a backend is used that doesn't support a particular function :)

almarklein:
@rajveermalviya Ah it is! Thanks for clarifying 😄

almarklein:
> This is an appropriate use case for the NotImplementedError when a backend is used that doesn't support a particular function :)

True, but I also want to discourage people to use this non-standard functionality. What triggered me is the idea that if/when we get wgpu-py working via PyScript (#407), people will want to run examples of pygfx/fastplotlib/etc. in the browser. Any code that uses non-standard stuff will then be less portable.

almarklein:
The solution applied in #408 is to provide backend-specific API as functions in the corresponding backend module. We now have `request_device_tracing` and `enumerate_adapters` in `wgp.backends.wgpu_native`.

almarklein:
I think that with that we can close this issue. One issue to watch out for is that certain features (i.e. extensions) are backend-specific, but when using features you are probably aware of potentially breaking portability anyway.

PR #412, opened 2023-11-12
Implement `enumerate_adapters` for rs-backend

### LINKED ISSUE #409: Add convenience function `enumerate_adapters`
Hi wgpu-py Team,

I can not find `enumerate_adapters()` to just get an idea of all gpu available in a machine. 

is there an alternative way to do that and this is something to add?

Thank you for the feedback.

dskkato:
I originally used wgpu with WebGPU, but recently I came across this library. As mentioned in #409, I also needed to choose an adapter to try out compute shaders, so I implemented it on a trial basis. I'm not sure if it conforms to your coding style, but I would appreciate it if you could review it. If you are already working on something similar, please feel free to close it as appropriate.

Resolves #409

### DIFF
--- a/wgpu/backends/rs.py
+++ b/wgpu/backends/rs.py
@@ -176,6 +176,79 @@ error_handler = ErrorHandler(logger)
 libf = SafeLibCalls(lib, error_handler)
 
 
+def _create_adapter(adapter_id):
+    # H: nextInChain: WGPUChainedStructOut *, vendorID: int, vendorName: char *, architecture: char *, deviceID: int, name: char *, driverDescription: char *, adapterType: WGPUAdapterType, backendType: WGPUBackendType
+    c_properties = new_struct_p(
+        "WGPUAdapterProperties *",
+        # not used: nextInChain
+        # not used: deviceID
+        # not used: vendorID
+        # not used: name
+        # not used: driverDescription
+        # not used: adapterType
+        # not used: backendType
+        # not used: vendorName
+        # not used: architecture
+    )
+
+    # H: void f(WGPUAdapter adapter, WGPUAdapterProperties * properties)
+    libf.wgpuAdapterGetProperties(adapter_id, c_properties)
+
+    def to_py_str(key):
+        char_p = getattr(c_properties, key)
+        if char_p:
+            return ffi.string(char_p).decode(errors="ignore")
+        return ""
+
+    adapter_info = {
+        "vendor": to_py_str("vendorName"),
+        "architecture": to_py_str("architecture"),
+        "device": to_py_str("name"),
+        "description": to_py_str("driverDescription"),
+        "adapter_type": enum_int2str["AdapterType"].get(
+            c_properties.adapterType, "unknown"
+        ),
+        "backend_type": enum_int2str["BackendType"].get(
+            c_properties.backendType, "unknown"
+        ),
+        # "vendor_id": c_properties.vendorID,
+        # "device_id": c_properties.deviceID,
+    }
+
+    # ----- Get adapter limits
+
+    # H: nextInChain: WGPUChainedStructOut *, limits: WGPULimits
+    c_supported_limits = new_struct_p(
+        "WGPUSupportedLimits *",
+        # not used: nextInChain
+        # not used: limits
+    )
+    c_limits = c_supported_limits.limits
+    # H: bool f(WGPUAdapter adapter, WGPUSupportedLimits * limits)
+    libf.wgpuAdapterGetLimits(adapter_id, c_supported_limits)
+    limits = {to_snake_case(k): getattr(c_limits, k) for k in sorted(dir(c_limits))}
+
+    # ----- Get adapter features
+
+    # WebGPU features
+    features = set()
+    for f in sorted(enums.FeatureName):
+        key = f"FeatureName.{f}"
+        i = enummap[key]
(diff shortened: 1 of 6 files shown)


wgpu/backends/rs.py
```diff
@@ -176,6 +176,20 @@ def check_struct(struct_name, d):
 libf = SafeLibCalls(lib, error_handler)
 
 
+def enumerate_adapters():
+    """Return a list of all available adapters."""
+    # The first call is to get the number of adapters, and
+    # the second call is to get the actual adapters
```
almarklein, line 182: Let's mention the option here.
```suggestion
    # The first call is to get the number of adapters, and
    # the second call is to get the actual adapters.
    # Note that the second arg (now NULL) can be a
    # `WGPUInstanceEnumerateAdapterOptions` to filter by backend.
```

wgpu/backends/rs.py
```diff
@@ -176,6 +176,20 @@ def check_struct(struct_name, d):
...
+
+    # H: size_t f(WGPUInstance instance, WGPUInstanceEnumerateAdapterOptions const * options, WGPUAdapter * adapters)
+    adapter_count = libf.wgpuInstanceEnumerateAdapters(
+        get_wgpu_instance(), ffi.NULL, ffi.NULL
+    )
+    adapters = ffi.new("WGPUAdapter[]", adapter_count)
+    libf.wgpuInstanceEnumerateAdapters(get_wgpu_instance(), ffi.NULL, adapters)
+    return [GPUAdapter(adapter) for adapter in adapters]
```
almarklein, line 190: The codegen has trouble with the refactoring of the adapter metadata. A relatively easy way to work around that is to split that part out of `request_adapter()`, into a a new `GPU._create_adapter(adapter_id)`.

Then this here becomes:
```suggestion
    gpu = GPU()
    return [gpu._create_adapter(adapter) for adapter in adapters]
```

wgpu/backends/rs.py
```diff
@@ -261,80 +275,7 @@ def callback(status, result, message, userdata):
             error_msg = error_msg or "Could not obtain new adapter id."
             raise RuntimeError(error_msg)
```
almarklein, line 277:
```suggestion
        return self._create_adapter(adapter_id)

    def _create_adapter(adapter_id):
        ... and undo the deletion below :)
```

almarklein (commented):
Really nice!

The codegen sometimes makes refactorings like this a bit tricky. I added a comment for a way to work around this.

I'm not 100% sure about the API, relates to #410. But the solution you offer here - to make the user go into `wgpu.backends.rs` - seems like a pretty good alternative. We'd have to figure out how to document this function though. 

This also relates somewhat to #408. I propose that we move this PR forward and figure out the #401 stuff later.

wgpu/backends/rs.py
dskkato: When I turned _create_adapter into a member function of the GPU, a TypeError occurred in enumerate_adapters as below. Therefore, I made it a free function.

```sh
_______________________________ test_rs_enumerate_adapters ________________________________

    @mark.skipif(not can_use_wgpu_lib, reason="Needs wgpu lib")
    def test_rs_enumerate_adapters():
        # Get all available adapters
>       adapters = wgpu.backends.rs.enumerate_adapters()

tests/test_rs_basics.py:152: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

    def enumerate_adapters():
        """Return a list of all available adapters."""
        # The first call is to get the number of adapters, and
        # the second call is to get the actual adapters.
        # Note that the second arg (now NULL) can be a
        # `WGPUInstanceEnumerateAdapterOptions` to filter by backend.
    
        # H: size_t f(WGPUInstance instance, WGPUInstanceEnumerateAdapterOptions const * options, WGPUAdapter * adapters)
        adapter_count = libf.wgpuInstanceEnumerateAdapters(
            get_wgpu_instance(), ffi.NULL, ffi.NULL
        )
        adapters = ffi.new("WGPUAdapter[]", adapter_count)
        # H: size_t f(WGPUInstance instance, WGPUInstanceEnumerateAdapterOptions const * options, WGPUAdapter * adapters)
        libf.wgpuInstanceEnumerateAdapters(get_wgpu_instance(), ffi.NULL, adapters)
>       gpu = GPU()
E       TypeError: 'NoneType' object is not callable

wgpu/backends/rs.py:194: TypeError
```

dskkato:
FYI: The example for `wgpuInstanceEnumerateAdapters` I referenced is this -> https://github.com/gfx-rs/wgpu-native/blob/trunk/examples/enumerate_adapters/main.c

almarklein (approved):
LGTM. This is going to be interesting merging with #408 😄   I propose we merge this into `main` and then I'll merge `main` into #408.

Korijn (approved):

almarklein:
Thanks for your contribution!

wgpu/backends/rs.py
almarklein: Ah yes, sorry. This is due to a bug in the `_register_backend` decorator, that will be fixed in #408.

almarklein:
For the record, when #408 is merged, it will be `wgpu.backends.wgpu_native.enumerate_adapters()`.

PR #421, opened 2023-11-17
Replace pkg_resources with importlib.resources

### LINKED ISSUE #420: Replacing pkg_resources.resource_filename
I saw warnings on CI that `pkg_resources.resource_filename` is deprecated, and that `importlib` has functionality that can be used instead. Also see https://importlib-resources.readthedocs.io/en/latest/migration.html

The migration page tells me to install an atexit hook to clear temporary files in case they were copied to the filesystem from a zipfile. But as far as I can tell, wgpu is always installed as a plain directory. And when freezing with PyInstaller, the files are also on the filesystem (in `_internal/wgpu/resources/`). @Korijn do you know whether this (the files always being on the filesystem) is an assumption we can safely make?

almarklein:
Closes  #420

### DIFF
--- a/docs/guide.rst
+++ b/docs/guide.rst
@@ -246,3 +246,9 @@ Freezing apps
 In wgpu a PyInstaller-hook is provided to help simplify the freezing process
 (it e.g. ensures that the wgpu-native DLL is included). This hook requires
 PyInstaller version 4+.
+
+Our hook also includes ``glfw`` when it is available, so code using ``wgpu.gui.auto``
+should Just Work.
+
+Note that PyInstaller needs ``wgpu`` to be installed in `site-packages` for
+the hook to work (i.e. it seems not to work with a ``pip -e .`` dev install).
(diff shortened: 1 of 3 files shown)


almarklein:
This also removes `setuptools` as a runtime dependency. We did not list that as a runtime dependency, which confuses me a bit. Was that not necessary because people almost always have `setuptools` because they have pip?

Korijn (approved):
I had to read the docs on contextlib.ExitStack and importlib.resources to understand what's going on here, but it looks nice!

Korijn:
> This also removes `setuptools` as a runtime dependency. We did not list that as a runtime dependency, which confuses me a bit. Was that not necessary because people almost always have `setuptools` because they have pip?

`pkg_resources` is/was a builtin module

almarklein:
> I had to read the docs on contextlib.ExitStack and importlib.resources

The code was mostly copy-pasting from https://importlib-resources.readthedocs.io/en/latest/migration.html 😉 


> `pkg_resources` is/was a builtin module. Maybe that isn't the case anymore?

Not anymore, at least:
```
>>> pkg_resources
<module 'pkg_resources' from '/Users/almar/.pyenv/versions/3.9.9/lib/python3.9/site-packages/pkg_resources/__init__.py'>
```

Issue #405, opened 2023-11-04
Device objects are not cleared on wgpu-native


almarklein:
This is a potential memory leak. Not a severe one, because people should probably not create many devices. See the memtests we added in #403. It looks like this is the only object that leaks.

I've looked into this, but have not been able to find a way to fix this in wgpu-py. 

Plan / ideas:
* [ ] Wait until we update to wgpu-native 0.18.0.1, with luck its fixed in a newer version of wgpu-core.
* [ ] Check if this issue is already known.
* [ ] Create a minimal example to reproduce this using wgpu-native directly, and create an issue for it.

Vipitis:
I think I am running into a memory leak as well, after running something between 50 and 100 different shadertoys through the utility (just creating the class, not even rendering a single frame), I can see my vram usage grow to about 4GB and then the python process exits. It doesn't throw any exceptions, nor are there any rust panics and it seems like it just finished (it's a kernel crash when when doing this inside a jupyter notebook).

It could be due to my Intel A750 GPU and their clearly buggy compute runtime, so I will double check the behavior on my laptop with nvidia GPU tomorrow.

Is there a some method to clear up vram apart from killing the python process?

Korijn:
Is the CPU ram also growing? Or just the VRAM?

almarklein:
You could try to modify the `shadertoy.py` in `_prepare_render()`. Where it requests an adapter and then a device, it could use `get_default_device()` (from `wgpu.utils.device`). I think that would make sense anyway, and it might just solve your leak?

Vipitis:
> Is the CPU ram also growing? Or just the VRAM?

there was no increase in CPU RAM.

> You could try to modify the shadertoy.py in _prepare_render(). Where it requests an adapter and then a device, it could use get_default_device() (from wgpu.utils.device). I think that would make sense anyway, and it might just solve your leak?

I changed [this specific line](https://github.com/pygfx/wgpu-py/blob/49b4ae4503bd28e0e9d40424d8803b512758e035/wgpu/utils/shadertoy.py#L306) with your recommendation and it seems to be an improvement. There is no longer any uncontrolled vram rise. It looks steady for the most part. I still encountered a crash after 450 shaders, but will investigate a bit more. 

I can make a PR for that (if there isn't any more changes given the other PRs right now).

almarklein:
That's good to hear! Perhaps the crash after 450 shaders is not related to the leak. Still odd though. A PR for this part of the fix would be great!

[...]

PR #429, opened 2023-11-20
Update to wgpu-native v0.18.1.1

### LINKED ISSUE #352: [Qt, Windows] Crash when switching from monitor to monitor with different DPI values
When I drag the window from one monitor to another, with different DPI settings, pygfx crashes:

```
❯ python .\examples\feature_demo\colormap_channels.py
QWindowsWindow::setDarkBorderToWindow: Unable to set dark window border.
Requested size 640x480 is outside of the supported range: Extent3d { width: 1600, height: 1200, depth_or_array_layers: 1 }..=Extent3d { width: 1600, height: 1200, depth_or_array_layers: 1 }
Uncaught WGPU error (Unknown):
Outdated
thread '<unnamed>' panicked at 'Unable to end render pass: RenderPassError { scope: Pass((0, 122, Vulkan)), inner: MissingAttachments }', src\command.rs:298:10
```

almarklein:
closes #352

* [x] Update the header files.
* [x] Run codegen, fix parsing.
* [x] Handle fixme's.
* [x] Refactor canvas context, removing swap chain concepts in favor of wgpuSurfaceXX (check all).
* [x] use new WGPUCompositeAlphaMode?
* [x] What is FifoRelaxed present mode?
* [x] Update tests and examples
* [x] Fix that `CanvasContext`'s  don't clean up properly. 
* [x] ~Some calls can pass null pointer, check `WGPU_NULLABLE` and check for it.~  that nullable texture only applied to that particular situation, and we've got that covered.
* [x] Check #405 -> nope, still leaks
* [x] Check #352 -> yes! 
* [x] Update release notes. 
* [x] Prepare a pygfx PR: https://github.com/pygfx/pygfx/pull/617
  * [x] Adjust to API changes
  * [x] Perhaps shader changes. *Not needed*
  * [x] Can we use features like `Float32Filterable` rather than `WGPUNativeFeature_TextureAdapterSpecificFormatFeatures` -> not yet

After this is merged:
* [ ] Release wgpu-py.
* [ ] Merge pygfx-pr, and release pygfx.


almarklein:
As for #352, it looks like the warnings (when dragging a window from one monitor to another on Windows+Qt) cannot be avoided without breaking window resizing on other systems. Nevertheless, the window is not in a broken state anymore where a resize is needed to get it working again. In other words, the error messages are annoying but otherwise harmless.

almarklein:
almost done now 😅

almarklein:
> Can we use features like `Float32Filterable` rather than `WGPUNativeFeature_TextureAdapterSpecificFormatFeatures`

The reason why I want this is that the latter is a native-only feature (not part of the webgpu spec). This means that using it also unlocks other stuff, and accidentally using that may cause downstream code to become incompatible with WebGPU.

I tried adding it to wgpu-native (https://github.com/gfx-rs/wgpu-native/pull/322), but the feature is simply not implemented in wgpu-core yet. I'm looking into adding it there, but it will take some time before that lands in a new wgpu-native release.

almarklein:
Fixed the pypy memytest with another well-placed `gc.collect` 😉

CHANGELOG.md
```diff
@@ -17,6 +17,27 @@ Possible sections in each release:
 * Security: in case of vulnerabilities.
 
 
+### [v0.13.0] - TBD
+
+IN this release we
```
Korijn, line 22: I would just remove this line

almarklein: Woops. I think I started a sentence there, but made it a bulletpoint.

Korijn (approved):

PR #434, opened 2023-11-24
update doc of `get_bind_group_layout`

dskkato:
After #432, wgpu-native now raises error instead of immediate abort.


Korijn (approved):

almarklein:
Nice catch!

PR #441, opened 2023-12-01
Support turning on wgpu validation flag

almarklein:
* [x] Turn on the flag.
* [ ] Make it conditional (maybe en env var?)
* [ ] Figure out what it does 😆 
    * I *think* this turns on the Vulkan validation layers, if you have the Lunar SDK installed.
    * On MacOS with Metal I see no difference in the (debug) logs.
    * ... I'm having trouble running on Vulkan on MacOS.
    * Windows: With the Vulkan Configurator on, the validation layers are active, and it prints (to sys stdout / file 0) that the validation layer is active ... but it also does this without the change in this pr.

https://github.com/gfx-rs/wgpu-native/issues/327


almarklein:
Closing, because I don't fully understand (yet) what it does. And from what I've seen from it, it does not seem to help much for us (at least not in the case that we were struggling with).

PR #443, opened 2023-12-04
Add texture input to Shadertoy util (iChannel0, ...)

Vipitis:
Task for this PR:
- [x] add bindings to wglsl
- [x] add bindings to glsl
- [x] add ShadertoyChannel class
- [x] add full example/test with multiple inputs
- [x] fix binding conflict when running multiple Shadertoy instances in the same python file (like during pytest)
***
as it seems to be the most common issue according to #426 (even though the filtered dataset has no shadertoys with inputs declared), the built in `iChannel0 .. 3` samplers are still used. They output 0 if nothing is attached to the channels. 

The Shadertoy website allows for various kinds of inputs, be it textures, cubemaps, buffers (other Shadertoys) or even webcam and audio. The most basic support will be to resolve "Unknown Variable iChannel0" by having a sort of default input of 0 that can be sampled. Texture input would be most useful, so that is my aim with this PR.

**So oddities I encountered:**
- only seems to work with unit8 (not float32)
- has to be flipped vertically?

**Some ideas I have had:**
- a `ShadertoyChannel` struct/class that can be used to hold all the various settings for the sampler as well as useful defaults (like the 0 texture), do conversions (as array, dtype)
- would Buffers be just the code or a whole instance of the Shadertoy class?

**I need advice on:**
- can this be done without adding numpy as a dependency
- use a single texture array or individual bindings for channels 0..3 with individual samplers
- can glsl be done at this point or is there releases required from wgpu/naga (bunch of glsl-in commits recently)
- how to structure the user facing args to Shadertoy if texture inputs are used?

I will add some more commits but would welcome some guidance or references on the above points. 

<details>

<summary>How channel inputs are structured on the Shadertoy website</summary>

```json
{
     "inputs": [
    {
     "id": "4sfGRn",
     "filepath": "/media/a/fb918796edc3d2221218db0811e240e72e340350008338b0c07a52bd353666a6.jpg",
     "type": "texture",
     "channel": 0,
     "sampler": {
      "filter": "mipmap",
      "wrap": "repeat",
      "vflip": "true",
      "srgb": "false",
      "internal": "byte"
     },
     "published": 1
    },
    { "..." 
    }
  ],
```
</details>

### DIFF
--- a/wgpu/utils/shadertoy.py
+++ b/wgpu/utils/shadertoy.py
@@ -5,8 +5,6 @@ import wgpu
 from wgpu.gui.auto import WgpuCanvas, run
 from wgpu.gui.offscreen import WgpuCanvas as OffscreenCanvas, run as run_offscreen
 
-import numpy as np
-
 vertex_code_glsl = """#version 450 core
 
 layout(location = 0) out vec2 uv;
@@ -144,9 +142,21 @@ var<uniform> input: ShadertoyInput;
 
 @group(0) @binding(1)
 var i_channel0: texture_2d<f32>;
+@group(0) @binding(3)
+var i_channel1: texture_2d<f32>;
+@group(0) @binding(5)
+var i_channel2: texture_2d<f32>;
+@group(0) @binding(7)
+var i_channel3: texture_2d<f32>;
 
 @group(0) @binding(2)
-var r_sampler: sampler;
+var sampler0: sampler;
+@group(0) @binding(4)
+var sampler1: sampler;
+@group(0) @binding(6)
+var sampler2: sampler;
+@group(0) @binding(8)
+var sampler3: sampler;
 
 @fragment
 fn main(in: Varyings) -> @location(0) vec4<f32> {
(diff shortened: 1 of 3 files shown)


almarklein:
> only seems to work with unit8 (not float32)

Float32 textures are not filterable by default. Need to enable a device feature for this. For now you can use "texture_adapter_specific_format_features". We're working on support for the official "float32-filterable" feature, see
https://github.com/pygfx/pygfx/issues/618

> has to be flipped vertically?

Depending on the definition of the coordinate frame images have a tendency of being upside down, yeah.

> can this be done without adding numpy as a dependency

Python's `memoryview` comes a long way.

> use a single texture array or individual bindings for channels 0..3 with individual samplers

I'm not that much into the details of shadertoy, but I expect the latter to be more flexible, e.g. easier to replace with something else.

> can glsl be done at this point or is there releases required from wgpu/naga (bunch of glsl-in commits recently)

What do you mean with "be done"?  As for recent commits in wgpu-core, the new release strategy of wgpu-native is that a new release is only made when there's a wgpu-core release. At that point a branch is created so there can be dot-releases. In other words, it may take a little while for commits that are made to wgpu-core now to land in a wgpu-py.

> how to structure the user facing args to Shadertoy if texture inputs are used?

I suppose that mostly up to the "shadertoy team" :)  which I think is you and @panxinmiao

Vipitis:
hey @almarklein and @Korijn do you have any feedback on this?

Korijn:
I'm assuming you tested all existing examples as well and nothing broke?

In general the changes look good to me.

I don't really have any advice for the concerns you shared in the description.

wgpu/utils/shadertoy.py
```diff
@@ -222,6 +240,62 @@ def __setitem__(self, key, val):
                 m[i] = val[i]
 
 
+class ShadertoyChannel:
+    """
+    Represents a shadertoy channel. It can be a texture.
+    Parameters:
+        data (memoryview): will be converted to memoryview. For example read in your images using ``np.asarray(Image.open("image.png"))``
```
almarklein, line 247:
```suggestion
        data (array-like): will be converted to memoryview. For example read in your images using ``np.asarray(Image.open("image.png"))``
```

almarklein (commented):
Only one small comment.

Korijn (approved):

PR #445, opened 2023-12-06
update wgpu-native version

### LINKED ISSUE #187: Support for Raspberry Pi (and other linux_arm64)
* [x] User's system:
  * Don't forget to run `sudo apt install vulkan-tools mesa-vulkan-drivers`. 
* [ ] Here wgpu-py
  * [x] cffi must provide binary wheels for linux_arm64.
  * [x] Detect raspberry pi so that the download script can download the correct libs. 
  * [x] Make our `get_surface_id_from_canvas` works (with a bit of luck it does).
  * [ ] Build the appropriate wheels on CI.
* [x] wgpu-native
  * [ ] Build the binary on CI: [issue 142](https://github.com/gfx-rs/wgpu-native/issues/142) -> [pr 148](https://github.com/gfx-rs/wgpu-native/pull/148).  Remaining issue is that the glibc version does not match.
* [x] wgpu-core
  * Should, work, but there are some issues.
  * https://github.com/gfx-rs/wgpu/issues/1069

dskkato:
This is a follow-up to the `0.18.1.3` release of wgpu-native. See https://github.com/gfx-rs/wgpu-native/pull/326#issuecomment-1840189996

## Changes

- update wgpu-native version from `0.18.1.2` to `0.18.1.3`
- use "aarch64" for mac and linux instead of arm/arm64

Verified to work on Raspberry Pi OS (64-bit).

Resolves #187.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -226,14 +226,42 @@ jobs:
     - name: Build wheels
       uses: pypa/cibuildwheel@v2.16.2
       env:
-          CIBW_MANYLINUX_X86_64_IMAGE: quay.io/pypa/manylinux_2_28_x86_64
-          CIBW_ARCHS_LINUX: x86_64
-          CIBW_SKIP: cp39-musllinux_x86_64
+        CIBW_MANYLINUX_X86_64_IMAGE: quay.io/pypa/manylinux_2_28_x86_64
+        CIBW_ARCHS_LINUX: x86_64
+        CIBW_SKIP: cp39-musllinux_x86_64
       with:
         output-dir: dist
     - name: Twine check
       run: |
           twine check dist/*
+    - name: Upload distributions
+      uses: actions/upload-artifact@v2
+      with:
+        path: dist
+        name: dist
+
+  # Thees release builds uses QEMU so that we can build wheels for arm64.
+  # We build wheels and upload the wheel as an artifact, but we don't test them here.
+  qemu-release-builds:
+    name: Build wheels on ubuntu-latest with QEMU
+    timeout-minutes: 10
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+    steps:
+    - uses: actions/checkout@v3
+    - name: Set up QEMU
+      uses: docker/setup-qemu-action@v3
+      with:
+        platforms: arm64
+    - name: Build wheels
+      uses: pypa/cibuildwheel@v2.16.2
+      env:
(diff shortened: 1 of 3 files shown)


Korijn:
Macos ARM is apple Silicon right? We don't want to drop support for apple Silicon?

dskkato:
@Korijn  I do not intend to discontinue the support for Apple Silicon's macOS, simply a follow-up to the wgpu-native binary name change. See this comment : https://github.com/gfx-rs/wgpu-native/pull/326#discussion_r1409500755

But, I do not have Apple Silicon's macOS on hand, so I cannot confirm if this is causing any problems. If it is causing problems, could you please let me know?

almarklein:
Also see https://github.com/gfx-rs/wgpu-native/releases, where one can see the new names.

We should also update the CI so the Linux arm wheels are also created.

Korijn:
I was mostly confused because the python platform tag for the .whl file is officially arm64 on macos. Can you confirm that the proper binary is included in the arm64 .whl file in the build artifacts on this PR?

almarklein:
> Can you confirm that the proper binary is included in the arm64 .whl file in the build artifacts on this PR?

I just confirmed.

The renaming of "arm64" to "aarch64" only affects the zipfile that is downloaded from the wgpu-native release page. The library is just called "libwgpu_native-release.dylib", so the wheel building is not affected at all by the new name of the zipfile.

Korijn (approved):

Korijn:
Nice one! 🚀

dskkato:
I added the `qemu-release-builds` job because the aarch64 build necessitates QEMU. Additionally, Twine checks need to be disabled.

You can check the build artifacts here https://github.com/dskkato/wgpu-py/releases/tag/v0.13.1 .

(though this PR was merged before I could finish writing this description.)

Korijn:
> I added the `qemu-release-builds` job because the aarch64 build necessitates QEMU.

Yeah that makes sense!

> Additionally, Twine checks need to be disabled.

Why is that? I didn't notice you left that out.

dskkato:
It's because the built wheel is for aarch64 but twine runs on the host x86_64 env. Perhaps I should have installed twine inside the aarch64 docker container so that we could check the artifact.

Korijn:
Twine check is used to verify the metadata such as the long description (readme on pypi). I don't expect any differences between platforms for that, so I think it's fine like this. 👍

dskkato:
Oops, I was mistaken about twine not working for aarch64 artifacts, I just thought it was another error I encountered while debugging the CI build.

almarklein:
Just pushed a new [release](https://github.com/pygfx/wgpu-py/releases), so the new binary wheels are available via pip!

PR #450, opened 2023-12-13
Implement timestamp QuerySet

wpmed92:
This PR adds the following functionality

- `device.create_query_set`
- `timestamp_writes` support in `command_encoder.begin_compute_pass`
- `command_encoder.resolve_query_set`
- add a new example (`examples/compute_timestamps.py`) that demonstrates the usage of the newly added features

Tests added

- `tests_mem/test_objects.py` to check release of `QuerySet`s
- `tests/test_wgpu_native_query_set.py` to test the newly added functionality

### DIFF
--- a/tests/test_wgpu_native_query_set.py
+++ b/tests/test_wgpu_native_query_set.py
@@ -1,0 +1,151 @@
+import wgpu.utils
+
+from testutils import run_tests, can_use_wgpu_lib
+from pytest import mark
+
+
+@mark.skipif(not can_use_wgpu_lib, reason="Needs wgpu lib")
+def test_query_set():
+    shader_source = """
+    @group(0) @binding(0)
+    var<storage,read> data1: array<f32>;
+
+    @group(0) @binding(1)
+    var<storage,read_write> data2: array<f32>;
+
+    @compute
+    @workgroup_size(1)
+    fn main(@builtin(global_invocation_id) index: vec3<u32>) {
+        let i: u32 = index.x;
+        data2[i] = data1[i] / 2.0;
+    }
+    """
+
+    n = 1024
+    data1 = memoryview(bytearray(n * 4)).cast("f")
+
+    for i in range(n):
+        data1[i] = float(i)
+
+    adapter = wgpu.gpu.request_adapter(power_preference="high-performance")
+    device = adapter.request_device(
+        required_features=[wgpu.FeatureName.timestamp_query]
+    )
+
+    assert repr(device).startswith("<wgpu.backends.wgpu_native.GPUDevice ")
+
+    cshader = device.create_shader_module(code=shader_source)
+
+    # Create buffer objects, input buffer is mapped.
+    buffer1 = device.create_buffer_with_data(data=data1, usage=wgpu.BufferUsage.STORAGE)
+    buffer2 = device.create_buffer(
+        size=data1.nbytes, usage=wgpu.BufferUsage.STORAGE | wgpu.BufferUsage.COPY_SRC
+    )
+
+    # Setup layout and bindings
+    binding_layouts = [
+        {
+            "binding": 0,
+            "visibility": wgpu.ShaderStage.COMPUTE,
+            "buffer": {
+                "type": wgpu.BufferBindingType.read_only_storage,
+            },
+        },
+        {
+            "binding": 1,
+            "visibility": wgpu.ShaderStage.COMPUTE,
+            "buffer": {
+                "type": wgpu.BufferBindingType.storage,
+            },
+        },
+    ]
+    bindings = [
+        {
+            "binding": 0,
+            "resource": {"buffer": buffer1, "offset": 0, "size": buffer1.size},
+        },
+        {
+            "binding": 1,
+            "resource": {"buffer": buffer2, "offset": 0, "size": buffer2.size},
+        },
+    ]
+
+    bind_group_layout = device.create_bind_group_layout(entries=binding_layouts)
+    pipeline_layout = device.create_pipeline_layout(
+        bind_group_layouts=[bind_group_layout]
+    )
+    bind_group = device.create_bind_group(layout=bind_group_layout, entries=bindings)
+
+    compute_pipeline = device.create_compute_pipeline(
+        layout=pipeline_layout,
+        compute={"module": cshader, "entry_point": "main"},
+    )
+
+    query_count = 2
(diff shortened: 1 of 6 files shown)


examples/compute_noop.py
```diff
@@ -59,7 +59,8 @@
 
 # %% Create device
 # Create device and shader object
-device = wgpu.utils.get_default_device()
+adapter = wgpu.gpu.request_adapter(power_preference="high-performance")
+device = adapter.request_device(required_features=[wgpu.FeatureName.timestamp_query])
```
Korijn, line 63: Let's keep the compute_noop example minimal if we can. Can you move your timestamp query example to its own example module?

wgpu/_classes.py
```diff
@@ -2024,23 +2024,27 @@ def messages(self):
...
 class GPUQuerySet(GPUObjectBase):
-    """TODO"""
-
-    # IDL: undefined destroy();
-    def destroy(self):
-        """Destroy the queryset."""
-        raise NotImplementedError()
+    def __init__(self, label, internal, type, count):
+        super().__init__(label, internal, None)
+        self._label = label
```
almarklein, line 2029: All objects created by the device have a reference to the device, and label is already stored in `self._label` by the `super().__init__`.
```suggestion
    def __init__(self, label, internal, device, type, count):
        super().__init__(label, internal, device)
```

wgpu/backends/wgpu_native/_api.py
```diff
@@ -2495,7 +2513,15 @@ def finish(self, *, label=""):
...
     ):
-        raise NotImplementedError()
+        # H: void f(WGPUCommandEncoder commandEncoder, WGPUQuerySet querySet, uint32_t firstQuery, uint32_t queryCount, WGPUBuffer destination, uint64_t destinationOffset)
+        libf.wgpuCommandEncoderResolveQuerySet(
+            self._internal,
+            query_set._internal,
+            int(first_query),
+            int(query_count),
+            destination._internal,
+            destination_offset,
```
almarklein, line 2523:
```suggestion
            int(destination_offset),
```

wgpu/backends/wgpu_native/_api.py
```diff
@@ -2841,14 +2867,16 @@ def _destroy(self):
...
-    def destroy(self):
+    def _destroy(self):
         if self._internal is not None and libf is not None:
             self._internal, internal = None, self._internal
             # H: void f(WGPUQuerySet querySet)
             libf.wgpuQuerySetRelease(internal)
 
+    # FIXME: new method to implement
+    def destroy(self):
+        raise NotImplementedError()
```
almarklein, line 2878: Can just call `self._destroy()` here.

wgpu/_classes.py
```diff
@@ -2024,23 +2024,27 @@ def messages(self):
...
+        return self._type
 
     # IDL: readonly attribute GPUSize32Out count;
     @property
     def count(self):
-        """The type of the queries managed by this queryset."""
+        """The number of the queries managed by this queryset."""
+        return self._count
+
+    # FIXME: new method to implement
```
almarklein, line 2045: This implementation is fine, so the fixme can be removed.

almarklein (changes requested):
Nice work!

examples/compute_noop.py
wpmed92: I moved it to `examples/compute_timestamps.py`

wgpu/_classes.py
```diff
@@ -2024,23 +2024,31 @@ def messages(self):
...
     @property
     def count(self):
-        """The type of the queries managed by this queryset."""
+        """The number of the queries managed by this queryset."""
+        return self._count
+
+    # IDL: undefined destroy();
+    def destroy(self):
+        """Destroy this QuerySet."""
         raise NotImplementedError()
```
Korijn, line 2052: Can't ee implement this?

almarklein: This is the base class. All `destroy()` methods in this module raise NotImplemented, and are implemented in `wgpu_native/_api.py`. I think it makes sense for these here to just call `_destroy()` instead, but since there are multiple such cases, we can leave that for another day imo.

almarklein: To provide some context: almost all classes have a `_destroy()` method that gets automatically called when the object is deleted by the garbage collector (via `__del__`). But the WebGPU spec also gives a few objects an explicit destroy method, which is what we have here.

almarklein (changes requested):
LGTM!

One final request: could you also add a unit test? That touches the newly added code?

I initially regarded that the example would be sufficient, but we (currently) cannot include compute examples in our example-tests, because we expect a canvas present.

wpmed92:
@almarklein Sure, I'll add tests! Should I add new test files, thinking of _test_wgpu_native_device_ and _*_command_encoder_ or should I add the new tests to an existing file, say _buffer?

almarklein:
A new file `test_wgpu_native_query_set` sounds good. We have quite some tests that touch a lot of API because you need so much stuff to render an image :)  For stuff like added in this PR, it makes sens to have something relatively self-contained. Basically a minimalistic version of the example that you already contributed.

wgpu/_classes.py
Korijn: Makes sense!

tests/test_wgpu_native_query_set.py
```diff
@@ -0,0 +1,149 @@
+import wgpu.utils
+
+from testutils import run_tests
+
+
+def test_query_set():
```
Korijn, line 6: I think this needs `@mark.skipif(not can_use_wgpu_lib, reason="Needs wgpu lib")`?

wpmed92: Sorry, I missed that. Just added.

wpmed92:
> A new file `test_wgpu_native_query_set.py` sounds good. We have quite some tests that touch a lot of API because you need so much stuff to render an image :) For stuff like added in this PR, it makes sens to have something relatively self-contained. Basically a minimalistic version of the example that you already contributed.

I added the test.

Korijn (approved):

almarklein (approved):

almarklein:
Thanks!

wpmed92:
@almarklein When can we expect the new release to be out?

almarklein:
We can run a release now: #454

wpmed92:
Great, thank you! 🚀

PR #461, opened 2024-02-01
Enable manual CI trigger

almarklein:
When CI errors in a PR, it's sometimes convenient to trigger CI for `main` to see if its due to changes in the pr or something external that changed.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,6 +1,7 @@
 name: CI
 
 on:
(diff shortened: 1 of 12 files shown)


Korijn (approved):

almarklein:
This also fixes the Windows wheel build.

The other errors are due to a change in Black. We can ignore these for now because they are fixed in #458.

Korijn (approved):

PR #472, opened 2024-03-05
Update to latest IDL

almarklein:
Only affects docs a bit.

(Was looking into webgpu api and wanted to see if anything significant had changed. Barely a change, so I figured we might as well update it now, and thereby reduce noise in a future update.)


Korijn (approved):

Korijn:
Since browsers have landed their implementations already I don't expect it to change significantly anymore. Do you have any sources that indicate otherwise?

PR #479, opened 2024-03-22
Don't warn in present()

almarklein:
The warning was intended to help people who forgot to call `get_current_texture()`. But it looks like it does more harm then help, because when an exception happens in the draw function, the warning is printed below, making it easy to interpret that *that* is the error message. See e.g. https://github.com/pygfx/pygfx/issues/682


Korijn (approved):

Korijn:
You could also lower the logging level to debug. But anyway, I approve :)

almarklein:
Lol, silly how I did not think of that. I think making it a debug message is much better. I also added some context in the message.

Korijn (approved):

Issue #430, opened 2023-11-22
Use own created glfw window.


MackyK:
I'm creating my own window with glfw. When I look at wgpu example its using wgpucanvas which creates the glfw window for you. How do I render with the window I created.

almarklein:
Interesting! Could you tell us more about the use-case? Why can't you use the window that `GlfwWgpuCanvas`  creates for you? 

The wgpu canvas needs to hook up with the window to keep track of its size, when to redraw, etc. Knowing more about the why, can help us determine if/how this case can be supported.

MackyK:
I'm using glfw to create the window then getting the HWND ptr to then changing the style of the window with pywin32. I need to have complete control of the styling of the window as most APIs that create the window for you don't allow this. Most windows created use the WS_CAPTION window style. I need to be able to create a window with the WS_THICKFRAME styling.

If wgpu exposed the GlfwWindow with WgpuCanvas. I could do the styling no problem.


```
canvas = WgpuCanvas()
window = wgpu.get_glfw_window(canvas)
hwnd = glfw.get_win32_window(window)

style = win32api.GetWindowLong(hwnd, win32con.GWL_STYLE)
style |= win32con.WS_THICKFRAME
style &= ~win32con.WS_CAPTION
win32api.SetWindowLong(hwnd, win32con.GWL_STYLE, style)
```

almarklein:
I'm pretty sure that [canvas.get_window_id](https://wgpu-py.readthedocs.io/en/stable/generated/wgpu.gui.WgpuCanvasInterface.html#wgpu.gui.WgpuCanvasInterface.get_window_id) is the HWND that you're looking for. Incidentally wgpu needs it too to create a surface.

Otherwise you can try using `canvas._window` which is the glfw window reference. If you need this one, we can make it a public property to support use-cases like this.

Another option is to be able to pass a window object like `GlfwWgpuCanvas(... window=my_window)`. This may better if you'd need to initialize the window in a specific way.

MackyK:
In all honesty I'd rather just have the bindings of wgpu made to work with python. I'd prefer to do most of the stuff myself to a degree. It's better for learning and It's easier to get help.

I hope this isn't coming across as rude. I'm grateful there is even a module for this on python. I hoping for something more barebones in the future.

Korijn:
[...]

PR #480, opened 2024-03-24
Manual GLFW integration

Korijn:
Direct integration of glfw and wgpu-py without using the wgpu.gui Canvas abstraction/class hierarchy.

This is a demonstration for hardcore users that need total low-level control.

Changes:

* Adds `examples/triangle_glfw_direct.py`
* Refactors `examples/triangle.py` and `wgpu/gui/glfw.py` to make some utility code more easily accessible.

Background: 

* I was thinking about https://github.com/pygfx/pygfx/issues/495 and decided to see if I can implement a closed loop render mode with wgpu-py and glfw.
* Turns out it was pretty challenging so this example should support users looking to do the same, for example #430
* The main thing making this complicated is that instead of providing direct bindings to webgpu-native, we directly map the webgpu-native API to the IDL-standardized webgpu API
* We could consider "splitting" the webgpu-native backend in 2 parts, one being autogenerated direct python bindings, and one being the translation layer to the IDL-standardized webgpu API, I documented this idea here: #487

Note: I guess another question is if this can be done with a pygfx renderer.

image


almarklein:
Yes, the `WgpuCanvasInterface` class defines what is needed. It can be used as an interface (as the name suggests)- it does not have to be subclassed.

https://github.com/pygfx/wgpu-py/blob/570738c9b9dcf49e734cf7cea7cb0cf925a785f9/wgpu/gui/base.py#L8-L60


We also have a test to make sure this works:

https://github.com/pygfx/wgpu-py/blob/570738c9b9dcf49e734cf7cea7cb0cf925a785f9/tests/test_gui_glfw.py#L159-L163


The `_vsync` attribute being checked is something we should probably refactor. Perhaps make it part of the `WgpuCanvasInterface` too.

almarklein:
To respond to your questions:

> I can't render a triangle to the screen using the wgpu_native backend with glfw, unless I re-implement the Canvas API. 

Yes, some parts. Perhaps we can improve our docs on how this works. The example in this PR will certainly help.

> That means I am also bound to its design choices in terms of event loop and abstraction.

I don't think this is the case.

> Is our Canvas abstraction and it's API based on any public specs such as webgpu or something by w3c? I don't think so right?

No.

Korijn:
I think I can see what your idea is now after having had some more time to play with it. I could implement a minimal Canvas class that exposes only the needed methods, but keep control flow out of it. That's cool. I'll try to do that in this branch next.

almarklein:
> I think I can see what your idea is now after having had some more time to play with it.

Definitely room to improve the docs in this regard then 😅

Korijn:
Well, I think it's just unexpected that the Canvas _interface_ from the gui package defines an API that you _must_ use to work with the rust backend.

In essence that means it's actually a type defined by the rust backend. That's the unclear part.

examples/triangle_glfw_direct.py
```diff
@@ -0,0 +1,77 @@
...
+from wgpu.gui.glfw import get_surface_info, get_physical_size
+from wgpu.utils.device import get_default_device
+
+
+sys.path.insert(0, str(Path(__file__).parent))
+
+from triangle import setup_draw  # noqa: E402
+
+
+class GlfwCanvas:
```
almarklein, line 26: I wonder whether subclassing from `WgpuCanvasInterface` would make this even easier. That class has docs explaining what must be implemented. And hides the context abstraction a bit.

I think you prefer a completely new class because inheriting from *something*  already feels buying into our GUI layer?

almarklein (approved):
I don't think we can get it much more minimal than this.

The only thing I wonder is if 
```py
class Glf
```

examples/triangle_glfw_direct.py
almarklein: > It's defined in gui, whereas I feel belongs in the backend: `get_surface_info` and `get_physical_size` are required for the wgpu-native backend to work(!), so they're bound to that backend

I thought about this and find that I disagree. The `get_surface_info()` method produces the stuff that the backend needs to create a surface, but in a GUI-specifc way. So ... it 'belongs' to both the backend and the gui, IMO.

Practically, I think it makes more sense to put in near the GUI, because if someone want to use wgpu with another GUI that we don't support (yet), that person can implement `get_surface_info` and it will work.

And if we add another backend (alternative to wgpu-native) then the dict returned by `get_surface_info()` can be populated with additional fields to support that backend.

As for `get_physical_size()` I think it's likely that other backends would need it too.

----

 > It's got `get_context` defined on it, which feels like it isn't necessary (as demonstrated here)

This is because the WebGPU API specifies it as such. That said, this is a piece that is rather JS/DOM specific, and I think it's reasonable if we deviate from the WebGPU spec here.

---

Maybe something in this direction:

* Make `get_surface_info()` the sole way by which a canvas communicates its wishes to the context.
  * Include physical size.
  * Optionally include a vsync preference.
  * Possibly tell the context that the canvas is offscreen.
* The `GPUCanvasContext` does not necessarily need a canvas, just a `get_surface_info`function is enough.
* Provide an alternative way to create a context. Maybe `wgpu.gpu.create_canvas_context(get_surface_info_func, canvas=None)`?
* Remove the custom `wgpu.gui.offscreen.GPUCanvasContext`, but communicate the canvas needs via `get_surface_info`.

Then this example can truly be written without the word canvas in it (if you don't count `create_canvas_context()`).

PR #481, opened 2024-03-26
Provide more control to select a specific GPU

almarklein:
## Considerations

The idea is to allow users to specify a specific GPU, which is particularly useful in a multi-GPU setup. The tricky bits are that the kinds of GPU's differ widely between systems, and there's a human choice in the loop. You don't want to prompt the user for input each time a script is run. Instead you want a simple way to present the user with the options on the current system, and do this once. Then in subsequent runs allow the user to specify one of these options using a string.

This PR simply makes it easier to enumerate adapters and get a description of each. The user is responsible for the selection mechanism, because this is always fuzzy, and we don't know the best approach yet, see #482. When we do, we will add a builtin selection mechanism to either pygfx or wgpu.utils.

## Tasks

* [x] Add `gpu.enumerate_adapters()`.
* [x] Add `adapter.summary`, which is a single-line string that describes the adapter.
* [x] Add a unit test.
* [x] Update docs.

### DIFF
--- a/wgpu/backends/wgpu_native/_api.py
+++ b/wgpu/backends/wgpu_native/_api.py
@@ -176,7 +176,12 @@ libf = SafeLibCalls(lib, error_handler)
 
 class GPU(classes.GPU):
     def request_adapter(
-        self, *, power_preference=None, force_fallback_adapter=False, canvas=None
+        self,
+        *,
+        gpu_preference=None,
+        power_preference=None,
+        force_fallback_adapter=False,
+        canvas=None,
     ):
         """Create a `GPUAdapter`, the object that represents an abstract wgpu
         implementation, from which one can request a `GPUDevice`.
@@ -184,6 +189,13 @@ class GPU(classes.GPU):
         This is the implementation based on wgpu-native.
 
         Arguments:
+            gpu_preference (str): keywords (separated by commas) to select an adapter.
+                Matches on device name, backend type
+                (metal/vulkan/d3d12/opengl) and adpater type
+                (CPU/IntegratedGPU/DiscreteGPU). Case insensitive. Multiple keywords
+                can be given, in order of preference. The best match is selected,
+                but no guarantee is made whether the returned adapter
+                matches the preference. If given, it overrides power_preference.
             power_preference (PowerPreference): "high-performance" or "low-power".
             force_fallback_adapter (bool): whether to use a (probably CPU-based)
                 fallback adapter.
@@ -205,7 +217,7 @@ class GPU(classes.GPU):
 
         # Try to read the WGPU_BACKEND_TYPE environment variable to see
         # if a backend should be forced.
-        force_backend = os.getenv("WGPU_BACKEND_TYPE", None)
+        force_backend = os.getenv("WGPU_BACKEND_TYPE", "").strip()
         backend = enum_str2int["BackendType"]["Undefined"]
         if force_backend:
             try:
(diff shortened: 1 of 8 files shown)


almarklein:
@Korijn @kushalkolar thoughts about the current API?

The idea is that in pygfx there will be a `set_gpu_preference()` similar to our current `enable_wgpu_features()` that must be called before the first renderer is created.

Fastplotlib can provide a similar function that calls into the pygfx one.

kushalkolar:
> The idea is that in pygfx there will be a `set_gpu_preference()` similar to our current `enable_wgpu_features()` that must be called before the first renderer is created.
> 
> Fastplotlib can provide a similar function that calls into the pygfx one.

Sounds good to me!

Korijn:
> @Korijn @kushalkolar thoughts about the current API?

No objections here.

wgpu/backends/wgpu_native/_api.py
```diff
@@ -218,7 +230,41 @@ def request_adapter(
...
+                gpu_keywords.insert(0, force_backend.lower())
+            # Score all adapters
+            for adapter in adapters:
+                adapter_info_str = str(adapter.request_adapter_info()).lower()
+                adapter._score = 0
+                for i, keyword in enumerate(gpu_keywords):
+                    if keyword in adapter_info_str:
+                        adapter._score += len(gpu_keywords) - i
+            # Prioritize. For equal scores, the original order is maintained.
+            adapters.sort(key=lambda adapter: -adapter._score)
```
kushalkolar, line 251: I just realized it's common to have compute nodes with ex. 4 of the same GPU. What do you think about also allowing `gpu_preference` to be an `int` which maps directly to the index of adapters returned from `enumerate_adapters()`, assuming that `enumerate_adapters` always returns GPUs in the same order? I can test this out tomorrow (Wednesday)!

almarklein: Oh dear, I had not considered that yet. I wonder what the `request_adapter_info()` dicts of these look like, and whether it's possible to discern them ... Would be great if you can post the result!

I find the idea of using an int a bit scary, because it will break down so easily. If you add a gpu, if one disconnects for some reason, if you run it on another machine, if wgpu-core lists more or less adapters, or decides to change the order ...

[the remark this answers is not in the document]
almarklein: Good question. One of the reasons why I'm not happy with the current approach.

almarklein:
I created #482 to dive a bit deeper into what info we can extract from the adapter_info.

toloudis:
> I actually feel a bit uncomfortable with the current state of this pr, because the gpu selection is very fuzzy and arguably arbitrary, while everything else in wgpu-py is very well-defined. I want to propose a slightly different approach:
> 
> * In wgpu-py, expose `gpu.enumerate_adapters()`, producing adapter instances.

Watching from afar, as I originated the enumerate_adapters PR in wgpu-native... I agree this is a good approach - just expose enumerate_adapters as a Python call and leave the selection to callers to do their own evaluation.  This is a great alternative to using the standard "webgpu selection rules" to auto-select.

almarklein:
Ok, I reduced this PR to only adding `gpu.enumerate_adapters()` and `adapter.summary`. This makes selecting an adapter much easier for downstream code. When we know more about the best way to select an adapter, we may add a utility.

Korijn (approved):

Issue #485, opened 2024-03-28
WxWgpuWindow object has no attribute 'add_event_handler'


cansik:
At least on MacOS, it seems that the `WxWgpuWindow` integration seems not to work. The [example found](https://pygfx.readthedocs.io/en/stable/_gallery/other/cube_wx.html) here can not be run because of the following issue:

```python
Traceback (most recent call last):
  File "/pyfx_image_viewer.py", line 11, in <module>
    canvas = WgpuCanvas()
  File "/venv/lib/python3.9/site-packages/wgpu/gui/wx.py", line 146, in __init__
    self._subwidget.add_event_handler(weakbind(self.handle_event), "*")
AttributeError: 'WxWgpuWindow' object has no attribute 'add_event_handler'
```

Versions:
- pygfx `0.1.18`
- wgpu `0.15.0`
- wxpython `4.2.1`
- python `3.9.18`

cansik:
@Korijn I have created a PR to fix this in wgpu and another one which updates the wx example in pygfx. Would be great if you could have a look.

cansik:
@Korijn Would it be possible to create a new wgpu release, since in `0.15.1` my changes of the merge request #486  are not present? This blocks using wgpu & pygfx together with wx.

almarklein:
#496

almarklein:
new release is available. Thanks for the heads-up!

almarklein:
Closed by #552

PR #486, opened 2024-03-28
WxWgpuWindow object has no attribute 'add_event_handler'

cansik:
As already mentioned in #485, this issue fixes the missing event handler method call, by removing it. It also forces the wait-time calculation to be `>=1` and converts the wait-time to an integer after conversion to ms. This PR also adds the `.idea` file to the .gitignore.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -121,3 +121,5 @@ venv.bak/
 # mypy
 .mypy_cache/
 
+# jetbrains idea
+.idea/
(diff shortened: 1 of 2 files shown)


almarklein (approved):
Looks good!

Yeah the wx backend is not a *full* backend, in the sense that it can be used in as an "auto-backend". It use is only to support wgou canvas in a wx app. I do get confused about this sometimes, and the fact that   `add_event_handler()` was used is a consequence of that, I suppose 😅

Korijn (approved):

PR #505, opened 2024-05-23
Support timestamp queryset for begin_render_pass()

almarklein:
Implementing this because I'm looking into benchmarking. See #450 for the initial implementation for compute passes.


almarklein:
This one is ready.

Korijn (approved):

Issue #383, opened 2023-10-19
PySide 6.6 asyncio support 


Korijn:
Great news from the PySide camp: https://www.qt.io/blog/qt-for-python-6.6

An official asyncio compatible Qt event loop!

See https://github.com/fork-tongue/observ/pull/93/files for an example of how to integrate.

The main benefit of supporting this in wgpu-py gui backends is that users will be able to use async/await syntax in the render function called by the canvas.

almarklein:
Nice!!!

Issue #355, opened 2023-03-21
Sync events with draws


almarklein:
@korijn mentioned in a comment in a pygfx, that  game engines usually run a tight loops with the following steps:

   1. Query input device state (joysticks, buttons, mice, etc)
   2. Update world state
   3. Render frame
 
That got me thinking. Right now, when a mouse event happens, this causes certain state to be changed. If two (or more) such events happen in between two draws, the state is updated twice, but we only see the final state.

We already do some event throttling to reduce this, but we could also aggregate the events and only dispatch them right before doing the draw, which could potentially result in more performance because it avoids calculating stuff that is never shown anyway.

One problem with this is that right now, if nothing changes, there are also no new draws, which is nice to avoid depleting your battery, and safes network traffic when using the notebook. We'd have to dispatch the events on a regular interval. I'm not sure what happens if we ask the GUI to schedule a draw and then not draw it.

Any other caveats?

Korijn:
Responding to your question:

In pygame, and I think Unity and other engines do this as well, events from the OS are stored in a queue-like buffer as they come in.

Then, the first thing that happens in the game loop, is that a copy of the buffer is retrieved and the original buffer is emptied. The events on the copy are processed, and then the next frame is drawn. Meanwhile new events coming in go into the fresh buffer.

You could perform event aggregation on the buffer copy before it is processed to reduce the amount of events.

---

Independently thought:

I think this tight loop pattern is something users should be able to implement on top of pygfx, since it is mostly useful for game-like applications which have animation, sound, physics and AI going on constantly as well. It's almost a guarantee that you have to process events and render new frames on a regular/fixed rate.

For plotting and viz type applications it's much less useful. You only want to draw when it is needed. So now, when is the right time to buffer and aggregate? I guess you could say that you target a "max fps" and therefore when a first event comes in, you can set a timer to decide when the buffer should be processed.

For unlimited fps, you could disable buffering I guess.

Korijn:
Also see https://github.com/pygfx/pygfx/issues/495

almarklein:
Implemented in rendercanvas.

PR #507, opened 2024-05-27
Try to optimize proxy_func

almarklein:
In benchmarking for https://github.com/pygfx/pygfx/issues/763 it looked like the `proxy_func` is called a lot, so I figured that of it could be optimized it would be an easy win. It looks like the result is not really measurable, but nevertheless might be worth applying.

* Use a `deque` instead of a list.
* Move the capturing logic to the `ErrorHandler` class, so that the logic in the `proxy_func` becomes simpler. Also cleaner code imo.
* The above also means that the proxy_func closure only captures the `error_handler` object and no longer uses `self`.

### DIFF
--- a/wgpu/backends/wgpu_native/_helpers.py
+++ b/wgpu/backends/wgpu_native/_helpers.py
@@ -3,6 +3,7 @@
 
 import sys
 import ctypes
+from queue import deque
 
 from ._ffi import ffi, lib
 from ..._diagnostics import DiagnosticsBase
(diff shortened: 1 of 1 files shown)


almarklein:
BTW: I tried to find other areas for improving performance. One is that our glfw GUI backend consumes a bit more CPU than Qt, because of the custom event-loop to keep the windows alive. This can be refactored to use asyncio to do more of the scheduling. But let's do this when we work on #355  / #383 / update modes.

almarklein:
BTW, a Cython-based wrapper for wgpu-native might provide obsevable performance improvement :)

Korijn (approved):

Korijn (approved):

PR #511, opened 2024-06-03
Update the default parameters of "set_bind_group()"

panxinmiao:
When setting bind groups, we often have code like this:
`render_pass.set_bind_group(0, bind_group, [], 0, 99)`

According to the WebGPU specification, this method has two overloaded different method signatures. 
The `dynamic_offsets` parameter has a default value of `[]`, and the last two parameters can also be omitted, indicate to direct passing of an array of dynamic offsets without copy.

The current usage is quite odd, as it requires users to provide parameters that are not actually used, which can cause confusion.

See: https://www.w3.org/TR/webgpu/#gpubindingcommandsmixin-setbindgroup

When setting a bind group, we can simplify the code to:
`render_pass.set_bind_group(0, bind_group)`

### DIFF
--- a/wgpu/_classes.py
+++ b/wgpu/_classes.py
@@ -1424,9 +1424,9 @@ class GPUBindingCommandsMixin:
         self,
         index,
         bind_group,
-        dynamic_offsets_data,
-        dynamic_offsets_data_start,
-        dynamic_offsets_data_length,
+        dynamic_offsets_data=[],
+        dynamic_offsets_data_start=None,
+        dynamic_offsets_data_length=None,
     ):
         """Associate the given bind group (i.e. group or resources) with the
         given slot/index.
--- a/wgpu/_classes.py
+++ b/wgpu/_classes.py
@@ -1437,9 +1437,9 @@ class GPUBindingCommandsMixin:
         Arguments:
             index (int): The slot to bind at.
             bind_group (GPUBindGroup): The bind group to bind.
-            dynamic_offsets_data (list of int): A list of offsets (one for each bind group).
-            dynamic_offsets_data_start (int): Not used.
-            dynamic_offsets_data_length (int): Not used.
+            dynamic_offsets_data (list of int): A list of offsets (one for each entry in bind group marked as buffer.has_dynamic_offset).Default [].
+            dynamic_offsets_data_start (int): Offset in elements into dynamic_offsets_data where the buffer offset data begins.Default None.
+            dynamic_offsets_data_length (int): Number of buffer offsets to read from dynamic_offsets_data.Default None.
         """
         raise NotImplementedError()
 
--- a/wgpu/_classes.py
+++ b/wgpu/_classes.py
@@ -1420,6 +1420,9 @@ class GPUBindingCommandsMixin:
     """Mixin for classes that defines bindings."""
 
     # IDL: undefined setBindGroup(GPUIndex32 index, GPUBindGroup? bindGroup, Uint32Array dynamicOffsetsData, GPUSize64 dynamicOffsetsDataStart, GPUSize32 dynamicOffsetsDataLength);
+    @apidiff.change(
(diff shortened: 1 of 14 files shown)


panxinmiao:
I am not very familiar with the codegen utility yet. I just followed the steps outlined in the guide and I am not sure if there were any mistakes. 😅

panxinmiao:
The validation of the `dynamic_offsets_data_start` and `dynamic_offsets_data_length` parameters in the code can cause compatibility issues with some downstream code. Previously, we did not use these two parameters, which led to the presence of code like `set_bind_group(0, bind_group, [], 0, 99)`.

However, if we strictly follow the specification, these parameters must be validated. I am not sure if this will have a significant impact. Perhaps we can continue not using the last two parameters, This is equivalent to only supporting another method signature with only three parameters in the WebGPU specification.

wgpu/_classes.py
```diff
@@ -1420,23 +1420,26 @@ class GPUBindingCommandsMixin:
...
 
         Arguments:
             index (int): The slot to bind at.
             bind_group (GPUBindGroup): The bind group to bind.
-            dynamic_offsets_data (list of int): A list of offsets (one for each bind group).
-            dynamic_offsets_data_start (int): Not used.
-            dynamic_offsets_data_length (int): Not used.
+            dynamic_offsets_data (list of int): A list of offsets (one for each entry in bind group marked as buffer.has_dynamic_offset).Default [].
+            dynamic_offsets_data_start (int): Offset in elements into dynamic_offsets_data where the buffer offset data begins.Default None.
+            dynamic_offsets_data_length (int): Number of buffer offsets to read from dynamic_offsets_data.Default None.
```
almarklein, line 1442: Small stylistic tweaks:
```suggestion
            dynamic_offsets_data (list of int): A list of offsets (one for each entry in bind group marked as ``buffer.has_dynamic_offset``). Default ``[]``.
            dynamic_offsets_data_start (int): Offset in elements into dynamic_offsets_data where the buffer offset data begins. Default None.
            dynamic_offsets_data_length (int): Number of buffer offsets to read from dynamic_offsets_data. Default None.
```

wgpu/backends/wgpu_native/_api.py
```diff
@@ -2050,10 +2050,28 @@ def set_bind_group(
...
-        dynamic_offsets_data_length,
+        dynamic_offsets_data=[],
+        dynamic_offsets_data_start=None,
+        dynamic_offsets_data_length=None,
     ):
+        if (
+            dynamic_offsets_data_start is not None
+            or dynamic_offsets_data_length is not None
+        ):
+            assert (
```
almarklein, line 2061: Can you please rewrite these assertions to use `if`  and `raise`?

almarklein: I think we also want to check that `dynamic_offsets_data_length >= 0`, because negative indices in Python work, but will produce unexpected results in this case.

almarklein (commented):
Nice! This indeed solves an ugly bit of the API :)  Just two small comments.

wgpu/backends/wgpu_native/_api.py
```diff
@@ -2050,10 +2050,28 @@ def set_bind_group(
...
+                dynamic_offsets_data_start is not None
+                and dynamic_offsets_data_start >= 0
+            )
+            assert (
+                dynamic_offsets_data_length is not None
+                and dynamic_offsets_data_start + dynamic_offsets_data_length
+                <= len(dynamic_offsets_data)
```
almarklein, line 2068: For backwards compatibility we can allow lengths that are too large. Since Python's slicing already behaves this way, we only have to remove the test for it.

wgpu/backends/wgpu_native/_api.py
panxinmiao: Updated.

Korijn (approved):

PR #515, opened 2024-06-05
Support for pinch events

almarklein:
Experimental PR to look into pinch events.

### Status

Seems it can be made to work for Qt. Likely also for jupyter. Not (for now) on glfw. Also quite some work to do in pygfx. Looked into it, and made this PR to save what I found.

### Tasks

* [x] Implement for Qt.
* [ ] Implement for wx?
* [ ] Implement for glfw. Looks like we can't. https://github.com/glfw/glfw/issues/90
  * This is sad, because it makes code that uses pinch less portable. 
* [ ] Implement for `jupyter_rfb`.
* [ ] In pygfx add support for pinch events.
* [ ] Support pinch events in pygfx controllers.

### DIFF
--- a/wgpu/gui/qt.py
+++ b/wgpu/gui/qt.py
@@ -27,6 +27,7 @@ if libname:
     try:
         WA_PaintOnScreen = QtCore.Qt.WidgetAttribute.WA_PaintOnScreen
         WA_DeleteOnClose = QtCore.Qt.WidgetAttribute.WA_DeleteOnClose
+        PinchGesture = QtCore.Qt.GestureType.PinchGesture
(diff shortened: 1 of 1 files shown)


Korijn:
In my experience, at least on the web, instead of working with these high level gesture events and abstractions, you just work with pointer events and use the pointer IDs to discern the fingers.

almarklein:
Yes, that moves some of the complexity to the application level, but it is more flexible. So an alternative approach would be to make Qt work for multi-touch events. The event spec already supports multi-touch events btw.

hmaarrfk:
You may be right. I’ll boot up on windows to check

PR #516, opened 2024-06-07
Integrate with Imgui

panxinmiao:
This PR integrates ImGui using the [imgui_bundle](https://github.com/pthom/imgui_bundle) library.
the imgui_bundle provides Python bindings for ImGui.

I implemented an ImGui backend based on `wgpu-py`, as well as an ImGui renderer using the `wgpu-py` Canvas.

The main challenge I faced in this work was the encapsulation of the API. Ultimately, I abstracted two classes: `ImguiWgpuBackend` and `ImguiRenderer`.

- `ImguiWgpuBackend` is a lower-level API that is independent of any GUI framework. It encapsulates the core logic for rendering ImGui using wgpu. Essentially, it wraps the GPU DrawCall required for rendering ImGui, allowing it to be embedded in any Render Pass, — You only need to insert the GPU DrawCall at the appropriate place in your application's rendering loop.  Note that UI events need to be handled yourself (via `backend.io`).

- `ImguiRenderer` uses the wgpu-py `Canvas` as its concrete implementation and interfaces with the wgpu-py event system. It draws ImGui using an independent render pass, rendering directly to the Canvas' surface texture. Therefore, we need to be able to retrieve the same surface texture multiple times in different render passes within the same frame.

For specific usage, please refer to the examples included in this PR.

Additionally, I am still considering the integration and API design with `pygfx`. It could be treated as a special world object in pygfx or deeply integrated with the pygfx renderer to leverage features such as anti-aliasing. 

Currently, in `pygfx`, you can also use the `ImguiRenderer` implementation to integrate ImGui into the pygfx scene rendering. This is entirely independent of the pygfx rendering logic, sharing only the `Canvas` instance. Therefore, any methods or features in `pygfx` will not affect ImGui. (such as `snapshot()` method in the `pygfx.renderer` )

Example:

https://github.com/pygfx/wgpu-py/assets/8044566/e5126e46-d3a7-49a4-aef2-fa709d0ef117


kushalkolar:
This looks really nice! :smile: . I'll play with it in `pygfx` over the next few days :)

Korijn:
I had never heard of imgui_bundle before. Based on stars it seems pyimgui is a bit more popular. Can you explain the difference between the two libraries and why you picked this one?

I tried to integrate with pyimgui before but I really struggled.

panxinmiao:
> Can you explain the difference between the two libraries and why you picked this one?
> 

When starting this work, I investigated the Python binding libraries listed in the `imgui` project. I found that the `imgui_bundle` project was relatively newer and more active. In contrast, `pyimgui` was last updated in April of last year. Additionally, `imgui_bundle` also includes bindings for many third-party extensions based on imgui, such as `ImPlot`, `ImGuizmo`, and other libraries. I thought that using `imgui_bundle` integration might help in obtaining a more powerful ecosystem integration. 

Overall, however, I didn't particularly carefully consider this choice, and don't have a particular preference. I just tried starting with `imgui_bundle`, and it worked out quite well. 😅

Korijn:
* pygfx currently requires python >= 3.8, but imgui-bundle latest version only supports python >= 3.10
* what is the minimum version of imgui-bundle you (want to) support with this code?
* imgui-bundle 1.0.0 is the last version with py 3.9 support
* imgui-bundle 0.6.6 is the last version with py 3.8 support

wgpu/backends/wgpu_native/_api.py
```diff
@@ -547,12 +547,13 @@ def get_current_texture(self):
...
         if self._texture:
-            self._drop_texture()
-            logger.warning(
-                "get_current_texture() is called multiple times before pesent()."
-            )
+            return self._texture
+            # self._drop_texture()
+            # logger.warning(
+            #     "get_current_texture() is called multiple times before pesent()."
+            # )
```
Korijn, line 556: @almarklein can you comment on this change? is it OK?

Korijn (dismissed):
Apologies for the delay. The changes look great to me. I'm OK with the location of the new modules in `wgpu.utils`.

wgpu/backends/wgpu_native/_api.py
almarklein: My motivation for the earlier behavior is to force a user to use pairs of `get_current_texture()` and `present()`, and that errors in user-code would be detected this way.

But I think @panxinmiao has a good case that just returning the existing texture provides more flexibility (for cases where another subsystem must render something), which would otherwise be cumbersome due to passing the current texture around ...

But let's remove the old commented code.

wgpu/backends/wgpu_native/_api.py
```diff
@@ -547,12 +547,13 @@ def get_current_texture(self):
         # * return the existing texture
         # * warn about it, and create a new one
         # * raise an error
-        # Right now we do the warning, so things still (kinda) keep working
+        # Right now we return the existing texture, so user can retrieve it in different render passes withing the same frame.
```
almarklein, line 550:
```suggestion
        # Right now we return the existing texture, so user can retrieve it in different render passes that write to the same frame.
```

wgpu/utils/imgui_backend.py
```diff
@@ -0,0 +1,510 @@
...
+
+            global_vtx_offset += commands.vtx_buffer.size()
+            global_idx_offset += commands.idx_buffer.size()
+
+    def _invalidate_device_objects(self):
+        self._render_pipeline._destroy()
+        self._uniform_buffer._destroy()
+        self._font_texture._destroy()
+        self._font_texture_view._destroy()
+        self._font_texture_sampler._destroy()
```
almarklein, line 508: Using private methods here. These were also changed recently. Can `.destroy()` textures and buffers. The rest can simply be set to None, I think, to let the gc do its thing.

wgpu/gui/qt.py
```diff
@@ -244,12 +247,27 @@ def _key_event(self, event_type, event):
...
         self._handle_event_and_flush(ev)
 
+    def _char_input_event(self, char, char_str):
+        ev = {
+            "event_type": "char",
+            "char": char,
+            "char_str": char_str,
+            "modifiers": None,
+        }
+        self._handle_event_and_flush(ev)
```
almarklein, line 257: A new event is introduced here. For the sake of portability, we should add it to jupyter_rfb (and its spec) too. I'm not sure if an event like this exists in JS. The `input` even comes close.

How badly is this new needed? Can we omit it for now, so we have time to investigate proper compat?

Or we we make it a private event for the time being ...

wgpu/gui/qt.py
```diff
@@ -244,12 +247,27 @@ def _key_event(self, event_type, event):
         }
         self._handle_event_and_flush(ev)
 
+    def _char_input_event(self, char, char_str):
+        ev = {
+            "event_type": "char",
+            "char": char,
+            "char_str": char_str,
```
almarklein, line 254: Only `char_str` is used, and `char` is not always set, so cannot be relied on. So let's use just one field.

almarklein (commented):
Finally took the time to do a proper review. I made a few comments.

The new event requires some thought.

How portable is this code for using other imgui libs (alternative Python wrappers)? Just to get an impression to know how easy it would be to adjust of the ecosystem changes.

I think it makes sense to put the imgui stuff in a `utils.imgui` subpackage:
```py
from wgpu.utils.imgui import ImguiRenderer
from wgpu.utils.imgui import ImguiWgpuBackend
```

wgpu/gui/qt.py
panxinmiao: > A new event is introduced here. For the sake of portability, we should add it to jupyter_rfb (and its spec) too. I'm not sure if an event like this exists in JS. The `input` even comes close.
> 
> How badly is this new needed? Can we omit it for now, so we have time to investigate proper compat?
> 
> Or we we make it a private event for the time being ...

It primarily handles character input events, which are somewhat different from key events. It is mainly used for UI controls that accept character input. Most importantly, it allows obtaining results from the input method, which is useful for inputting non-English characters.

Like this:
image

wgpu/utils/imgui_backend.py
panxinmiao: I recently tested and discovered that after the update, actively calling the destroy method on an existing texture while recreating the font texture in the `create_fonts_texture` method causes a Rust panic (texture already destroyed). It seems the old font texture cannot be immediately destroyed here. Therefore, I removed the explicit destroy call and will let the garbage collector handle it instead.

panxinmiao:
> I think it makes sense to put the imgui stuff in a utils.imgui subpackage:

Updated.

> How portable is this code for using other imgui libs (alternative Python wrappers)? Just to get an impression to know how easy it would be to adjust of the ecosystem changes.
> 

I haven't rigorously tested it, but I believe integration should be quite easy. Since we've implemented a generic backend for imgui, these imgui-based extension libraries should naturally receive support. 

I tried `implot`, and it works well.

image

wgpu/utils/imgui_backend.py
almarklein: To provide a bit of context: the previous destroy methods in wgpu-py actually called `xxRelease`, which basically reduces the ref-counter for that object, i.e. Rust knows that we no longer use it. The new `destroy()` methods in wgpu-py call the actual destroy functions in wgpu-native, so the low-level object is marked as not-usable, but this can indeed lead to errors when the object is still in use internally.

wgpu/gui/glfw.py
```diff
@@ -502,6 +503,14 @@ def _on_key(self, window, key, scancode, action, mods):
         }
         self._handle_event_and_flush(ev)
 
+    def _on_char(self, window, char):
```
almarklein, line 506:
```suggestion
    def _on_char(self, window, char):
        # Undocumented char event to make imgui work, see https://github.com/pygfx/wgpu-py/issues/530
```

wgpu/gui/qt.py
almarklein: Ok, let's go with what we have.

almarklein:
This is looking great! Awesome work!

The only hurdle is that this relies on the new `char` event, which means it does not yet work via jupyter_rfb, sorry @kushalkolar. I created #530 to track this, as it requires some more thought and work. I don't think we need to hold this PR for that. We can merge with desktop-only support, and add full support later.

almarklein (approved):

Korijn (approved):

kushalkolar:
> The only hurdle is that this relies on the new `char` event, which means it does not yet work via jupyter_rfb, sorry @kushalkolar. I created #530 to track this, as it requires some more thought and work. I don't think we need to hold this PR for that. We can merge with desktop-only support, and add full support later.

yea been playing with it in jupyter_rfb, and yes of course one step at a time, we can do this fully later :smile:

PR #517, opened 2024-06-08
Add args to utils.get_default_device()

Vipitis:
### Motivation
The `wgpu.utils.get_default_device()` function is really convenient, especially for the test suite. However it only request the device without any features (or limits). So in cases where a specific feature is always required, it would be nice to use this same function - with features passed to the `request_device()` method. I have a usecase for this in `wgpu-shadertoy` now.

### Issues
Since this sets a global device, it will actually skip if you subsequently request a device with more features. One case being https://github.com/pygfx/wgpu-py/blob/ed9c451fe8eb2f1210df632cae8b7ec1199f982d/examples/compute_timestamps.py#L44-L47 which will work on it's own, but fail during the example tests. 
A possible solution, is likely to discard the current global device, if it is is missing the required features and request a new one.

### DIFF
--- a/examples/cube.py
+++ b/examples/cube.py
@@ -22,8 +22,7 @@ for a in wgpu.gpu.enumerate_adapters():
 canvas = WgpuCanvas(title="wgpu cube", size=(640, 480))
 
 # Create a wgpu device
-adapter = wgpu.gpu.request_adapter(power_preference="high-performance")
-device = adapter.request_device()
+device = wgpu.utils.get_default_device()
 
 # Prepare present context
 present_context = canvas.get_context()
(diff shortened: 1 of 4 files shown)


almarklein:
The `get_default_device()`  was added in part to avoid the problem that when devices were requested repeatedly, the process would fail because of a memory leak. Typical cases are unit tests. IIRC this issue has since be resolved upstream.

To be honest, I don't like the idea to add args to the function, because it's not default anymore in that case :P, and you run into problems like the one you already described.

What exactly is your use-case? Is it specific to testing? In that case I'd just implement your own version of `get_default_devce()` that has the feature you need it to have.

Vipitis:
alright. I can think of a different solution already. 

I was using default device so far, but now needed to `"float32-filterable"` feature and couldn't find a way to dynamically enable that. So my first idea was to request that feature with the default device.

almarklein:
Pygfx requires the same feature. It has its own mechanic to create a global device. I think it makes sense to handle this lib/application specific. I would not be surprised if we deprecated wgpu-py's `get_default_device()`  at some point.

PR #518, opened 2024-06-10
Poll waits for queue to be empty

almarklein:
I encountered a use-case where one call to poll was not enough.


Korijn:
This has the potential to deadlock. I would recommend a timeout or max poll count argument.

almarklein:
My bad, I had changed the `wait` arg from False to True while  working on benchmarks. That's why it failed.
```py

libf.wgpuDevicePoll(self._internal, True, ffi.NULL)  
libf.wgpuDevicePoll(self._internal, False, ffi.NULL)
# ____________________________________/\

PR #519, opened 2024-06-10
Implement destroy

almarklein:
* [x] Rename the `_destroy()`  methods to `_release()`, becauase that's what they do.
* [x] Make the public `destroy()` methods do the actual destroy call.
     * Only device, queryset, buffer, and texture have these.
     * Only actually does something for buffer and texture. But when wgpu-native implements it for device, we will do so automatically.
    * See https://github.com/gfx-rs/wgpu-native/blob/33133da4ec5a0174cb21539ef2d3346f75200411/src/lib.rs#L2491-L2493  
* [x] Improve docsrings for `destroy()`.
* [x] Tests.

### DIFF
--- a/wgpu/_classes.py
+++ b/wgpu/_classes.py
@@ -256,9 +256,9 @@ class GPUCanvasContext:
 
     def __del__(self):
         self._ot.decrease(self.__class__.__name__)
-        self._destroy()
+        self._release()
 
-    def _destroy(self):
+    def _release(self):
         pass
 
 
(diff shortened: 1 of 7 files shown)


almarklein:
I looked into this because with the benchmarks that I'm working on, the results depend a lot on the order of the tests, which is very annoying. I was looking for a reset button.

wgpu/_classes.py
```diff
@@ -1274,9 +1283,11 @@ def create_view(
...
     # IDL: undefined destroy();
     def destroy(self):
-        """An application that no longer requires a texture can choose
-        to destroy it. Note that this is automatically called when the
-        Python object is cleaned up by the garbadge collector.
+        """Destroy the texture.
+
+        Explicitly destroys the texture, freeing its memory and putting
+        the object in an unusable state. In general its easier (and
+        safer) to just let the garbadhe collector do its thing.
```
Korijn, line 1290: Misspelled garbage in a few places

almarklein: fixed

Korijn (approved):

Issue #526, opened 2024-06-18
wgpu-py requires unneeded arguments.


fyellin:
This is a hodge-podge of small bugs, all basically that wgpu-py is more verbose or less expressive than the corresponding JS and Rust code.

1. In wgpu-py, the `timestamp_writes` argument to `encoder.begin_render_pass` and `encoder.begin_compute_pass` requires that you pass both `beginning_of_pass_write_index` and an `end_of_pass_write_index` argument, and that they both be integers.  The actual API allows either of these arguments to be elided.  

The Rust API data structure [`ComputePassTimestampWrites`](https://docs.rs/wgpu/latest/wgpu/struct.ComputePassTimestampWrites.html) has both fields as `Option<u32>`.  

It is impossible to indicate in wgpu-py that I only want one timestamp written.  (In my case, I was doing 20 compute passes, and I only wanted the start of the first pass and the end of the last pass.)

2. To create a bind group requires me to write:
```
    render_bind_group = device.create_bind_group(
        layout=render_pipeline.get_bind_group_layout(0),
        entries=[
            {'binding': 0, 'resource': {'buffer': random_buffer, 'offset': 0, 'size': random_array.nbytes}},
        ]
    )
```
In JS and Rust, I don't need to specify either the `offset` or `size`.  It's a bit painful that for each buffer, I have to keep track of its origins.  [`BufferBinding`](https://docs.rs/wgpu/latest/wgpu/struct.BufferBinding.html)'s third value is an `Option<>`.

3. wgpu-py only has the 5-argument version of set_bind_group.   JS provides 2-, 3-, and 5-argument versions of the function. 99% of the code I've seen uses the 2-argument version.

Adding insult, wgpu-py's implementation of `setBindGroup` completely ignores the last two arguments altogether.

4.  Everywhere that we are dealing with the size of a texture, wgpu-py requires specifying all three of `height`, `width`, and `depth_or_array_layers`. The field [GPUExtent3DDict](https://www.w3.org/TR/webgpu/#typedefdef-gpuextent3d) defaults the latter two fields to 1. 

================

I am happy to fix 3 and 4.  For 1 and 2, I'd need some advise on how to tell  `new_struct_p` and `new_struct` that a field is optional and its value is `None`.

fyellin:
I see that #3 has already been fixed in the recent version.

almarklein:
Good point. When we started wgpu-py, there were less optional arguments in WebGPU, and we've not kept up :)

[...]

PR #528, opened 2024-06-19
Don't call _tuple_from_tuple_or_dict directly, but got thrown an intermediate function

fyellin:
Don't call _tuple_from_tuple_or_dict, but go through an intermediate function for each possible type that knows the names of the arguments and what their defaults are.  No need to duplicate code. This fixes default arguments for Extend3d, Blend, and Origin3D.  Color (which requires all four components) is handled, too.  

Update _tuple_from_tuple_or_dict so that it can handle a list of default results.

Add lots of tests.

Fix one typo that is user visible.

Partial fix for #526.

### DIFF
--- a/tests/test_wgpu_native_basics.py
+++ b/tests/test_wgpu_native_basics.py
@@ -78,15 +78,7 @@ def test_tuple_from_tuple_or_dict():
     assert func({}, ("width", "height"), (10, 20)) == (10, 20)
 
     # Test that with dictionaries, you can elide values at the beginning, if we have them
-    assert func(
-        {"z": 5},
-        ("x", "y", "z"),
-        (
-            1,
-            2,
-            3,
-        ),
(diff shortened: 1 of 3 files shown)


almarklein (dismissed):
Thanks! This code looks great, and I'm very happy to see such extensive unit tests 🙏 

I think you need to run `black .` and it should be good (perhaps update black?)

fyellin:
Sigh.  Black was turning a couple of lines into an ugly mess.  I was able to find a compromise that made both me and the reformatter happy.

Korijn (dismissed):

almarklein (approved):
Thanks!

fyellin:
Hmm.  I fixed the codegen test pass, but that caused the approval to become stale.  Please re-approve.

Korijn (approved):

PR #529, opened 2024-06-20
Deal better with missing size and missing query index arguments

fyellin:
Second (and final part) of bug fix for #526 

Modify `create_bind_group` to handle missing `offset` and `size.
Modify `timestamp=` argument in `create_xxxx_pipeline` so that either timestamp index can be missing.

More controversially (and I'll delete this if you want me to):

Modify `set_vertex_buffer`, `set_index_buffer`, and `clear_buffer` to mark the size as "not given by user" rather than making a guess as to what the size actually is.  

This doesn't fix any bugs, but seemed to be the right thing to do. In the normal cases, the current code works correctly. But if we ever end up with arrays of vertexes or the like, the current code won't work.

Added tests: 
Deleted `offset` and `size` from both compute and render pipelines.  
Added half-indexed timestamp arguments to make sure they gave correct results.
Tested all sorts of iterations of clear_buffer.

### DIFF
--- a/wgpu/backends/wgpu_native/_api.py
+++ b/wgpu/backends/wgpu_native/_api.py
@@ -37,6 +37,8 @@ from ._helpers import (
     SafeLibCalls,
 )
 
+WGPU_WHOLE_SIZE = 0xFFFF_FFFF_FFFF_FFFF  # 2**64 - 1
+WGPU_QUERY_SET_INDEX_UNDEFINED = 0xFFFF_FFFF
 
 logger = logging.getLogger("wgpu")  # noqa
 
@@ -1263,8 +1265,8 @@ class GPUDevice(classes.GPUDevice, GPUObjectBase):
                     "WGPUBindGroupEntry",
                     binding=int(entry["binding"]),
                     buffer=resource["buffer"]._internal,
-                    offset=resource["offset"],
-                    size=resource["size"],
+                    offset=resource.get("offset", 0),
+                    size=resource.get("size", WGPU_WHOLE_SIZE),
                     sampler=ffi.NULL,
                     textureView=ffi.NULL,
                     # not used: nextInChain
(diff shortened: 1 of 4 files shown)


fyellin:
The failure seems to be setting up QEMU before it ever looks at my code:

 ```
Pulling binfmt Docker image
  /usr/bin/docker pull tonistiigi/binfmt:latest
  Error response from daemon: Head "https://registry-1.docker.io/v2/tonistiigi/binfmt/manifests/latest": received unexpected HTTP status: 503 Service Unavailable
```
I'll try restarting the tests later.

File: wgpu/backends/wgpu_native/_api.py [hunk 1]
```diff
@@ -37,6 +37,8 @@
     SafeLibCalls,
 )
 
+WGPU_WHOLE_SIZE = 0xFFFF_FFFF_FFFF_FFFF  # 2**64 - 1
+WGPU_QUERY_SET_INDEX_UNDEFINED = 0xFFFF_FFFF
```
almarklein, line 41: These should be taken from the `lib.WGPU_WHOLE_SIZE`. That way we don't have to worry about these values changing (it has happened).

fyellin: I do not see anywhere else in _api.py or _classes.py where a `#define`d constant is pulled from webgpu.h.  Is there documentation somewhere on how to do so, or at least to ensure that the values are the same in both places? 

Aside from this, are you happy with the push? Should I keep the changes to `set_vertex_buffer`, `set_index_buffer` and `clear_buffer`?  Tests will stay, no matter what.

almarklein: If you search for `lib.` in the source, you'll see some. Could be that its only enum values so far, but `#define`'d constants are there too. The header file is parsed by ffi, producing the `lib` object.

I am happy about the other changes 👍

fyellin: I initially misunderstood what you were asking me to do.  I've made the easy fix rather than the hard fix.

Korijn (approved):

almarklein:
Thanks for this contribution! And sorry for the misunderstanding about these constants, I could have been more clear.

fyellin:
No problem.

PR #531, opened 2024-06-21
Verify constants in _api.py

fyellin:
This push is intended to precede #529 and address the concerns that @almarklein had about the push.

Add code to codegen that looks for constants starting with WGPU_ in _api.py and confirms that their value matches that given in C definition files.


fyellin:
Let me know if you want me to merge this and #529 into a single push.  It seemed like they dealt with separate issues.

Also I added all the constants to _api.py, even though I only need a few.  It seems that throughout _api.py, there are commented out references to the other constants.

almarklein:
Thanks for the careful work here! However, `cffi` already parses these constants and puts them in the `lib` object. So I don't think its necessary to redo that work ourselves.

fyellin:
I'm not sure I understand what you're saying. 

In `_api.py`, I need to set specific fields of structs to `WGPU_QUERY_SET_INDEX_UNDEFINED` and `WGPU_WHOLE_SIZE`.  These fields can't be left blank, and these fields have to be set to these specific values by Python code. I cannot find these constants available to Python anywhere else.  

As I'm looking through `_api.py`, I also see comments of the form "WGPU_COPY_STRIDE_UNDEFINED ?" and "WGPU_ARRAY_LAYOUT_COUNT_UNDEFINED ?" which makes me believe that most of the other constants would be useful to future developers.

I tried to decide whether I was better off creating these constants in one of the generated files, or just verifying them in `_api.py`.  They didn't fit in really well with either any of the currently generated files, and I didn't feel confident enough to create a new generated file from scratch.  So I just verify what's in `_api.py` against `webgpu.h` and complain if they don't match.

If I'm missing your point, and there really is some way that Python code has access to these constants, let me know.  I'm not finding them anywhere outside of C header files.

Korijn:
You can just write lib.WGPU_COPY_STRIDE_UNDEFINED and cffi will be aware of this constant because it has parsed exactly the header file you mention.

fyellin:
I'll close this push request and update the other one appropriately.

Issue #532, opened 2024-06-25
Recent pull request has a bug


fyellin:
I made a mistake in a recent pull request.   I will fix it this afternoon.  (California time).

In `_tuple_from_blend_component`, I accidentally put the fields in the wrong order.  I discovered this when I wrote actual code that elided these fields.

I will fix the bug and make sure it is adequately tested.

fyellin:
I may want the advise of one of the more experienced developers in knowing the correct fix for this.  I see where my confusion came from.

In the [specification](https://www.w3.org/TR/webgpu/#dom-gpublendstate-color), the fields are in the order `operation`, `srcFactor`, `dstFactor`.

In wgpu-py, they are used in the order `srcFactor`, `dstFactor`, `operation`.  See lines 1583-1953 in _api.py and line 269-271 of test_gui_glfw.py.

The specification also says nothing about allowing a tuple instead of a dictionary.  This seems to be a wgpu-py innovation.  In the other cases where we allow a tuple instead of a dictionary, this is explicitly allowed by the spec.

So should I be using (1) the order as given in the spec? (2) the order that wgpu-py seems to have settled on? or (3) only allow [a possibly empty] dictionary here and not allow a tuple or list at all.

almarklein:
Thanks for the quick response to the found bug. And thanks for taking the time to compare our implementation with the spec. I think its important to follow the spec rather precisely, otherwise we may drift away from it over time.

The fact that the spec does not allow a tuple is a blessing; it means we don't have to choose between 1 and 2. So my answer would be to go with 3. It may break people's code who use tuples, but over time its worth it.

fyellin:
Will do.  I haven't been around long enough to make backwards-incompatible
changes without authorization.

Korijn:
This can closed now right @fyellin ?

almarklein:
The PR says `Fix bug #532` but GH did not pick that up.

PR #533, opened 2024-06-25
Fix bugs.  Defaults were listed in the order given by GPUBlendCompone…

fyellin:
Fix bug #532.  I accidentally listed defaults in the order given by the spec in GPUBlendComponent rather than in the order they were actually being used.

Fix tests for this code.

In one test, have it actually elide the blend components, so that we can see that the default values are doing the right thing.

## API changes

* xx can no longer be a tuple; use a dict instead. (AK: @fyellin could you fill this in?)

### DIFF
--- a/wgpu/backends/wgpu_native/_api.py
+++ b/wgpu/backends/wgpu_native/_api.py
@@ -172,15 +172,6 @@ def _tuple_from_extent3d(size):
     )
 
 
-def _tuple_from_blend_component(components):
-    return _tuple_from_tuple_or_dict(
-        # defaults to "add", "one", "zero"
-        components,
-        ("src_factor", "dst_factor", "operation"),
-        ("one", "zero", "add"),
-    )
-
-
 def _tuple_from_origin3d(destination):
     fields = destination.get("origin", (0, 0, 0))
     # Each field individually is 0 if not specified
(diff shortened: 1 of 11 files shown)


fyellin:
Note that I mention several possible fixes for #532.  I'm assuming that backwards compatibility is more important than precisely following the spec.  If a decision is made otherwise, I will update this.

fyellin:
Removed backwards compatibility. Code requires a dictionary for "blend".  Fixed all places in the tests and documentation that were not compatible.

almarklein (approved):

wgpu/backends/wgpu_native/_api.py
```diff
@@ -1133,7 +1124,7 @@ def create_bind_group_layout(
                 # H: nextInChain: WGPUChainedStruct *, type: WGPUBufferBindingType, hasDynamicOffset: WGPUBool/int, minBindingSize: int
                 buffer = new_struct(
                     "WGPUBufferBindingLayout",
-                    type=info["type"],
+                    type=info.get("type", "uniform"),
```
almarklein, line 1127: Looks like some changes from #534 snuck in.

Korijn (approved):

PR #534, opened 2024-06-27
More defaults

fyellin:
My final removal of unnecessary arguments.  Note that for each case, I've also removed the field from at least one test so that we can verify that the code is working correctly.

In the four [`BindGroupLayoutEntry`](https://www.w3.org/TR/webgpu/#dictdef-gpubindgrouplayoutentry)s that wgpu supports, three of them can can take an empty dictionary as its argument: 

1. In a [`BufferBindingLayout`]( https://www.w3.org/TR/webgpu/#dictdef-gpubufferbindinglayout), all three fields are optional, and so the dictionary can be the empty dictionary.  The "type" field defaults to "uniform".  

 2. In a [`SamplerBindingLayout`](https://www.w3.org/TR/webgpu/#dictdef-gpusamplerbindinglayout) the single field `type` defaults to `"filtering"`.
 
 3. In a [`TextureBindingLayout`](https://www.w3.org/TR/webgpu/#dictdef-gputexturebindinglayout) all fields can have default values.  We did not recognize `"texture": {}` as valid.
 
 4.  A [`StorageTextureBindingLayout`](https://www.w3.org/TR/webgpu/#dictdef-gpustoragetexturebindinglayout) has required fields.  But the access defaults to `write-only`.

In a [`VertexState`](https://www.w3.org/TR/webgpu/#dictdef-gpuvertexstate), the list of buffers defaults to the empty list.

### DIFF
--- a/examples/triangle.py
+++ b/examples/triangle.py
@@ -90,12 +90,6 @@ def setup_draw(context, device):
             "module": shader,
             "entry_point": "vs_main",
         },
-        # primitive elided, since we are using its default value and want to test that.
-        # primitive={
-        #     "topology": wgpu.PrimitiveTopology.triangle_list,
-        #     "front_face": wgpu.FrontFace.ccw,
-        #     "cull_mode": wgpu.CullMode.none,
-        # },
         depth_stencil=None,
         multisample=None,
         fragment={
--- a/examples/triangle.py
+++ b/examples/triangle.py
@@ -89,13 +89,13 @@ def setup_draw(context, device):
         vertex={
             "module": shader,
             "entry_point": "vs_main",
-            "buffers": [],
         },
-        primitive={
-            "topology": wgpu.PrimitiveTopology.triangle_list,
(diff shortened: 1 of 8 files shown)


examples/triangle.py
```diff
@@ -89,13 +89,13 @@ def setup_draw(context, device):
...
             "module": shader,
             "entry_point": "vs_main",
-            "buffers": [],
-        },
-        primitive={
-            "topology": wgpu.PrimitiveTopology.triangle_list,
-            "front_face": wgpu.FrontFace.ccw,
-            "cull_mode": wgpu.CullMode.none,
         },
+        # primitive elided, since we are using its default value and want to test that.
```
Korijn, line 93: I understand what you're trying to say and do here, but I don't think this should be merged into the examples. Just remove the commented code.

Korijn (approved):
@almarklein I'll let you merge if you're OK with this

PR #539, opened 2024-07-11
Imgui multi canvas

kushalkolar:
Exposes the context of `ImguiRenderer` so that it can be used for context switching. As far as I understand, this is required for rendering different UIs to multiple different canvases (added an example). Not sure if there's a better way to do this, welcome to suggestions! @panxinmiao 

tested in glfw & qt

### DIFF
--- a/examples/imgui_multi_canvas.py
+++ b/examples/imgui_multi_canvas.py
@@ -1,0 +1,99 @@
+"""
+Example showing how to use multiple imgui contexts to draw to multiple canvases
+
+# run_example = false
+"""
+
+import wgpu
+from imgui_bundle import imgui
+from wgpu.gui.auto import WgpuCanvas, run
+from wgpu.utils.imgui import ImguiRenderer
+
+# Create a canvas to render to
+canvas = WgpuCanvas(title="imgui", size=(512, 256))
+canvas2 = WgpuCanvas(title="imgui", size=(512, 256))
+canvas3 = WgpuCanvas(title="imgui", size=(512, 256))
+
+# Create a wgpu device
+adapter = wgpu.gpu.request_adapter(power_preference="high-performance")
+device = adapter.request_device()
+
+# create a imgui renderer for each canvas
+imgui_renderer = ImguiRenderer(device, canvas)
+imgui_renderer2 = ImguiRenderer(device, canvas2)
+imgui_renderer3 = ImguiRenderer(device, canvas3)
+
+
+# Separate GUIs that are drawn to each canvas
+def update_gui():
+    imgui.new_frame()
+
+    imgui.set_next_window_size((300, 0), imgui.Cond_.appearing)
+    imgui.set_next_window_pos((0, 20), imgui.Cond_.appearing)
+
+    imgui.begin("window1", None)
+    imgui.button("b1")
+
+
+    imgui.end()
+
+    imgui.end_frame()
+    imgui.render()
+
+    return imgui.get_draw_data()
+
+
+def update_gui2():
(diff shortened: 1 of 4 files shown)


panxinmiao:
I didn't consider the situation of multiple` ImguiRenderer` instances. 😅 

LGTM.

examples/imgui_multi_canvas.py
```diff
@@ -0,0 +1,98 @@
...
+    imgui_renderer3.render(update_gui3())
+
+    # done! request draw
+    canvas.request_draw()
+    canvas2.request_draw()
+    canvas3.request_draw()
+
+
+if __name__ == "__main__":
+    canvas.request_draw(draw_frame)
```
kushalkolar, line 97: this is maybe a bit weird, I think this should be separated into 3 different `draw_frame()` functions for each canvas?

kushalkolar:
> I didn't consider the situation of multiple` ImguiRenderer` instances. 😅

No worries, your work on implementing imgui integration has been amazing! :smile:

examples/imgui_multi_canvas.py
almarklein: Yes, I think that would be the correct way. Although there's one event-loop so the current implementation should also be fine.

examples/imgui_multi_canvas.py
```diff
@@ -0,0 +1,98 @@
...
+    imgui.end_frame()
+    imgui.render()
+
+    return imgui.get_draw_data()
+
+
+def draw_frame():
+    # set corresponding imgui context before rendering
+    imgui.set_current_context(imgui_renderer.imgui_context)
+    imgui_renderer.render(update_gui())
```
Korijn, line 82: Why did you decide not to call set current context within the render function? Seems like if it's always necessary you could just do it there?

panxinmiao: It's not the `render` function that needs it; the `render` method only requires `draw_data`. However, to obtain `draw_data`, we need to switch to the correct context before updating the GUI code. This means that setting the context should be done in the `update_gui()` method. The user needs to build or update the GUI in the correct imgui context environment.

If we want to handle this within the `render` method, we need to modify the API by adding a property to the `ImguiRenderer` as a hook function for updating the GUI. Then, in the `render` function, we would do something like this:

```python
imgui.set_current_context(...)
draw_data = self.update_gui_callback()  # Call the hook method
# Do render with draw_data
```

kushalkolar: > If we want to handle this within the `render` method, we need to modify the API by adding a property to the `ImguiRenderer` as a hook function for updating the GUI. Then, in the `render` function, we would do something like this:
> 
> ```python
> imgui.set_current_context(...)
> draw_data = self.update_gui_callback()  # Call the hook method
> # Do render with draw_data
> ```

I think I like this idea! I updated the PR to implement this. Users do not have to be aware of context switching.

almarklein: The other imgui examples will need to be updated to use the new API, right?

kushalkolar: Yup, I wanted your thoughts on the changes before modifying the other examples.

kushalkolar: I went ahead and updated the other examples,if anyone (especially @panxinmiao ) has more suggestions I can incorporate them :)

panxinmiao: Thanks for the updates! 
LGTM :)

panxinmiao (approved):

Korijn (approved):

Issue #537, opened 2024-07-02
Keyerror when trying to run triangle.py


Mats-Janssens:
```raceback (most recent call last):
  File "/home/mats/CLONES/wgpu-py/examples/triangle.py", line 146, in <module>
    main(canvas)
  File "/home/mats/CLONES/wgpu-py/examples/triangle.py", line 68, in main
    return _main(canvas, device)
  File "/home/mats/CLONES/wgpu-py/examples/triangle.py", line 137, in _main
    draw_frame = setup_draw(canvas.get_context(), device)
  File "/home/mats/CLONES/wgpu-py/examples/triangle.py", line 88, in setup_draw
    render_pipeline = device.create_render_pipeline(
  File "/home/mats/VENV/sim/lib/python3.10/site-packages/wgpu/backends/wgpu_native/_api.py", line 1427, in create_render_pipeline
    for buffer_des in vertex["buffers"]:
KeyError: 'buffers'
Segmentation fault (core dumped)
```

I have just pip install wgpu and glfw on a fresh linux mint install where the proper nvidia drivers are installed. I dont think I need to install any glfw or wgpu official binaries because they should be included in the respective wheels so I dont understand what is going wrong. Thank you very much in advance for your help on this issue.

Korijn:
* Looking at your stack trace, it seems like you have cloned the repo, so you are probably running the example code from the latest development commit on the main branch
* You also mention you have installed wgpu from pip, so that's the last release, not the latest development commit

See the problem?

You have two options to resolve (pick one):

1. Checkout the commit from the repo that matches the version you installed with pip (`git checkout v0.16.0`)
2. Uninstall the version you installed with pip (`pip uninstall wgpu`) and setup a development environment so you can run the latest development commit code, instructions can be found here: https://github.com/pygfx/wgpu-py?tab=readme-ov-file#developers

almarklein:
What Korijn said. The `buffers` field is no longer required in the latest version. The example has been updated to demonstrate this, but this means that the latest version of the example no longer works on the older version of wgpu-py :)

PR #540, opened 2024-07-14
Fixed triangle example

lmarzocchetti:
I have copy pasted the triangle example to try the library and it doesn't work. So after some debugging i find that the "create_render_pipeline" methods from GPUDevice, in the vertex attribute want a "buffers" empty list, and in the fragment the alpha is just not required (the library crash if there is something).

### DIFF
--- a/examples/triangle.py
+++ b/examples/triangle.py
@@ -89,6 +89,7 @@ def setup_draw(context, device):
         vertex={
             "module": shader,
             "entry_point": "vs_main",
+            "buffers": [],
         },
         depth_stencil=None,
         multisample=None,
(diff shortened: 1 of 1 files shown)


almarklein:
Ah, this is due to a recent change related to default arguments. The example was updated with these changes. So it does work with the latest `main`, but not with the latest release. 

Best to checkout the examples on the tag of the latest release: https://github.com/pygfx/wgpu-py/tree/v0.16.0/examples

We should at some point add a gallery in the docs that shows the example code matching the released version.

Korijn:
You can read more about your mistake here: https://github.com/pygfx/wgpu-py/issues/537#issuecomment-2202270028

almarklein:
It's a common mistake :)

PR #541, opened 2024-07-14
cmap picker with imgui

kushalkolar:
I'm doing a bunch of imgui stuff now which has finally forced me to learn how to use wgpu directly :smile: . I can keep adding more examples if you're interested. There are certain imgui things which are backend-specific, for instance there are many examples out there on using textures with opengl but not with wgpu-py so it took me a while to figure out how to implement something like this.

https://github.com/user-attachments/assets/be28e2a0-72a9-4755-945d-f1cb48329c81

draft because I want to clean it up a bit and add comments

### DIFF
--- a/examples/imgui_cmap_picker.py
+++ b/examples/imgui_cmap_picker.py
@@ -1,5 +1,12 @@
 """
-Example showing how to use multiple imgui contexts to draw to multiple canvases
+Imgui example that shows how to create a colormap picker menu
+
+Uses the cmap library: https://github.com/tlambert03/cmap
+
+pip install cmap
+
+Instead of using the cmap library you can create 2D arrays in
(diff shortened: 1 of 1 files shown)


panxinmiao:
I think we can add an `add_texture()` method in the `ImguiWgpuBackend` class.

kushalkolar:
> I think we can add an `add_texture()` method in the `ImguiWgpuBackend` class.

I was thinking that too! I want to get more familiar with imgui & wgpu before making a PR on that, I want to see what else in imgui might need things like this.

kushalkolar:
added a border and comments, should be good to go :) 

image

almarklein (approved):

Korijn (approved):

Issue #459, opened 2024-01-29
Run examples on CI on Windows / Mac / Linux


almarklein:
I think on CI we now only run stuff on Linux with LavaPipe. But it should now also work on MacOS, see https://github.com/gfx-rs/wgpu-native/pull/364. Windows should also work, but IIRC not all our tests pass for DX12 yet. 

Would be good to:
* Run tests of wgpu-py on all possible platforms.
* Have insight into what platform is fastest (since they're emulated GPU's). 
* Use the faster variant for e.g. pygfx's validation examples and docs.

PR #544, opened 2024-07-29
Implement occlusion query

fyellin:
Implementation turned out to be straightforward.

### DIFF
--- a/tests/test_wgpu_occlusion_query.py
+++ b/tests/test_wgpu_occlusion_query.py
@@ -19,9 +19,13 @@ elif is_ci and sys.platform == "win32":
 
 default_shader_source = """
 
+// Draws a square with side 0.1 centered at the indicated location.
+// If reverse, we take the vertices clockwise rather than counterclockwise so that
+// we can test culling.
+
 struct Uniform {
-    offset: vec3f,
-    scale: vec2f,
+    center: vec3f,
+    reverse: u32,  // Actually a bool
 }
 
 @group(0) @binding(0) var<uniform> uniform : Uniform;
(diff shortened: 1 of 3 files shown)


tests/test_wgpu_occlusion_query.py
```diff
@@ -0,0 +1,172 @@
...
+
+import wgpu
+from pytest import skip
+from testutils import run_tests, get_default_device
+from testutils import can_use_wgpu_lib, is_ci
+from wgpu import flags
+
+if not can_use_wgpu_lib:
+    skip("Skipping tests that need the wgpu lib", allow_module_level=True)
+elif is_ci and sys.platform == "win32":
```
Korijn, line 16: Is it possible to make this conditional on DirectX being used?

fyellin:
I don't have access to a Windows machine, so I'm not exactly sure what this
is supposed to be and I have no way to test it.

I copied these lines directly from `test_wgpu_native_render.py`.

tests/test_wgpu_occlusion_query.py
almarklein: At some point we should get back to these and see if they work now on CI. Also see #459

almarklein (approved):

fyellin:
@almarklein .  You need to approve again.  Apparently because there was one unresolved conversation at the time you approved, it didn't count.

Korijn:
Thank you! We greatly appreciate your contributions.

fyellin:
By the way, I've also got RenderBundles working, though I haven't yet
written any significant tests.  (I have rewritten some vanilla code to use
them, so I know the basic idea is sound.)

Is there some bulletin or forum I should use to announce this? How do I
find out information like "Which Python versions do we support"?

almarklein:
> By the way, I've also got RenderBundles working,

Nice! 🚀 

> Is there some bulletin or forum I should use to announce this?

There's the release notes where new features are mentioned.

There's also https://github.com/pygfx/wgpu-py/discussions but it is not used much, to be honest. Maybe it'd be nice to have a monthly newsletter of sorts 🤔

PR #546, opened 2024-07-31
Implement Render Bundle and Tests

fyellin:
- Implement Render Bundle.
- Remove "return" statements in the `GpuDebugCommandsMixin` code.  
-- The functions are implemented in `libf`, so might as well call them. I have determined if they do anything or not. But we're prepared when they finally are implemented.
- Add tests.  
-- All rendering tests in `test_wgpu_native_render` are run both with direct rendering and render bundles.
-- Appropriate memory tests added.

@almarklein 
@Korijn 

I was uncertain about the best ways to handle the polymorphism in `GPUDebugCommandsMixin`, `GPUBinidngsCommandMixin` and `GPURenderCommandsMixin`.  In each case, we run identical code no matter what the type of `self`, except for calling a different library function.

I implemented this three different ways, in the hopes that you'd let me know which you like best.  I'll use that implementation everywhere.

- In  `GPUDebugCommandsMixin`, I do the simplest solution. I call, for example,  `type(self).push_debug_group_function(..args..)`. Each of the three classes that extends `GPUDebugCommandsMixin` has a class variable defining `push_debug_group_function`. While simple, this means that the call to `libf.XXXXPushDebugGroup` is separated from the appearance of `libf.XXXXPushDebugGroup` in the code, so we will not necessarily see its codegen'ed annotations or notice if they've changed. 

- In the single function of `GPUBindingCommandsMixin`, I do the same, but I also add a comment indicating what are the possible libf functions that I might be calling.  While redundant, it does mean that I see the annotations.

- In the functions of `GPURenderCommandsMixin`, I use `isinstance` hidden in a helper function.  The libf functions that I'm calling are right there for me to see.

- I didn't try using a `match` statement.  This might provide a cleaner syntax than using `isinstance`, but requires Python 3.10.  I didn't know what the minimum Python requirement is for this project.  [Amended.  I see that ci.yml tests back to 3.8, so I'm assuming that is the current minimum.]

### DIFF
--- a/wgpu/backends/wgpu_native/_api.py
+++ b/wgpu/backends/wgpu_native/_api.py
@@ -1680,7 +1680,30 @@ class GPUDevice(classes.GPUDevice, GPUObjectBase):
         depth_read_only: bool = False,
         stencil_read_only: bool = False,
     ):
-        raise NotImplementedError()
+
+        c_color_formats, color_formats_count = ffi.NULL, 0
+        if color_formats:
+            color_formats_list = [enummap["TextureFormat." + x] for x in color_formats]
+            c_color_formats = ffi.new("WGPUTextureFormat []", color_formats_list)
+            color_formats_count = len(color_formats_list)
+
+        # H: nextInChain: WGPUChainedStruct *, label: char *, colorFormatCount: int, colorFormats: WGPUTextureFormat *, depthStencilFormat: WGPUTextureFormat, sampleCount: int, depthReadOnly: WGPUBool/int, stencilReadOnly: WGPUBool/int
+        render_bundle_encoder_descriptor = new_struct_p(
+            "WGPURenderBundleEncoderDescriptor *",
+            label=to_c_label(label),
+            colorFormatCount=color_formats_count,
+            colorFormats=c_color_formats,
+            depthStencilFormat=depth_stencil_format or 0,
+            sampleCount=sample_count,
+            depthReadOnly=depth_read_only,
+            stencilReadOnly=stencil_read_only,
+            # not used: nextInChain
+        )
+        # H: WGPURenderBundleEncoder f(WGPUDevice device, WGPURenderBundleEncoderDescriptor const * descriptor)
+        render_bundle_id = libf.wgpuDeviceCreateRenderBundleEncoder(
+            self._internal, render_bundle_encoder_descriptor
+        )
+        return GPURenderBundleEncoder(label, render_bundle_id, self)
         # Note: also enable the coresponing memtest when implementing this!
 
     def create_query_set(self, *, label="", type: "enums.QueryType", count: int):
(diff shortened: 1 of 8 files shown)


wgpu/backends/wgpu_native/_api.py
```diff
@@ -2276,15 +2295,35 @@ def draw_indexed(
...
     classes.GPUCommandEncoder, GPUCommandsMixin, GPUDebugCommandsMixin, GPUObjectBase
 ):
+
+    # GPUDebugCommandsMixin
+    # H: void f(WGPUCommandEncoder commandEncoder, char const * groupLabel)
+    push_debug_group_function = libf.wgpuCommandEncoderPushDebugGroup
+    # H: void f(WGPUCommandEncoder commandEncoder)
+    pop_debug_group_function = libf.wgpuCommandEncoderPopDebugGroup
+    # H: void f(WGPUCommandEncoder commandEncoder, char const * markerLabel)
+    insert_debug_marker_function = libf.wgpuCommandEncoderInsertDebugMarker
```
almarklein, line 2325: I'd prefer if these were prefixed with an underscore, otherwise they look like public API.

wgpu/backends/wgpu_native/_api.py
```diff
@@ -2160,101 +2183,92 @@ def set_bind_group(
...
-            # H: void f(WGPUComputePassEncoder computePassEncoder, char const * groupLabel)
-            libf.wgpuComputePassEncoderPushDebugGroup(
-                self._internal, c_group_label, color
-            )
-        else:
-            # H: void f(WGPURenderPassEncoder renderPassEncoder, char const * groupLabel)
-            libf.wgpuRenderPassEncoderPushDebugGroup(
-                self._internal, c_group_label, color
-            )
+        function = type(self).push_debug_group_function
```
almarklein, line 2202: For the record, getting the type should not be necessary, this should work as well.
```py
        function = self.push_debug_group_function
```

Though it could be that your version is faster, because it won't have to check `self.__dict__` first, so we can leave it as-is.

wgpu/backends/wgpu_native/_api.py
```diff
@@ -2276,15 +2295,35 @@ def draw_indexed(
...
-        libf.wgpuRenderPassEncoderDrawIndexedIndirect(
-            self._internal, buffer_id, int(indirect_offset)
-        )
+        function = self.__pick_render_function(
+            # H: void f(WGPURenderPassEncoder renderPassEncoder, WGPUBuffer indirectBuffer, uint64_t indirectOffset)
+            render_pass=libf.wgpuRenderPassEncoderDrawIndexedIndirect,
+            # H: void f(WGPURenderBundleEncoder renderBundleEncoder, WGPUBuffer indirectBuffer, uint64_t indirectOffset)
+            render_bundle=libf.wgpuRenderBundleEncoderDrawIndexedIndirect,
+        )
+        function(self._internal, buffer_id, int(indirect_offset))
```
almarklein, line 2304: You're now using two different approaches to select the correct function. Can we use the set-function-as-class-attribute approach here as well? I like that one better, TBH

almarklein (commented):
Nice work!

wgpu/backends/wgpu_native/_api.py
almarklein: Sorry, I jumped right into the code and read your post later.

Good point about the annotations. I think its well worth having these near the code that does the call. So please go with the `__pick_render_function` approach 👍

wgpu/backends/wgpu_native/_api.py
almarklein: Moot comment once we move to `self.__pick_render_function()`.

wgpu/backends/wgpu_native/_api.py
almarklein: Moot comment once we move to `self.__pick_render_function()`.

wgpu/backends/wgpu_native/_api.py
fyellin: I'm glad to see that I'm not the only one who was surprised that I couldn't write this.  I did originally write `self.push_debug_group_function`, but this doesn't work!.  

Here's a simpler code snippet.

```
def global_printer(a):
    print("Printer:", a)

class MyClass:
    printer = global_printer

    def print_it(self, object):
        function = self.printer
        function(object)

MyClass().print_it("Hello")
```
which when run gives the strange error message.
```
TypeError: global_printer() takes 1 positional argument but 2 were given
```
When Python sees that that value of  `<instance>`.`<field>` is an unbound function, it binds `<instance>` to the first argument.

I have to write either `self.__class__` or `type(self).` to force Python to access the field without binding the first argument.  The latter looked a little bit nicer.

wgpu/backends/wgpu_native/_api.py
fyellin: If I'm reading between the lines correctly, you think the `type(self)._xxx_yyy_function` implementation is cleaner, but you don't like that the annotations are missing when we do that.  I think I've come up with a good hybrid approach that gives me both.  I'll be updating this pull in a few hours.  

If after seeing the hybrid approach, you still prefer `self.__pick_render_function`, etc, then I'll change it back.

fyellin: If I'm reading between the lines correctly, you think the `type(self)._xxx_yyy_function` implementation is cleaner, but you don't like that the annotations are missing when we do that.  I think I've come up with a good hybrid approach that gives me both.  I'll be updating this pull in a few hours.  

If after seeing the hybrid approach, you still prefer `self.__pick_render_function`, etc, then I'll change it back.

fyellin:
Update coming soon.

wgpu/backends/wgpu_native/_api.py
Korijn: The latter also has better portability guarantees

fyellin:
I think I've now got the best of both worlds.  If you disagree, say so, and I'll revert back to the `instanceof` check.

When I see, for example:
```
     _draw_indexed_function = libf.wgpuRenderBundleDrawIndexed
```
I don't write out the annotation (it just gets in the way), but I keep trace of the fact that `_draw_indexed_function` was assigned this value.  Later when `type(self)._draw_indexed_function` is actually called, I dump out all the saved annotations at the call.

fyellin:
This push request is ready for review.  

1) If you want, I can also do the same for _release(self).  The same implementation of _release() is duplicated for every subclass of ObjectBase, and this could be moved to ObjectBase.  (One implementation has a little bit of extra stuff.)

fyellin:
Any update on this? I'm waiting to hear if you're happy with this implementation, which uses codegen to put comments in the right place. 

Or do you want me to use `isinstance` instead?

almarklein:
I'm on holiday right now with very little time (and very bad wifi), will review when I'm back.

fyellin:
> I'm on holiday right now with very little time (and very bad wifi), will review when I'm back.

I'm leaving on vacation August 26 and will have spotty internet for three week.  I hope we manage to overlap.  If not, this is certainly not urgent.

fyellin:
@almarklein.  Waiting for review.  At the least, let me know if you like this implementation which uses codegen to put in comments, or if you'd prefer me to go back to using instanceof.

almarklein (approved):
Excellent work! I like what you did with the methods of the abstract classes, solving the issue with annotations in the codegen. I have no further comments.

Korijn (approved):

PR #547, opened 2024-08-04
update to wgpu-native 22.1

Vipitis:
I am really excited for better error handling, ~~compilation info~~(not yet implemented in wgpu-native, I might try to do that myself to move it along) and glsl const built-ins, so I already started this. 
Feel free to cherry pick my changes or commit into this branch if it helps.

I got a lot of time to work on this, but need to read up on error handling a bit.

(from the issue template:)
## Preparations

* [x] Perhaps warp up outstanding PR's.
* [x] Might be a good time to roll a release of wgpu-py.
* [x] Perhaps for pygfx too.

## Upstream
pending gfx-rs/wgpu-native#402
* [x] wgpu-native is updated to a recent wgpu-core.
* [x] wgpu-native uses a recent `webgpu.h`.
* [x] wgpu-native has a release with these changes.

## Update to latest IDL

*Optional, can also only update wgpu-native.*

*The lines below can be copied in the PR's top post.*

* [x] Download the latest `webgpu.idl` and place in the resources folder.
* [x] Run `python codegen` to apply the automatic patches to the code.
* [?] It may be necessary to tweak the `idlparser.py` to adjust to new formatting.
* [x] Check the diff of `flags.py`, `enums.py`, `structs.py` for any changes that might need manual work.
 * [x] Go through all FIXME comments that were added in `_classes.py`:
        * Apply any necessary changes.
        * Remove the FIXME comment if no further action is needed, or turn into a TODO for later.
        * Note that all new classes/methods/properties (instead those marked as hidden) need a docstring.
* [x] Run `python codegen` again to validate that all is well. Repeat the step above if necessary.
* [x] Make sure that the tests run and provide full coverage.
* [x] Make sure that the examples all work.
* [x] Make sure that pygfx works (again).
* [x] Update release notes for API changes.

## Update to latest wgpu-native

*Optional, can also only update IDL*

*The lines below can be copied in the PR's top post.*

* [x] Run `python download-wgpu-native.py --version xx` to download the latest webgpu.h and DLL.
* [x] Run `python codegen` to apply the automatic patches to the code.
* [?] It may be necessary to tweak the `hparser.py` to adjust to new formatting.
* [x] Diff the report for new differences to take into account.
* [x] Diff `wgpu_native/_api.py` to get an idea of what structs and functions have changed.
* [x] Go through all FIXME comments that were added in `_api.py`:
        * Apply any necessary changes.
        * Remove the FIXME comment if no further action is needed, or turn into a TODO for later.
* [x] Run `python codegen` again to validate that all is well. Repeat the steps above if necessary.
* [x] Make sure that the tests run and provide full coverage.
* [x] Make sure that the examples all work.
* [x] Make sure that pygfx works (again).
* [x] Update release notes for changes in behavior.

## Wrapping up

* [x] Update pygfx.
* [ ] Release new wgpu-py.
* [ ] Release new pygfx.
* [ ] This can be a good moment to deal with certain issues, or perhaps some issues can be closed.


wgpu/backends/wgpu_native/_api.py
```diff
@@ -499,7 +508,13 @@ def configure(
             str_val = enum_int2str["CompositeAlphaMode"][int_val]
             capable_alpha_modes.append(str_val.lower())
 
-        # H: void f(WGPUSurfaceCapabilities capabilities)
+        # The format is used as-is
+        if format is None and capabilities.formatCount > 0:
+            format = enum_int2str["TextureFormat"][capabilities.formats[0]]
+        else:
+            raise RuntimeError("No formats supported by the surface.") # doesn't really make sense, maybe should be higher up.
```
Vipitis, line 515: I am getting capabilities with Null pointers for formats and preset modes when I use Vulkan backend(default). 

tried today with D3D12 and OpenGL and it populates those fields. Don't think this anything in my control right now, so I will use the other backends with `WGPU_BACKEND_TYPE`

Vipitis:
so seems like D3D12 and OpenGL cause issue with the tests in `test_wgpu_native_compute_tex.py` now - always returning all zeros. these do pass with vulkan (but render tests fail). Might need to try this on different hardware next week.

examples do pass on all backends (using offscreen via pytest)

remaining test errors are related to the specific Error formatting, so I will try to find out what changes are needed

Vipitis:
So I got the error tests passing - but only when run on their own... not when doing the whole suite. Reminds me of having issues with gc last time but don't think that is the case here. So maybe it's an issue with how pytest does caching for default device (had trouble with that in `wgpu-shadertoy`) - or the changes to error handling need some further tweaks.

image

might look into enabling some new features next.

Vipitis:
> so seems like D3D12 and OpenGL cause issue with the tests in `test_wgpu_native_compute_tex.py` now - always returning all zeros. these do pass with vulkan (but render tests fail). Might need to try this on different hardware next week.

have now tested this on my laptop with an Nvidia GPU and getting the same result:
Vulkan returns null pointers for `getSurfaceCapabilities`, D3D12 fails compute tests with all zeros.

almarklein:
I'm on holiday right now with very little time (and very bad wifi), will review/help when I'm back.

fyellin:
I downloaded this code and tried to use it on my MacOS.  I'm interested in looking at `override` and `push_constant`, both of which appear to be implemented (partly??) in the newest version.

When I would run `python download-wgpu-native.py` I'd get the error message that the downloaded file was not a zip file.  Turns out the "zip file" contained "Not Found".

It seems that `https://github.com/gfx-rs/wgpu-native/releases/download/v22.1.0/wgpu-macos-aarch64-release.zip` does not exist.

1. Is there something I need to do to get that file to exist?
2 Is there any way we can a slightly more informative error message if the URL doesn't in fact exist?

Vipitis:
> 1. Is there something I need to do to get that file to exist?

Wgpu-native hasn't yet released. The commit hash points to the PR branch, not a release tag yet. So you need to clone the repo(and it's submodules), build it yourself (need to install a working rust tool chain and others), and then replace the lib file (`.dylib` on OS X) or use `WGPU_LIB_PATH` env var to point to the new file.
The header files are already part of this PR branch.

Korijn:
To expand on that, you can track upstream progress here: https://github.com/gfx-rs/wgpu-native/pull/402

fyellin:
I read the title of this and assumed 22.1 was further along than it actually is.  I will wait.

I am also anxious to see some of the features here and I hope some of the bugs have been fixed

Vipitis:
> I'm on holiday right now with very little time (and very bad wifi), will review/help when I'm back.

Now I am on holiday too with absolutely no reception.Earliest I can continue working on this on is Friday.

Korijn:
> To expand on that, you can track upstream progress here: [gfx-rs/wgpu-native#402](https://github.com/gfx-rs/wgpu-native/pull/402)

Upstream release happened 10 hrs ago!

Vipitis:
so CI differs from my systems... But the fail case seems to be a device lost not raising the appropriate exception class. I am not sure if device lost is the intended error there in the first place. But there needs to be some adjustments to raise that properly. Will work on that tomorrow and Saturday. 

pytest showing different behavior when running tests isolated versus the whole suite is a common thread, but makes it difficult to debug, as it could be caching or gc.

wgpu/backends/wgpu_native/_api.py
```diff
@@ -827,14 +850,40 @@ def _request_device(
         def device_lost_callback(c_reason, c_message, userdata):
             reason = enum_int2str["DeviceLostReason"].get(c_reason, "Unknown")
             message = ffi.string(c_message).decode(errors="ignore")
-            error_handler.log_error(f"The WGPU device was lost ({reason}):\n{message}")
+            msg = f"The WGPU device was lost ({reason}):\n{message}"
+            # This is afaik an error that cannot usually be attributed to a specific call,
+            # so we cannot raise it as an error. We log it instead.
```
Vipitis, line 855: I am often encountering something like 

```bash
thread '<unnamed>' panicked at src\lib.rs:582:5:
Error in wgpuQueueSubmit: Validation Error

Caused by:
  Parent device is lost

```
where the python process crashes and is lost. As the message is logged and seen by stderr(I think), it would be great to handle this right here.
The message says Validation error, and for some CI tests for me locally is equivalent to the actually handled error.

Could this be an issue here the device lost callback is reached before the uncaptured error callback (which does handle the error and riseses and exception?).

almarklein: > I am often 

How often, and under what conditions?

Vipitis: let me get up and reduce some of these fatal errors into minimal repros. Maybe adding them as a `test_fatal_errors.py`? I collected quite a few but it's in a messy private repo.

almarklein: On second thought, reading this from the WebGPU docs:
image

I think the device getting lost is not supposed to result in an exception. It means the device is used in a bad way. The `device.lost` promise can be used to do something application-specific (but we have not implemented that). 

The fact that trying to use the device anyway, should be considered a bug in wgpu-core; it should result in an error (Python exception) instead.

Vipitis: My earlier change to handle this as validation error was to address CI/pytest which had `buffer_init3` and `native_errors` throw a "did not raise exception" failure. But still log a message that looked like the one expected. 

See for example here https://github.com/pygfx/wgpu-py/actions/runs/10704356989/job/29677159712?pr=547
The root cause is odd to me however. Since these tests fail as expected (with `GPUValidationError` raise) when run in isolation or some subset... Couldn't figure out what combination of tests caused this - apart from the full suite

wgpu/_classes.py
```diff
@@ -380,16 +382,16 @@ def is_fallback_adapter(self):
...
         return self._adapter_info
 
     @apidiff.add("Useful in multi-gpu environments")
     @property
     def summary(self):
-        """A one-line summary of the info of this adapter (name, adapter_type, backend_type)."""
+        """A one-line summary of the info of this adapter (description, adapter_type, backend_type)."""
         d = self._adapter_info
-        return f"{d['device']} ({d['adapter_type']}) via {d['backend_type']}"
+        return f"{d['description']} ({d['adapter_type']}) via {d['backend_type']}"
```
almarklein, line 394: This was needed because of a mistake in wgpu-native. This was fixed in https://github.com/gfx-rs/wgpu-native/pull/418. We should change it back to `device` when we update wgpu-native :) I will either do that in this pr or directly after in a new pr.

wgpu/backends/wgpu_native/_api.py
Vipitis: I pulled your changes and the test suite now passes locally on Vulkan. (only the texture formats issue still fails)

almarklein:
Did the null pointers in the capabilities also happen in previous versions, or is this new? It sounds like a bug in the driver or wgpu-core, you'd expect at least one value of each ... so maybe when nothing is reported, we skip the check and hope for the best ...

Vipitis:
> Did the null pointers in the capabilities also happen in previous versions, or is this new? It sounds like a bug in the driver or wgpu-core, you'd expect at least one value of each ... so maybe when nothing is reported, we skip the check and hope for the best ...

previous release(0.16.0 from pypi) works just fine. I also tried on my laptop which has iGPU and dGPU. Vulkan has this issue on 22.1 with either adapter. D3D12 works, so it can't really be some change that needs to be made.
And the examples in wgpu-native also work as expected with either backend option.
So the lead I am looking at right now is that it's either the surface (where nothing has changed?) or the adapter...

It's not just the formats that are empty, also the present modes.

Vulkan
image

D3D12
image

And hardcoding them isn't enough. My current thinking is here 
https://github.com/Vipitis/wgpu-py/blob/ec9317171be7da776ec8c6ceef73ec030afc844b/wgpu/backends/wgpu_native/_api.py#L235C9-L238C68
since the adapter wants the surface, but we just give it null in all the examples...

Vipitis:
with the hardcoding or your changes, this will give you an 
"Surface does not support the adapter's queue family" in `wgpuSurfaceConfigure`
and the following advice https://github.com/gfx-rs/wgpu/issues/1344#issuecomment-823448056

note that the D3D12 adapter also gets null pointer during configure and historically so did Vulkan. It just seems the case that now it no longer works...

almarklein:
I booted a Windows machine and I can reproduce the problem 🚀

rajveermalviya:
Is it possible it's hitting the error here? — https://github.com/gfx-rs/wgpu-native/blob/5bd6da3851be5a214c746cac691cd372e50f7333/src/lib.rs#L3836-L3838
Here are the default values — https://docs.rs/wgpu-types/22.0.0/src/wgpu_types/lib.rs.html#5407

almarklein:
Hi @rajveermalviya, thanks for chiming in!

The situation seems related to us having been very lazy about passing `surface_if` into `WGPURequestAdapterOptions`, simply setting it to NULL. This e.g. allows creating an adapter before a GUI backend is selected.

This has worked well until now. But with wgpu-native v22.1.0.2, this produces null pointers in the surface capabilities struct (i.e. zero supported formats and present modes), and when we just force them get "Surface does not support the adapter's queue family".

I just tried to play nice, and pass the canvas/surface, but now it selects DX12 instead of Vulkan, and produces warnings about downlevel flags .... 🤔

almarklein:
Since the resulting struct has zero formats, not even bgra8unorm_srgb and fifo, it looks like we don't get the default, but something less than that.

rajveermalviya:
> Since the resulting struct has zero formats, not even bgra8unorm_srgb and fifo, it looks like we don't get the default, but something less than that.

Looks like it is the [default](https://docs.rs/wgpu-types/22.0.0/src/wgpu_types/lib.rs.html#5407) value to me.

fyellin:
Just so you aren't caught by surprise.  I had to completely rewrite the handling of limits to get push_constants to work.  I needed some wgpu-specific limits, and there are some extra shenanigans involved in fetching them and setting them.  Also as I mention in my comments on that pull-request, I think it's incorrect that features_use_hyphens while limits_use_underscore.

Essentially warning you that you may have a merge mess coming up.

almarklein:
> Looks like it is the [default](https://docs.rs/wgpu-types/22.0.0/src/wgpu_types/lib.rs.html#5407) value to me. 🙂

Ah, you're right. I was looking at the code below, which mentions certain guarantees, which I assumed to be part of the default.

almarklein:
It looks like wgpu has become unworkable on Windows (at least the machine I have at hand):

* Not specifying a surface: Panics with "Surface does not support the adapter's queue family".
  * This worked fine before, what has changed that causes this to fail now? 
* Specifying the surface id: Selects D3D12 and gives a big warning about downlevel flags.
  * If D3D12 worked, this was ok, but that warning does not sound good. Is this because D3D12 is not yet fully supported, or because D3D12 cannot provide the full wgpu functionality?
* Forcing Vulkan via `WGPURequestAdapterOptions.backendType` and specify the surface id: Errors with "No suitable adapters found".
  * Why is Vulkan suddenly not supported anymore?

This is on an Intel machine with Integrated graphics and Windows 11.

Vipitis:
The warnings also show in wgpu-native when I force the backend to D3D12 there. But otherwise it does work. Apart from compute shaders, which I don't think are meant to work on D3D12. 

Vulkan works in wgpu-native, so it should work in wgpu-py too. Offscreen canvas works, but glfw and qt did not. 

Were you able to try Vulkan on a non Windows OS?

almarklein:
> Vulkan works in wgpu-native, so it should work in wgpu-py too. Offscreen canvas works, but glfw and qt did not.

Oh, interesting ... I will try wgpu-native when I'm back at the office (where I have a Windows machine) tomorrow. What happens on your machine is you modify `cube.py` using:
```py
adapter = wgpu.gpu.request_adapter(power_preference="high-performance", canvas=canvas)
```

> Were you able to try Vulkan on a non Windows OS?

Yeah, worked fine on Linux.

Vipitis:
> Oh, interesting ... I will try wgpu-native when I'm back at the office (where I have a Windows machine) tomorrow. What happens on your machine is you modify `cube.py` using:
> 
> ```python
> adapter = wgpu.gpu.request_adapter(power_preference="high-performance", canvas=canvas)
> ```

exactly the same as forcing D3D12, warnings but works. If I force Vulkan here I do get the "No suitable adapter found" from `wgpuInstanceRequestAdapter` the status returned by the callback is "Unavailable". Which is spawned from a `NotFound` in `wgpu-core`

also tried OpenGL backend which works.

I poked around in the blame and couldn't find anything obvious. A few PRs in that region talked about backends in instance. e.g: https://github.com/gfx-rs/wgpu/pull/5535

almarklein:
found it! It was related to how the surface id was obtained. Somehow the one we creates was wrong. Patch underway.

almarklein:
@Vipitis can you confirm it works for you too now? If it does, I think this is about ready. There are no real API changes, but I'm going to have a go with pygfx to make sure no changes are needed to e.g. shaders.

Vipitis:
I will try this on both machines when I am home in ~90 minutes

Vipitis:
checked on both machines, seems to be good now. examples work with gui, pytest passes. Also tried with forced Vulkan and D3D12.

I will collect the fatal errors and raise a separate issue/PR to figure out if they can be handled and a panic avoided

Korijn (approved):

almarklein:
🥳

almarklein:
Thanks @Vipitis!

PR #549, opened 2024-08-06
Fix automated tests

### LINKED ISSUE #548: Test suite appears to be broken on github
Test suite is failing on git hub. Even "main", which used to pass, is failing.

Multiple tests are failing with:

```
      ImportError: cannot import name 'get_platform' from 'wheel.bdist_wheel' (/tmp/pip-build-env-o1du4o99/overlay/lib/python3.9/site-packages/wheel/bdist_wheel.py)
      [end of output]
```
The github runner image was updated on 30 July.

fyellin:
FIXES #548.

Changed import of `get_platform` to be from `sysconfig`

My one concern is that the original code had `get_platform(None)`, and the new `get_platform` doesn't take an argument.  There is a cryptic comment `force a platform tag`, and I'm not sure if that is referring to the whole statement, or the `None` argument that I had to elide.  Can't find any documentation.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -2,8 +2,8 @@ import re
 import platform
 
 from setuptools import find_packages, setup
-from wheel.bdist_wheel import get_platform, bdist_wheel as _bdist_wheel
-
+from wheel.bdist_wheel import bdist_wheel as _bdist_wheel
+from sysconfig import get_platform
 
 NAME = "wgpu"
 SUMMARY = "Next generation GPU API for Python"
@@ -14,7 +14,7 @@ with open(f"{NAME}/__init__.py") as fh:
 
 class bdist_wheel(_bdist_wheel):  # noqa: N801
     def finalize_options(self):
-        self.plat_name = get_platform(None)  # force a platform tag
+        self.plat_name = get_platform()  # force a platform tag
         _bdist_wheel.finalize_options(self)
 
 
--- a/setup.py
+++ b/setup.py
@@ -3,7 +3,7 @@ import platform
 
 from setuptools import find_packages, setup
 from wheel.bdist_wheel import bdist_wheel as _bdist_wheel
-from sysconfig import get_platform
+from wheel.bdist_wheel.bdist_wheel import get_platform
 
 NAME = "wgpu"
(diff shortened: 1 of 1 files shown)


Korijn:
By default building a wheel for wgpu-py results in a platform agnostic wheel. This is because we don't compile anything. The rust binaries embedded in the wheel are precompiled and are considered data files as such.

However those binaries are of course not platform agnostic. So this implementation of `finalize_options` forces a platform tag to be set for the built wheel files. This allows us to build a wheel for every platform. That's what the comment means.

It's kind of a hack for a shortcoming in setuptools. There's no clean public API for this scenario.

Korijn:
Looking at the build logs, you seem to have done this correctly:

```
1 wheels produced in 1 minutes:
  wgpu-0.16.0-py3-none-manylinux_2_28_x86_64.whl   3,487 kB
```

Korijn:
I do think you should call the function from setuptools instead of sysconfig though, see the implementation here: https://github.com/pypa/setuptools/blob/46fe69938e2b88fcad78d2dfbcfc701a4cfa6b3f/setuptools/command/bdist_wheel.py#L67

fyellin:
Changed the get_platform to the one you recommended.  Making sure that tests still pass.

fyellin:
@Korijn 
I'm not exactly sure what it is you want me to change.  The implementation of get_platform you pointed me to seems not to exist anymore.

I'm travelling for the next three days, so I'm not going to be able to do any more on this bug. I'm closing this pull request so that someone else can handle this.

Korijn:
As the error message you shared in #548 indicated, the function we import has moved to setuptools.

`Deprecated the bdist_wheel module, as the code was migrated to setuptools itself`

So we need to update setup.py to import `get_platform` from setuptools, that's all.

fyellin:
I just tried `from setuptools import get_platform`

      ImportError: cannot import name 'get_platform' from 'setuptools' (/tmp/pip-build-env-a5bs2y1b/overlay/lib/python3.11/site-packages/setuptools/__init__.py)

fyellin:
At this point, someone else is going to need to fix this.

PR #551, opened 2024-08-08
Get different get_platform

fyellin:
FIXES #549.
Needed to switch to a different 1-argument get_platform()

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -2,8 +2,8 @@ import re
 import platform
 
 from setuptools import find_packages, setup
-from wheel.bdist_wheel import get_platform, bdist_wheel as _bdist_wheel
-
+from wheel.bdist_wheel import bdist_wheel as _bdist_wheel
+from wheel._bdist_wheel import get_platform
 
 NAME = "wgpu"
 SUMMARY = "Next generation GPU API for Python"
--- a/setup.py
+++ b/setup.py
@@ -2,8 +2,8 @@ import re
 import platform
 
 from setuptools import find_packages, setup
+from setuptools.command.bdist_wheel import get_platform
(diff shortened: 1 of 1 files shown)


cansik:
I liked your approach with `sysconfig` more in https://github.com/pygfx/wgpu-py/pull/549 since it is a public method. Now, `_bdist_wheel` is a private namespace.

@Korijn What is the reason to not use `sysconfig` since it creates the correct tags (if used like this `get_platform().replace("-", "_")`)?

Korijn:
If you look at the implementation of bdist_wheel, it has some logic on top of sysconfig's method, for a good reason. Sysconfig's method was not designed with the goal of creating a wheel platform tag.

Replacing dashes with underscores is indeed one of the differences, but it is not complete.

In order to properly address this, we simply need to do as wheel's changelog indicated. The code was moved to setuptools, so we have to use the code that is now in setuptools.

So we need to import `get_platform` from its new home, `setuptools.command.bdist_wheel` I believe is the new namespace.

Korijn:
In fact, according to the notice on wheel's project page, we should stop using the `wheel` package and dependency as the components we use have been entirely absorbed into setuptools.

setup.py
```diff
@@ -2,8 +2,8 @@
 import platform
 
 from setuptools import find_packages, setup
-from wheel.bdist_wheel import get_platform, bdist_wheel as _bdist_wheel
-
+from setuptools.command.bdist_wheel import get_platform
+from wheel.bdist_wheel import bdist_wheel as _bdist_wheel
```
Korijn, line 6: This should also be imported from setuptools

Korijn (approved):

PR #554, opened 2024-08-13
fix example error

rede97:


examples/triangle.py
```diff
@@ -98,10 +99,7 @@ def setup_draw(context, device):
             "targets": [
                 {
                     "format": render_texture_format,
-                    "blend": {
-                        "color": {},
-                        "alpha": {},
-                    },
+                    "blend": None,
```
fyellin, line 102: Why not just leave `blend` off completely, both here and in cube.py?

examples/cube.py
```diff
@@ -222,7 +222,9 @@
     {
         "binding": 0,
         "visibility": wgpu.ShaderStage.VERTEX | wgpu.ShaderStage.FRAGMENT,
-        "buffer": {},
+        "buffer": {
```
fyellin, line 225: Just curious.  Is there a reason you added these?  In each case, you've added the default value for the option.

Korijn:
I'm assuming you are running example code from the tip of the main branch against a wgpu-py release installed with pip from pypi.

See this comment for an explanation of your situation: https://github.com/pygfx/wgpu-py/issues/537#issuecomment-2202270028

If not, please share full details of which commit of the repo you are running and how you created your wgpu-py development environment.

almarklein:
> See this comment for an explanation of your situation: https://github.com/pygfx/wgpu-py/issues/537#issuecomment-2202270028

This seems to be the same scenario indeed :)

The examples and tests contain some variants of using default values, sometimes omitting, sometimes setting to None or empty dict, mainly to touch a variety of them.

PR #563, opened 2024-09-02
set pixel ratio in imgui renderer for each render

### LINKED ISSUE #558: live `pixel_ratio` updates for `ImguiRenderer`
If the `pixel_ratio` of the canvas changes after it has been created then imgui elements are not drawn to the correct area, so it needs to be set when it changes or just set on ever render call.

I will suggest a PR to this tomorrow/later, just documenting so I don't forget because I fixed this locally a few days ago.

kushalkolar:
The pixel ratio of the canvas can change after the canvas has been created, this sets the pixel ratio on every render to make sure it's correct.

A better way to do this is if all canvases can have a callback when the pixel ratio changes. 

@panxinmiao 

closes #558

### DIFF
--- a/wgpu/utils/imgui/imgui_renderer.py
+++ b/wgpu/utils/imgui/imgui_renderer.py
@@ -126,6 +126,9 @@ class ImguiRenderer:
         imgui.set_current_context(self.imgui_context)
         draw_data = self._update_gui_function()
 
+        pixel_ratio = self._canvas_context.canvas.get_pixel_ratio()
+        self._backend.io.display_framebuffer_scale = (pixel_ratio, pixel_ratio)
+
         command_encoder = self._backend._device.create_command_encoder()
         current_texture_view = self._canvas_context.get_current_texture().create_view()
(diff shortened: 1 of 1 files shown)


panxinmiao:
Uh, in what situations would the `pixel_ratio` of a canvas change after it has been created? 🤔 
I thought `pixel_ratio` was a read-only attribute before. 
Does Pygfx's Renderer have the same issue?

Korijn:
> Uh, in what situations would the `pixel_ratio` of a canvas change after it has been created? 🤔 I thought `pixel_ratio` was a read-only attribute before. Does Pygfx's Renderer have the same issue?

Moving from screen to screen typically.

kushalkolar:
> Uh, in what situations would the `pixel_ratio` of a canvas change after it has been created? 🤔 
> I thought `pixel_ratio` was a read-only attribute before. 
> Does Pygfx's Renderer have the same issue?

In jupyter when you change the zoom level of the web page.

Korijn (approved):

panxinmiao (approved):

PR #566, opened 2024-09-04
Rename wgpu features to be consistent with webgpu features

### LINKED ISSUE #553: Inconsistent naming of required features
When specifying required features in `adapter.request_device()`, the naming convention of features found in the `enummap` have a different format then the native features that are part of WGPU.

- The former must be specified precisely in lower case with a hyphen separating the words.
- The latter must be specified either in CamelCase or in any mixture of upper and lower with _ separating the words.

So one writes: 

```
    device = adapter.request_device(
          required_features=["indirect-first-instance", "VERTEX_WRITABLE_STORAGE", ]
    )
```
This seems inconsistent and confusing.  A required feature is a required feature. Both should be specifed the same way.

My preference is for lower-case-words-separated-by-hyphens.

fyellin:
FIX #553 

Name wgpu features using hyphenated lowercase, just as webgpu features are named.

Automatically generate wgpu features from wgpu.h rather than having a static list.

### DIFF
--- a/codegen/wgpu_native_patcher.py
+++ b/codegen/wgpu_native_patcher.py
@@ -16,7 +16,7 @@ focuses on the API, here we focus on the C library usage.
 import re
 from collections import defaultdict
 
-from codegen.utils import print, blacken, Patcher
+from codegen.utils import print, blacken, Patcher, to_snake_case
 from codegen.hparser import get_h_parser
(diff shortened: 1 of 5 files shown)


almarklein (approved):
Nice

Korijn (approved):

PR #569, opened 2024-09-05
Prevent UI events from propagating beyond imgui

panxinmiao:
After a UI event is handled by imgui, we generally don't want the event to be further processed by subsequent handlers (e.g., Pygfx).

For instance, in the following case, it is quite inconvenient

https://github.com/user-attachments/assets/2d8caba3-e76b-4628-ab03-ec58202da08c

This PR allows imgui to capture canvas UI events and prevent them from further propagation, as shown below:

https://github.com/user-attachments/assets/09093c57-3c3b-49cb-a301-c748ed3a4c44

PS: The implementation might be a bit hacky, 😅 , and I’d be happy to hear your thoughts or suggestions.


wgpu/gui/base.py
```diff
@@ -274,18 +274,22 @@ def handle_event(self, event):
         """
         # Collect callbacks
         event_type = event.get("event_type")
-        callbacks = self._event_handlers[event_type] | self._event_handlers["*"]
+        callbacks = self._event_handlers[event_type] + self._event_handlers["*"]
         # Dispatch
         for callback in callbacks:
             with log_exception(f"Error during handling {event['event_type']} event"):
+                if event.get("stop_propagation", False):
+                    del event["stop_propagation"]
```
almarklein, line 282: I don't think deleting is necessary. An event is a one-time object.

wgpu/gui/base.py
```diff
@@ -330,8 +334,13 @@ def my_handler(event):
                 raise TypeError(f"Event types must be str, but got {type}")
 
         def decorator(_callback):
+            _callback.__dict__["__event_order"] = order
```
almarklein, line 337: I don't think its a good idea to store order on the callback object. E.g. the same function may be provided multiple times, for different event types, and with different orders. Also, callback may not allow setting attributes.

I think its better to instead do `self._event_handlers[type].append((order, _callback))`.

wgpu/utils/imgui/imgui_renderer.py
```diff
@@ -152,11 +154,17 @@ def _on_resize(self, event):
     def _on_mouse_move(self, event):
         self._backend.io.add_mouse_pos_event(event["x"], event["y"])
 
+        if self._backend.io.want_capture_mouse:
```
almarklein, line 157: Is `want_capture_mouse`  a think from imgui?

almarklein (commented):
I think the general approach is good. I made a few comments.

We should also document this as part of the jupyter_rfb event spec, but I can do that later.

kushalkolar:
Yay! Related: https://github.com/pygfx/pygfx/issues/832

wgpu/utils/imgui/imgui_renderer.py
panxinmiao: Yes, See: https://github.com/ocornut/imgui/blob/master/docs/FAQ.md#q-how-can-i-tell-whether-to-dispatch-mousekeyboard-to-dear-imgui-or-my-application.

panxinmiao:
Updated. :)

Korijn:
* It's indeed a bit hacky to use this `order` argument and just fix the imgui renderer on `-99`... I wonder what a more general solution would look like. Is purely insertion order and the `stop_propagation` feature not enough?

If not, I'm fine with merging this.

panxinmiao:
> * It's indeed a bit hacky to use this `order` argument and just fix the imgui renderer on `-99`... I wonder what a more general solution would look like. Is purely insertion order and the `stop_propagation` feature not enough?
> 

We need to provide a way to control the priority order of UI event propagation. For example, we would generally expect Imgui's event handler to be at the front of the queue. However, we cannot guarantee the order in which users initialize the ImguiRenderer and Pygfx Renderer, so we cannot control the insertion order of the event handlers are added.

almarklein (approved):
Looking good. I can't think of a better way to control the order. 

We can merge this before #573, so it ends up in the 0.17 release. Otherwise it will be part of a next release, which is also fine.

Korijn (approved):

PR #570, opened 2024-09-05
Refactor and fix cd

almarklein:
The release did not go through due to issues with cd. I fixed it (I think) by bumping the version numbers of actions. I also bumped for other actions, and refactored the cd into its own script.

* [x] Bump action versions
* [x] Increase Python versions to 3.12
* [x] Remove `set-output`, because its deprecated  
* [x] Intermediate build artifacts have separate names.


Korijn:
There have been a few releases where suddenly it became impossible to write multiple times to the same build artifact. This causes issues potentially. Did you confirm the output is correct now?

almarklein:
> Did you confirm the output is correct now?

Working on it ...

almarklein:
I made a draft release, with the pypi code disabled, to test that it works. I have deleted the draft release.

Korijn (approved):

PR #573, opened 2024-09-09
Fix pypy release in cd

almarklein:
Also bumps version


Korijn (approved):

almarklein:
Ah crap, the pypi action also needs to take into account that the builds are in subdirs now :/

PR #574, opened 2024-09-09
Implement push-constants

fyellin:
Initial implementation for push_constants. 

Since the only other implementations of this are in C-like interfaces, there are some open API questions.

1) In the binding code, we use start/end while in set_push_constant we use start/length.  This is from the underlying API.  Should we keep this?

2) Do we want to add a data_offset argument to set_push_constants()? It's not in the Rust API, but might be useful.

In general, any of the added functions or functionality could easily have their APIs better designed for consistency.


almarklein:
As for the API, we follow WebGPU, only deviating when necessary.

Looks like push constants are part of the `ProgrammableStage`  dict that's embedded in the args passed to `create_render_pipeline()`

image

That `constants` must be a dict mapping constant name to constant value.

fyellin:
That's a different feature, but with a similar name. It lets you create variables that are "almost constant" but can be changed at pipeline creation time.

fyellin:
The API for createPipelineLayout was inspired by Rust's [PipelineLayoutDescriptor](https://docs.rs/wgpu/latest/wgpu/struct.PipelineLayoutDescriptor.html) which has two public fields, `bind_group_layouts` and `push_constant_ranges`.  I let the fields of the dictionary match the fields of the wgpu.h API, except that stages become visibility in line with other locations.

The API for set_push_constants is precisely equivalent to that in wgpu.h.  I think there needs to be an optional `data_offset` field added to make this code similar to `write_buffer`.

I'm a little bit bothered that limits are snake_case_with_underscores while features are snake-case-with-hyphen.  This makes the code for `adapter.get_device()` look a little bit strange.  Should this be fixed, or should we allow either underscore or hyphen in the name, with hyphen being the default?  [That would be a different push, I susupect].

almarklein:
> That's a different feature, but with a similar name. 

Ah, both of these are new to me, so I mixed them up.

If I understand correctly, push constants are a feature specific to wgpu-native, not present in WebGPU. In that case, the API should not be in `_classes.py`, because that api should be mappable to any WebGPU(ish) backend. So far, we've implemented wgpu-native-specific features in [backends/wgpu_native/extras.py](https://github.com/pygfx/wgpu-py/blob/main/wgpu/backends/wgpu_native/extras.py), which get exposed in the `wgpu.backends.wgpu_native` namespace. That way it is clear when code uses 'special' functionality and might therefore be less portable.

wgpu/backends/wgpu_native/_api.py
```diff
@@ -778,24 +808,44 @@ def _request_device(
...
+
+        if required_limits:
+            # Do some quick error checking
+            unknown_keys = [x for x in required_limits if x not in self._limits]
+            if unknown_keys:
+                raise KeyError(f"Unknown required limits {', '.join(unknown_keys)}")
+        else:
+            # If required_limits isn't set, set it to self._limits.  This is the same as
+            # setting it to {}, but the loop below goes just a little bit faster.
+            required_limits = required_limits or {}
```
almarklein, line 836: The code does something different than what the comment says it does. Also `bool(required_limits)` is  already established to be false here.

almarklein (commented):
Nice work on the limits and features! I agree that we should use a consistent naming scheme.

For compatibility, are both `dash-case`  and `snake_case`  allowed when setting features/limits? Maybe have a test for this to keep it that way.

wgpu/backends/wgpu_native/_api.py
fyellin: Fixed.  My mistake.

fyellin:
Merge is complete and is ready to be reviewed (but not quite ready for submission).  

When I run pytest tests_mem on my Mac, everything works fine.  But the test fails as part of the test suite on Linux with 

``
----------------------------- Captured stdout call -----------------------------
  more after create: {'RenderBundle': (32, 32), 'RenderBundleEncoder': (32, 0)}
=========================== short test summary info ============================
FAILED tests_mem/test_objects.py::test_release_render_bundle - AssertionError: Expected:
{'RenderBundle': (32, 32)}
Got:
{'RenderBundle': (32, 32), 'RenderBundleEncoder': (32, 0)}
=================== 1 failed, 28 passed, 2 skipped in 8.11s ====================
``

I've no idea why I'm getting different results on the two.  I also confess I don't understand what `@test_and_release` is testing and what the two numbers are supposed to represent.  The decorator has no documentation saying what its arguments mean.

fyellin:
This is ready for review.
I'm not completely happy with test_objects(), but both both branches "main" and this branch both fail on my Mac. Apparently there is a slight difference in the implementation of render bundles.  

@almarklein: I do wish it were better documented what @create_and_release was testing, and what the two numbers are. Python objects and native objects?

almarklein:
> @almarklein: I do wish it were better documented what @create_and_release was testing, and what the two numbers are. Python objects and native objects?

Sorry for the missing docs here. Yeah, they are Python and native (Rust) objects. Would you mind adding that to the docs in this pr?

wgpu/backends/wgpu_native/_api.py
```diff
@@ -2782,6 +2838,7 @@ class GPUComputePassEncoder(
     GPUBindingCommandsMixin,
     GPUObjectBase,
 ):
+    """ """
```
almarklein, line 2841:
```suggestion
```

almarklein:
> I'm not completely happy with test_objects(), but both both branches "main" and this branch both fail on my Mac. Apparently there is a slight difference in the implementation of render bundles.

Is that resolved? I don't see a change in `test_objects()` now.

fyellin:
The test_objects problem was resolved. A rechecking of _api.py indicated
that I had missed a tiny change where self was replaced with self._device.

I will look at other comments this afternoon.

docs/backends.rst
```diff
@@ -59,6 +59,95 @@ The wgpu_native backend provides a few extra functionalities:
     :return: Device
     :rtype: wgpu.GPUDevice
 
+The wgpu_native backend provides support for push constants.
```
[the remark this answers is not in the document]
fyellin: Done.

I'm not really sure.  Googling "webgpu push_constant" led to some discussion and some indication that they were considering it, but nothing definitive.  Maybe not enough machines supported it?

https://github.com/gpuweb/gpuweb/issues/75

wgpu/backends/wgpu_native/extras.py
```diff
@@ -33,3 +33,32 @@ def request_device_tracing(
     return adapter._request_device(
         label, required_features, required_limits, default_queue, trace_path
     )
+
+
+def create_pipeline_layout(
```
fyellin, line 38: Quick question for @almarklein.  I noticed that for request_device, you gave the version in extras.py the name request_device_tracing.

I made the extras.py version of create_pipeline_layout have the same name.  I prefer that, but I'll change the name if you want consistency.

almarklein: Yeah good point. In that case it was pretty obvious, because it only added the tracing feature. But this won't hold up when the wgpu_native version of a function supports multiple extra functionalities. I like what you did. I think it would be good to rename `request_device_tracing` in a new pr after this.

docs/backends.rst
```diff
@@ -59,6 +59,104 @@ The wgpu_native backend provides a few extra functionalities:
...
+Since WebGPU does not this feature, documentation on its use is hard to find.
```
almarklein, line 63:
```suggestion
Since WebGPU does not support this feature, documentation on its use is hard to find.
```

docs/backends.rst
```diff
@@ -59,6 +59,104 @@ The wgpu_native backend provides a few extra functionalities:
...
+A full explanation of push constants and its use in Vulkan can be found
+`here <https://vkguide.dev/docs/chapter-3/push_constants/>`_.
+Using push constants in WGPU closely follows the Vulkan model.
+
+The advantage of push constants is that they are typically faster to update than uniform buffers.
+Modifications to push constants are included in the command encoder; updating a uniform
+buffer involves sending a separate command to the GPU.
+
+The disadvantage of push constants and that their size limit is much smaller. The limit
+is guaranteed to be at least 128 bytes, and 256 bytes is typical
```
almarklein, line 73:
```suggestion
The disadvantage of push constants is that their size limit is much smaller. The limit
is guaranteed to be at least 128 bytes, and 256 bytes is typical.
```

tests_mem/testutils.py
```diff
@@ -145,7 +145,37 @@ def ob_name_from_test_func(func):
...
 def create_and_release(create_objects_func):
-    """Decorator."""
+    """
+    This wrapper goes around a test that takes a single argument n. That test should
+    be a generator function that yields a descriptor followed
+    n different objects corresponding to the name of the test function.  Hence
+    a test named `test_release_foo_bar` would yield a descriptor followed by
+    n FooBar objects.
+
+    The descriptor is a dictionary with three fields, each optional.
```
almarklein, line 155:
```suggestion
    The descriptor is a dictionary with three fields, each optional.
    In a typical situation, there will be `n` FooBar object after the test, and after
    releasing, there will be zero. However, sometimes there are auxiliary objects,
    in which case its necessary to provide one or more fields.
```

tests_mem/testutils.py
```diff
@@ -145,7 +145,37 @@ def ob_name_from_test_func(func):
...
+    gives the state after the n objects have been released.
+
+    These sub-dictionaries have as their keys the names of WGPU object types, and
+    their value is a tuple of two integers: the first is the number of Python objects
+    expected to exist and the second is the number of native objects. Any type not in
+    the subdictionary has an implied value of (0, 0).
+
+    The key "ignore" has as its value a collection of object types that we should ignore
+    in this test. We do not have enough information to determine how many are created
+    or deleted.
```
almarklein, line 170:
```suggestion
    The key "ignore" has as its value a collection of object types that we should ignore
    in this test. Ideally we should not use this, but currently there are a few cases where
    we cannot reliably predict the number of objects in wgpu-native.
```

almarklein (commented):
Only a few typos and adding some extra context in the memtest docs.

almarklein (approved):

Korijn (approved):

almarklein:
Thanks!

PR #575, opened 2024-09-10
Fix CD by flattening dist dir

almarklein:


Korijn (approved):

almarklein:
The twine uploads still failed, because it gets confused by the directories that are still there. I just downloaded the artifacts from github, put them in my local `dist` folder and ran `twine upload`.

PR #577, opened 2024-09-11
typo pass

Vipitis:
Typos pass, used https://github.com/crate-ci/typos to find them and then manually checked and fixed all findings.

### DIFF
--- a/examples/wgpu-examples.ipynb
+++ b/examples/wgpu-examples.ipynb
@@ -109,7 +109,7 @@
    "name": "python",
    "nbconvert_exporter": "python",
    "pygments_lexer": "ipython3",
-   "version": "3.9.4"
+   "version": "3.9.9"
   }
  },
  "nbformat": 4,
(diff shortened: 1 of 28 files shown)


Korijn (approved):

examples/wgpu-examples.ipynb
```diff
@@ -109,7 +109,7 @@
    "name": "python",
    "nbconvert_exporter": "python",
    "pygments_lexer": "ipython3",
-   "version": "3.9.9"
+   "version": "3.9.4"
```
almarklein, line 112: This is generated content, a false positive?
```suggestion
   "version": "3.9.9"
```

almarklein:
Thanks, there are so many! 😊

examples/wgpu-examples.ipynb
Vipitis: I tried three times to only change the markdown cell... But I guess saving the notebook adds some information about my env. 
Also didn't show up in my diff due to the rich rendering of notebooks. 

will revert

PR #579, opened 2024-09-12
overrideable constants

fyellin:
Implement  [override constants](https://www.w3.org/TR/webgpu/#gpuprogrammablestage)
Allow entry_point to be elided if there is only one possible entry point.
Create functions for some common code in _api.py.


almarklein (approved):
Nice work!

Korijn (approved):

PR #580, opened 2024-09-13
Update native

almarklein:
Follow-up on #547. This bumps the minor version of wgpu-native. Two things have changed that this PR accounts for:

* A bug in adapter info was fixed that caused the fields to be mingled. 
* The archives have a nested folder structure.

### DIFF
--- a/wgpu/_classes.py
+++ b/wgpu/_classes.py
@@ -391,7 +391,7 @@ class GPUAdapter:
     def summary(self):
         """A one-line summary of the info of this adapter (description, adapter_type, backend_type)."""
         d = self._adapter_info
-        return f"{d['description']} ({d['adapter_type']}) via {d['backend_type']}"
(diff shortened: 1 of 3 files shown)


wgpu/_classes.py
```diff
@@ -391,7 +391,7 @@ def info(self):
     def summary(self):
         """A one-line summary of the info of this adapter (description, adapter_type, backend_type)."""
```
Vipitis, line 392: also device here

Vipitis: and here https://github.com/pygfx/wgpu-py/blob/de55ef8eabd3359af24bbd9c77009eebe03e6c30/wgpu/backends/wgpu_native/_api.py#L348

almarklein: Nice catch, thanks!

Korijn (approved):

PR #582, opened 2024-09-13
More binary builds v2

### LINKED ISSUE #536: Add Windows aarch64
Supported in wgpu-native: https://github.com/gfx-rs/wgpu-native/pull/398

almarklein:
* [x] Include Windows aarch64 and 32bit builds, closes #536
* [x] Musllinux would be nice, but still an open issue at wgpu-native https://github.com/gfx-rs/wgpu-native/issues/414
* [x] One build per binary, so easier to configure, and easier to extend in the future.

### DIFF
--- a/.github/workflows/cd.yml
+++ b/.github/workflows/cd.yml
@@ -33,6 +33,7 @@ jobs:
           - platform: windows
             arch: AMD64
             os: windows-latest
+            testable: true
           - platform: windows
             arch: ARM64
             os: windows-latest
@@ -42,6 +43,7 @@ jobs:
           - platform: macos
             arch: arm64
             os: macos-latest
+            testable: true
           - platform: macos
             arch: x86_64
             os: macos-13  # latest Intel MacOS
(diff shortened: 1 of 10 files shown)


.github/workflows/cd.yml
```diff
@@ -11,70 +22,75 @@ on:
...
+          - platform: windows
+            arch: x86
+            os: windows-latest
+          - platform: macos
+            arch: arm64
+            os: macos-latest
+            testable: true
+          - platform: macos
+            arch: x86_64
+            os: macos-13  # latest Intel MacOS
```
Korijn, line 49: latest or last? :o

almarklein: last :)

almarklein:
Need to update the rules for the protected branch to get match the new names of the builds

almarklein:
BTW, we already build a wheel for win32, but that was probably broken because it included the 64bit lib :)

Korijn (dismissed):
Impressive work! Shame the windows wheel test won't work. You could capture the filename in a variable, if that's the issue?

`filename=$(ls dist/*.whl)`
`pip install $filename`

almarklein:
That did the trick, thanks!

Korijn (approved):

Issue #92, opened 2020-04-16
Linux Wayland support


almarklein:
Linux X11 works fine. But Wayland not yet. We should probably wait until Wayland becomes more stable/mainstream.

almarklein:
wgpu-native just got support: https://github.com/gfx-rs/wgpu-native/pull/160/ but apparently GLFW's support for Wayland is still experimental and pyGLFW may not support it yet.

almarklein:
It looks like all the components are there to connect things up (with glfw), but when I run it, the system crashes and I'm back in my login screen 😆   I've spend some time checking and trying things, but no luck.

For Qt, this seems like a good staring point: https://doc.qt.io/qt-5/wayland-and-qt.html   I have not looked at that myself yet.

Korijn:
Current status: it's not known if the latest versions of GLFW and wgpu-py together support wayland or not. Needs verification before we can close.

hmaarrfk:
I'm not too sure about the packages on pypi, but I've compiled qt and wayland for conda-forge:
https://github.com/conda-forge/qt-wayland-feedstock

https://github.com/conda-forge/qt-main-feedstock/issues/120#issuecomment-1546916602

let me know if that helps.

If you feel that an other package is missing wayland support, I've found that:
1. Upstream typically supports it in their development environmnet.
2. Packaging strategies for PyPi somewhat lag support for fancy features due to packaging difficulties.

2 has been largely alleviated recently with the introduction of things like manylinux_20YY.

hmaarrfk:
> Qt did not work on Wayland.

can you expand on this? I was trying this a while back too, it seems that Python for Qt (PyQt6 and PySide6)  likely doesn't expose the API required to make this work. My next attempt was to dig deeper into the C++ API to see if there was a small patch I could make to pyside2 to expose the required functions. 

I'm not sure if you already tried this.

hmaarrfk:
> glfw did, kinda,[...] undecorated window. 

regarding undecorated windows, it seems that this is a "choice" by Wayland, they are pushing the decorations on the client side, many frameworks now have a "switch" that you have to toggle if you want the "default" decorations. 

I the decison to make the the client's (the application we are writing in our python+wgpu context) choice was one that came about because Chrome + Firefox were starting to try to make more use of the space typically reserved for "decorations".

almarklein:
>>    Qt did not work on Wayland.
>
> can you expand on this? 

[...]

Issue #521, opened 2024-06-11
How to get a transparent Canvas?


panxinmiao:
I want to render some scenes to the desktop using a transparent window. This requires setting the `alpha_mode` to `premultiplied` in the `CanvasContext.config()` method.

I have added a flag in the wgpu glfw backend code with a line like this: 
`glfw.window_hint(glfw.TRANSPARENT_FRAMEBUFFER, True)`. 

However, the `alpha_mode="premultiplied"` setting in the `CanvasContext.config()` method still seems to be unavailable.

The `capable_alpha_modes` here are always `['opaque']`.
https://github.com/pygfx/wgpu-py/blob/0cdef1deb7033a15adb646d4b151d6d97cef954c/wgpu/backends/wgpu_native/_api.py#L449-L454

Am I missing something? How can we get a canvas that supports `alpha_mode="premultiplied"`?

almarklein:
Interesting. I just tried with this:
```py
import glfw
import wgpu
from wgpu.gui.glfw import WgpuCanvas

glfw.init()
glfw.window_hint(glfw.TRANSPARENT_FRAMEBUFFER, True)
canvas = WgpuCanvas()
c = canvas.get_context()
c.configure(
    device=wgpu.utils.get_default_device(),
    format=wgpu.TextureFormat.bgra8unorm_srgb, 
    alpha_mode=wgpu.CanvasAlphaMode.premultiplied,
)
```

I get this error:
```
ValueError: Given format 'premultiplied' is not in supported formats ['opaque', 'unpremultiplied']
```

Interestingly, it does support `unpremultiplied`? which is not in the `CanvasAlphaMode` enum.


I found https://github.com/gfx-rs/wgpu/issues/3486 which relates to this issue. Following some links to referred issues suggests that others have similar issue. E.g. *Appears to be an issue with wgpu not correctly detecting surface capabilities. [SurfaceCapabilities::alpha_modes](https://docs.rs/wgpu/latest/wgpu/struct.SurfaceCapabilities.html#structfield.alpha_modes) only contains Opaque on my machine, Windows 10, nvidia gpu.*

PR #586, opened 2024-09-16
Refactor canvas context to allow presenting as image

almarklein:
### Background

This is a first step in the "update propagation" work. One of the goals of that work is to clearly separate rendering from events. And thus webgpu API from canvas API. A thought that played through my mind was to implement the `wgpu.gui` in a separate package. I don't think I want to go that far (for now), but I definitely want to avoid leaking any wgpu-specifics into the gui subpackage.

###  Problem

The CanvasContext is the thing that ties  webgpu and the gui layer together. It enables webgpu to draw into a GUI window. This works ok as it is, but e.g. the offscreen canvas implements its own version of a CanvasContext, which is an abstraction leak.

### Proposal

This PR refactors the `CanvasContext` as well as the `WgpuCanvasInterface` so that the canvas can indicate how it wants the canvas to draw into it. It can either let it draw to the screen, or provide the canvas with the resulting image.

This makes offscreen canvases less special, so they won't have to hack their own fake context. It also allows other canvases to use images where it makes sense, e.g. a Tk backend, or Qt on Wayland, etc.

### Details

The `WgpuCanvasInterface.get_surface_info()` returns a dict. That dict has at least a "method" field, which is either "screen" or "image". When it is "screen", it also provides "window" (and "display" too on Linux), like it did before. The dict can also provide capabilities like "formats" to indicate supported texture formats.

The `WgpuCanvasInterface.present_image()` method must be implemented if a canvas wants to support the image method.

### Benchmarks

FPS measurements (with `max_fps=999`). Shown are fps for screen / image. Using a light scene (cube) and a heavy one (sponza).

|  | 640x480 cube | fullscreen cube | 640x480 sponza | fullscreen sponza |
|--|----------------|-------------------|-------------|--------------|
| MacOS M1                                           | 120 / 130       |      110 / 33    |    28 / 22      |     23 / 14        |
| Windows with Geforce RTX 2070       | 144 /  105     |       67 / 47     |     9 / 9         |    8 / 8             |
| Linux with Intel UHD 730                     |  60 / 190       |      60 / 38     |    17 / 9          |    4 / 3          |
| Windows with Intel UHD 730 on Vulkan            |   60 / 150   |         56 / 37       |   23 / 14        |   10 / 7         |
| Windows with Intel UHD 730 on D3D12            |    60 / 120  |         56 / 35       |   na               |   na               |

Observations:
* There's certainly a performance penalty for rendering via an image.
* The penalty scales with the size of the window, as expected.
* When drawing to screen, we can observe an fps limit that's not present when drawing via an image.
* My Windows laptop's GPU is rubbish?

Remarks:
* This was tested with Qt. I would expect that Qt renders the image to screen really fast, but other solutions *might* do it faster. 

### Tasks

* [x] Basics working with Qt
* [x] Do performance benchmarks to have a better idea of the penalty for using the image method.
* [x] Remove flicker on Windows (introduced in (earlier version of) this pr).
* [x] Fix panic when process ends.
* [x] ~Can we get rid of `CanvasContext.present()`, since its not official webgpu api?~ later
* [x] Adjust other backends.
* [x] Use this as a workaround to fix Qt support on Wayland.
* [x] Clean up / check todos.
* [x] Update status in #92 

### Changelog

* The `WgpuOffscreenCanvasBase` class is removed. Offscreen canvases can be implemented by returning the appropriate dict in `canvas.get_surface_info()` and implementing `canvas.present_image()`.
* The `canvas.get_surface_info()` method is renamed `canvas.get_present_info()` and must the returned dict has new specifications.
* The `canvas.present_image()` method can be implemented to support presenting via an image.
* The Qt and wx backends now also work on Wayland (though with a performance penalty).


almarklein:
I have more plans with the GUI subpackage, but I'll apply/propose these in separate pr's.

almarklein:
@Korijn this is ready for review.

wgpu/_classes.py
```diff
@@ -196,6 +215,41 @@ def canvas(self):
...
+            self._capabilities = {}
+            if self._present_info["method"] == "screen":
+                # Query capabilities from the surface
+                self._capabilities.update(self._get_capabilities_screen(adapter))
+            else:
+                # Default image capabilities
+                self._capabilities = {
+                    "formats": ["rgba8unorm-srgb", "rgba8unorm"],
+                    "usages": 0xFF,
+                    "alpha_modes": [enums.CanvasAlphaMode.opaque],
```
Korijn, line 230: Is this true? Can't we render transparent PNGs?

Korijn (approved):
Nice improvements!

wgpu/_classes.py
almarklein: We probably can. But I wanted to leave that for another pr. See also #521

PR #591, opened 2024-09-19
Fix some small errors

fyellin:
Forgot to delete one _release

Allow most of tests in test_wgpu_vertex_instance to run even if multi-draw-indirect isn't available.  This will be useful in the future as we add more features. Also simplified the code a little bit.

Fixed one test where I believe naga's behavior is incorrect.  I updated the test to accept either naga's current behavior or the behavior of the JS implementation.

### DIFF
--- a/tests/test_set_override.py
+++ b/tests/test_set_override.py
@@ -2,7 +2,7 @@ import pytest
 
 import wgpu.utils
 from tests.testutils import can_use_wgpu_lib, run_tests
-from wgpu import TextureFormat
+from wgpu import GPUValidationError, TextureFormat
 
 if not can_use_wgpu_lib:
     pytest.skip("Skipping tests that need the wgpu lib", allow_module_level=True)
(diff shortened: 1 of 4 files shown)


tests/test_set_override.py
```diff
@@ -197,10 +197,16 @@ def test_override_compute_constants(runner):
...
 def test_numbered_constants_must_be_overridden_by_number(runner):
     overrides = {"c": 23, "d": 24}
-    # This does absolutely nothing.  It doesn't even error.
-    assert [1, 2, 3, 0, 1, 2, 3, 0] == runner.run_test(
-        render=True, vertex_constants=overrides, fragment_constants=overrides
-    )
+    try:
+        # In naga, the bad constant is ignored.
+        # In the JS implementation, this throws an exception, which I think is the
+        # correct behavior.  So just in case this ever gets fixed, we accept either.
```
Korijn, line 203: Would you want to be notified when it gets fixed? Cause the test to fail?

fyellin: Yeah, the alternative is to let the test fail, and leave a long comment explaining what's going on and how to fix it. 

You're right, that's probably better.  I'll fix.

fyellin:
I don't have sufficient privileges (write access) to do the merge.

PR #594, opened 2024-09-20
multi_draw_indirect_counter

fyellin:
Implementation of the final two functions.
Fix some typos I had previous made.


fyellin:
This code is working now, but it wasn't working an hour ago, and I didn't really do anything to fix it. I want to make sure I understand what's going on before this is merged.  But it's rather straightforward code.

docs/backends.rst
```diff
@@ -184,10 +188,51 @@ they reduce driver overhead on the CPU.
...
+They are identical to the previous two, except that the ``count`` argument is replaced by
+three arguments. The value at ``count_buffer_offset`` in ``count_buffer`` is treated as
+an unsigned 32-bit integer. The ``count`` is the minimum of this value and ``max_count``.
+
+.. py:function:: wgpu.backends.wgpu_native.multi_draw_indirect_count(render_pass_encoder, buffer, *, offset=0, count_buffer, count_offset=0, max_count):
+
+     Equivalent to::
+        count = min(<u32 at count_buffer_offset in count_buffer>, max_count)
+        for i in range(count):
+            render_pass_encoder.draw_indirect(buffer, offset + i * 16)
```
fyellin, line 207: Help! Sphinx is complaining about bad indentation here and on line 222.  Surprisingly there is no complaint about the previous two occurrences of this line.  Something to do with the <> on line 205?

I can't find any reasonable documentation on sphinx.  How do I fix this??

fyellin:
The naga bug is back.

docs/backends.rst
almarklein: I think it needs a newline after `::`.

fyellin: Thanks!

fyellin:
This code is waiting for a new wgpu-native-lib to include:

https://github.com/gfx-rs/wgpu-native/pull/432
https://github.com/gfx-rs/wgpu/pull/6194

We pull directly from wpgu-native, so their next update should include the change.  I'm not sure how wgpu-native decides which version of wgpu to use.

almarklein:
>  I'm not sure how wgpu-native decides which version of wgpu to use.

Here's an example: https://github.com/gfx-rs/wgpu-native/pull/402  Basically bumping versions in `cargo.toml` and then making the necessary changes. It usually updates for each release of wgpu-core.

fyellin:
@almarklein.

I ask because it seems that tag 22.1.0.5 was created on gfx-rs/wgpu-native on Sept 12.

Meanwhile I don't see any tags on gfx-rs/wgpu, and the fix was merged into "trunk" on Sept 2.  But it doesn't seem to have made it into the build of lib-native.

almarklein:
That's correct. The latest tag from wgpu-core is 22.1.0. The versioning of wgpu-native is to take the version of wgpu-core and add a fourth number to it. We've had a few releases (22.1.0.x) for bugfixes and changes to how the binary archives are composed.

fyellin:
I've confirmed that this code works with a patched wgpu. I've tried it out both on emulated GPUs and on real GPUs. 

I just need to wait for us to catch up.

fyellin:
Do we have any news as to when there will be a new wgpu-native?  This PR needs big foxes there in order to work.

almarklein:
The work has started some time ago: https://github.com/gfx-rs/wgpu-native/pull/441

fyellin:
I compiled that PR along with its wgpu.h and webgpu.h, and ran it against
our test suite.  Almost everything worked perfectly.  The only issues were
that some error messages changed slightly (our tests looked for the exact
test of the error message), and that the output of GlobalReport has changed
so that test_objects and test_destroy weren't working.

I did it on a Mac, so couldn't test multi draw, but I'll try again later on
a Linux.

So I'm hoping our porting efforts will be easy.

PR #596, opened 2024-09-20
Refactor build system

### LINKED ISSUE #585: Windows aarch64 wheel is named incorrectly
image

So the amd64 and arm64 builds both produce a wheel with the same name, ending in "_amd64.whl". I don't know which of the two is on pypi 🤷

almarklein:
Closes #585. 

The approach to use `cibuildwheel` is somewhat overkill because we don't need to compile anything. And it does not even work properly. We added extra logic to make it work in #582 but then we still got #585. So I figured it can be done simpler, using the modern approach of a `pyproject.toml`.

* [x] Use `pyproject.toml`
* [x] Remove `setup.py`, `dev-requirements.txt`
* [x] Setup build system so it creates clean sdist and wheels.
* [x] Enable cross-platform builds.
* [x] Script to build all wheels. The wheels are checked extensively.
* [x] Update `cd.yml`, test wheels on 3 main platforms.


.github/workflows/cd.yml
```diff
@@ -24,118 +24,71 @@ on:
...
+        path: dist
+    - name: Install and test wheel
       shell: bash
       run: |
           rm -rf ./wgpu
-          filename=$(ls dist/*.tar.gz)
-          pip install $filename
-          # don't run tests, we just want to know if the sdist can be installed
+          pip install --find-links dist wgpu
+          pip install --force-reinstall --no-deps --no-index --find-links dist wgpu
```
Korijn, line 81: what's going on with these two lines?

tools/hatch_build.py
```diff
@@ -0,0 +1,78 @@
...
+            # State that the wheel is not cross-platform
+            build_data["pure_python"] = False
+
+            # Download and set tag
+            platform_info = os.getenv("WGPU_BUILD_PLATFORM_INFO")
+            if platform_info:
+                # A cross-platform build
+                wgpu_native_tag, wheel_tag = platform_info.split()
+                opsys, arch = wgpu_native_tag.split("_", 1)
+                build_data["tag"] = "py3-none-" + wheel_tag
```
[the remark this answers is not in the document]
almarklein: Yeah, unfortunately the build-backend does not check the validity of the flag.

.github/workflows/cd.yml
almarklein: The first is basically to install the deps. The second is to be absolutely sure that wgpu is installed from the `dist` directory and nowhere else. Will add a comment.

Korijn (approved):

.github/workflows/ci.yml
```diff
@@ -69,7 +69,6 @@ jobs:
       run: |
           python -m pip install --upgrade pip
           pip install requests
-          python download-wgpu-native.py
           pip uninstall -q -y requests
```
Korijn, line 72: I think you're literally installing and then deinstalling requests here

README.md
```diff
@@ -115,18 +115,15 @@ This code is distributed under the 2-clause BSD license.
 ## Developers
 
 * Clone the repo.
-* Install devtools using `pip install -r dev-requirements.txt` (you can replace
+* Install devtools using `pip install -e .[dev]` (you can replace
   `pip` with `pipenv` to install to a virtualenv).
```
Korijn, line 119: I'd suggest to remove the pipenv note

.github/workflows/ci.yml
almarklein: Aah, I saw that, wondered wtf that was, and forgot to get back to it. In the diff it's clear what the intention was 😅  Will fix

README.md
Korijn, line 1: Is this a good opportunity to note in the "installation" that if people want to run examples taken straight from the repo, they should instead choose from (1) running example code from the stable docs, (2) checking out the appropriate tag of the repo for the version they install with pip, or (3) follow the Developers setup instructions to go bleeding edge

almarklein: I'm not sure what's the best place to put a warning like this. The installation section is an option, but then we should also add the notes to the installation section in the docs.

Maybe a `examples/README.md` with a notice will work best, because people probably see that when they look at the examples on github?

almarklein: I created a separate pr:  #597

Korijn (approved):

PR #598, opened 2024-09-25
Async API

almarklein:
See #391, also relates to https://github.com/pygfx/pygfx/issues/495

## Introduction

This PR changes the API to expose the async methods and their sync (convenience) variants. This does not implement any of the async stuff at the level of wgpu-native (the async methods simply call the sync variant for now). That's for another PR. This does however, make it possible to look into using wgpu-py in an async setting.

## In short

* The codegen is updated to discover async method and automatically generate the correct API.
* The methods (and properties) that are async in WebGPU (i.e. return a `Promise`), all get a `_async` suffix in wgpu-py.
* All these methods also get a sync variant, but they come in two flavours:
  * Methods that also have a sync variant in WebGPU, are simply available without the suffix.
  * Methods that don't have a sync variant in WebGPU get a `_sync` suffix in wgpu-py. This way, they are easily identified, to detect less portable code.
* Backwards compatibility is implemented, which shows a warning for that function once. We can remove this after 1 or 2 versions, giving downstream projects time to adapt.

## Affected methods

* `gpu.request_adapter()`
* `gpu.enumerate_adapters()` 
* `adapter.request_device()` 
* `buffer.map()` 
* `device.lost`  (but was not yet implemented)
* `shadermodule.get_compilation_info()` (but was not yet implemented)
*  `queue.on_submitted_work_done()` (but was not yet implemented)

These method also have (and already had) an async variant:
* `create_compute_pipeline()`
* `create_render_pipeline()`

## Todo

* [x] Update codegen.
* [x] Tests for codegen.
* [x] Ran codegen and made adjustments.  
* [x] Docs. 
* [x] Backwards compat.
* [x] Tests pass with using the compat methods.
* [x] Update examples and tests to use new version.
* [x] Ability to disable `_sync` version.
* [x] Don't forget to enable backwards compat again (turned off for tests).

## API changes

* `gpu.request_adapter` is deprecated, `request_adapter_sync` instead.
* `gpu.enumerate_adapters` is deprecated, use `enumerate_adapters_sync` instead.
* `adapter.request_device` is deprecated, use `request_device_sync`instead.
* `buffer.map` is deprecated, use `map_sync` instead.


almarklein:
Ready. This basically applies the API changes to move further with async support.

Next up would be creating examples that use the API in an async setting with asyncio and trio. Another further step could be to change the implementation of wgpu-native so that using the async api actually helps performance. But I'll probably leave that for later.

This PR also prepares the way for implementing update propagation, taking async and portability into account.

Korijn (approved):

Korijn (approved):

PR #611, opened 2024-10-03
Make depth_stencil_attachment follow the spec

fyellin:
https://www.w3.org/TR/webgpu/#depth-stencil-attachments

Make depth_stencil_attachment more closely follow the specification.  No longer do we need to specify stencil_xxx if it's a depth, and no need to specify depth_xxx if it's a stencil.  We only need to specify "color" if it's a "clear".

### DIFF
--- a/wgpu/_classes.py
+++ b/wgpu/_classes.py
@@ -701,7 +701,7 @@ class GPUDevice(GPUObjectBase):
     """
 
     def __init__(self, label, internal, adapter, features, limits, queue):
-        super().__init__(label, internal, self)
+        super().__init__(label, internal, None)
 
         assert isinstance(adapter, GPUAdapter)
         assert isinstance(features, set)
--- a/wgpu/_classes.py
+++ b/wgpu/_classes.py
@@ -701,7 +701,7 @@ class GPUDevice(GPUObjectBase):
     """
 
     def __init__(self, label, internal, adapter, features, limits, queue):
-        super().__init__(label, internal, None)
+        super().__init__(label, internal, self)
 
         assert isinstance(adapter, GPUAdapter)
         assert isinstance(features, set)
--- a/wgpu/_classes.py
+++ b/wgpu/_classes.py
@@ -73,8 +73,8 @@ apidiff = ApiDiff()
 object_tracker = diagnostics.object_counts.tracker
 
 # The 'optional' value is used as the default value for optional arguments in the following two cases:
-# * The method accepts a descriptior that is optional, so we make all arguments (i.e. descriptor fields) optional, and this one does not have a default value.
-# * In wgpu-py we decided that this argument should be optonal, even though it's currently not according to the WebGPU spec.
+# * The method accepts a descriptor that is optional, so we make all arguments (i.e. descriptor fields) optional, and this one does not have a default value.
+# * In wgpu-py we decided that this argument should be optional, even though it's currently not according to the WebGPU spec.
 optional = None
 
 
@@ -618,7 +618,7 @@ class GPUAdapter:
         """Request a `GPUDevice` from the adapter.
 
         Arguments:
-            label (str): A human readable label. Optional.
+            label (str): A human-readable label. Optional.
             required_features (list of str): the features (extensions) that you need. Default [].
             required_limits (dict): the various limits that you need. Default {}.
             default_queue (structs.QueueDescriptor): Descriptor for the default queue. Optional.
@@ -800,7 +800,7 @@ class GPUDevice(GPUObjectBase):
         """Create a `GPUBuffer` object.
 
         Arguments:
-            label (str): A human readable label. Optional.
+            label (str): A human-readable label. Optional.
             size (int): The size of the buffer in bytes.
             usage (flags.BufferUsage): The ways in which this buffer will be used.
             mapped_at_creation (bool): Whether the buffer is initially mapped.
@@ -815,7 +815,7 @@ class GPUDevice(GPUObjectBase):
         writes the given data to it, and then unmaps the buffer.
 
         Arguments:
-            label (str): A human readable label. Optional.
+            label (str): A human-readable label. Optional.
             data: Any object supporting the Python buffer protocol (this
                 includes bytes, bytearray, ctypes arrays, numpy arrays, etc.).
             usage (flags.BufferUsage): The ways in which this buffer will be used.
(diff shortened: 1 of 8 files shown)


wgpu/backends/wgpu_native/_api.py
```diff
@@ -2735,6 +2724,66 @@ def begin_render_pass(
...
+        depth_clear_value = stencil_clear_value = 0
+        depth_keys_okay = stencil_keys_okay = False
+        # All depth texture formats have "depth" in their name.
+        if "depth" in view.texture.format:
+            depth_read_only = ds_attachment.get("depth_read_only", False)
+            if not depth_read_only:
+                depth_keys_okay = True
+                depth_load_op = ds_attachment["depth_load_op"]
+                depth_store_op = ds_attachment["depth_store_op"]
+                if depth_load_op == "clear":
```
fyellin, line 2742: Is there a better way of doing this?  Will anyone specify "clear" by its integer value?

wgpu/backends/wgpu_native/_api.py
```diff
@@ -2735,6 +2724,66 @@ def begin_render_pass(
...
+            *(("depth_load_op", "depth_store_op") if not depth_keys_okay else ()),
+            *(("stencil_load_op", "stencil_store_op") if not stencil_keys_okay else ()),
+        ]
+        for key in unexpected_keys:
+            if key in ds_attachment:
+                warnings.warn(
+                    f"Unexpected key {key} in depth_stencil_attachment",
+                    DeprecationWarning,
+                    stacklevel=2,
+                )
```
almarklein, line 2770: We use `logger.warning()` instead of the warnings module.

Another thing is that (I think) this warning would be shown on each draw, which may be a bit much. Perhaps we can do something like:
```py
if not getattr(self._device, "_showed_depth_stencilwarning1", False):
    self._device._showed_depth_stencilwarning1 = True
    logger.warning(....)
```

almarklein (commented):
Thanks for this!

fyellin:
Fixed some typos.
Discovered that some GPU objects had GPUDevice in the _device field while others had GPUDeviceImpl.  Made sure that everything uniformly had a GPUDevice, and that we test for this.

wgpu/backends/wgpu_native/_api.py
```diff
@@ -2731,10 +2719,69 @@ def begin_render_pass(
 
         # H: WGPURenderPassEncoder f(WGPUCommandEncoder commandEncoder, WGPURenderPassDescriptor const * descriptor)
         raw_encoder = libf.wgpuCommandEncoderBeginRenderPass(self._internal, struct)
-        encoder = GPURenderPassEncoder(label, raw_encoder, self)
+        encoder = GPURenderPassEncoder(label, raw_encoder, self._device)
         encoder._objects_to_keep_alive = objects_to_keep_alive
         return encoder
 
+    # Pulled out from create_render_pass because it was too large.
+    def _create_c_render_pass_depth_stencil_attachment(self, ds_attachment):
```
fyellin, line 2727: FYI.  I really wanted to put _create_render_pass_color_attachment() into a separate function.  However apipatcher.py::StructValidationChecker complained that it couldn't find `check_struct("Color", clear_value)` in `begin_render_pass()`.  

I could only get away with refactoring because the depth_stencil_attachment has no internal structures.

I wish there were a way that the validation checker could more easily support refactoring.

fyellin:
This PR has suddenly turned into a hodgepodge of three separate items:
      Fixing arguments for stencil
      Fixing the _device field so that it alway contains a device
      Random typos and grammar nits identified by PyCharm
I'm about to separate this into three PRs.   This will be just the first. 

Please don't bother reviewing until I give the go-ahead.

fyellin:
Deleted last comment.  More work than I realized for too little gain.

wgpu/_classes.py
```diff
@@ -701,7 +701,7 @@ class GPUDevice(GPUObjectBase):
     """
 
     def __init__(self, label, internal, adapter, features, limits, queue):
-        super().__init__(label, internal, None)
+        super().__init__(label, internal, self)
```
almarklein, line 704: I like the added test to make sure all objects have a proper `_device`  attribute, but for the device itself, this'd cause a circular reference, making it harder for the gc to clean up.

wgpu/backends/wgpu_native/_api.py
almarklein: I recently ran into a similar problem with the sync/async versions of a method to both call e.g. `_request_device()`. For these particular cases I could easily implement a fix/hack to see that the call was delegated to a method of the same name, but prefixed with an underscore.

Providing a mechanism to allow delegation mid-function would be rather complex. A quick fix could be would be to tell the check checked that its ok, by adding a commented version of the call, with a note to other humans to tell them that method X must do that check.

wgpu/_classes.py
almarklein: mmm, good point ;)

fyellin: Given that device already has a loop and we haven't noticed any problems, I think it would be cleaner if device._device = self.  

I'm happy to undo this last change if you agree.  Either way, I'll abide by your decision.

hmaarrfk: whatever happened to "weak references" i never learned how to use them....

hmaarrfk: but i always thought they were meant for these usecases.

fyellin: For a language like Swift that only has reference counting, this would be a perfect use case for a weak reference. In a language like Python which already has a cyclic garbage collector, it's less necessary.  A live reference to the queue will keep the device from being GCed, and a live reference to the device will keep the queue from being GC'ed. When there's no live reference to either of them, both will be collected when the GC runs.

If it were absolutely critical that devices be GCed as quickly as possible, weak references might make sense.  But no one even noticed until I pointed it out. In fifteen years of Python programming, I've hardly ever seen weak references used.

So my comment again to @almarklein. If you think that device._device should be the device, I'll bring back that code.  If you think it should stay as it is now, close this conversation, approve, and merge.

almarklein: I agree. Weak references have value when you want something to have access to a certain object, without it preventing that object to clean up. They feel like overkill in a situation like this. Cyclic references are still cleaned up as long as both are unreachable. It is my understanding that it only can take a bit longer for that to happen.

The device-object is a pretty long-lived object, so the speed of collection does not matter much, and trying to remove both cyclic references will make the code more complex, which is not worth it.

So yeah, let's move on :)

almarklein: >  I think it would be cleaner if device._device = self.

Yes, let's revert the last change.

almarklein (approved):

Korijn (approved):

PR #612, opened 2024-10-04
Add example that uses PySide6 with asyncio compat

### LINKED ISSUE #383: PySide 6.6 asyncio support

almarklein:
Closes #383

The example demonstrates how to use Qt to enable asyncio support.

### DIFF
--- a/examples/gui_qt_asyncio.py
+++ b/examples/gui_qt_asyncio.py
@@ -22,6 +22,24 @@ from wgpu.gui.qt import WgpuWidget
 from triangle import setup_drawing_sync
 
 
+def async_connect(signal, async_function):
+    # Unfortunately, the signal.connect() methods don't detect
+    # coroutine functions, so we have to wrap it in a function that creates
+    # a Future for the coroutine (which will then run in the current event loop).
+    #
+    # The docs on QtAsyncio do something like
+    #
+    #     self.button.clicked.connect(
+    #         lambda: asyncio.ensure_future(self.whenButtonClicked()
+    #     )
+    #
+    # But that's ugly, so we create a little convenience function
+    def proxy():
+        return asyncio.ensure_future(async_function())
+
+    signal.connect(proxy)
+
+
 class ExampleWidget(QtWidgets.QWidget):
     def __init__(self):
         super().__init__()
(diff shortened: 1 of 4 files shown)


examples/gui_qt_asyncio.py
```diff
@@ -0,0 +1,74 @@
...
+        self.button = QtWidgets.QPushButton("Hello world", self)
+        self.canvas = WgpuWidget(splitter)
+        self.output = QtWidgets.QTextEdit(splitter)
+
+        # With QtAsyncio, the callbacks can now return a future. You'd
+        # think that you could also return a coro, but we need to wrap
+        # it into a future, making this code a bit ugly.
+        # self.button.clicked.connect(self.whenButtonClicked)  # why not :/
+        self.button.clicked.connect(
+            lambda: asyncio.ensure_future(self.whenButtonClicked())
```
Korijn, line 42: What?? This looks like a mistake.

almarklein: Nope, this is the actual code they use in the example in the [introductory blog post](https://www.qt.io/blog/introducing-qtasyncio-in-technical-preview):

image

almarklein: It's silly, because they probably use `asyncio.isfuture()` somewhere, and they could also check `asyncio.iscoroutine()` ...

almarklein: Ah, I think I get it. The `ensure_future()`  actually schedules the coro to run in whatever is the current loop.

Still would be way more natural to be able to pass a coro-function, and Qt scheduling it to run. But maybe that requires changes in Qt itself. The whole QtAsyncio is pure Python.

Korijn: Seems like they're missing a `.connect_async` method that handles the boilerplate. :)

almarklein:
For reference, the QtAsyncio source: https://github.com/qtproject/pyside-pyside-setup/tree/dev/sources/pyside6/PySide6/QtAsyncio

examples/gui_qt_asyncio.py
almarklein: Refactored a little to make this code less ugly.

examples/gui_qt_asyncio.py
```diff
@@ -34,13 +52,8 @@ def __init__(self):
...
 
-        # With QtAsyncio, the callbacks can now return a future. You'd
-        # think that you could also return a coro, but we need to wrap
-        # it into a future, making this code a bit ugly.
-        # self.button.clicked.connect(self.whenButtonClicked)  # why not :/
-        self.button.clicked.connect(
-            lambda: asyncio.ensure_future(self.whenButtonClicked())
-        )
+        # self.button.clicked.connect(self.whenButtonClicked)  # see above :(
+        async_connect(self.button.clicked, self.whenButtonClicked)
```
Korijn, line 56: Just wondering how this handles arguments for the event handler?

almarklein: As it is implemente now, it is assumed there are no args. But since this is just an example, that seems fine, I think.

Korijn (approved):

PR #613, opened 2024-10-06
Nitpicky changes flagged by PyCharm

fyellin:
Typos and grammar mistakes found by PyCharm. 

With one exception ("name" was used as a loop variable within another loop that also used "name" as a loop variable), there are no changes to code.  

I only took those changes that I thought were really typos or bad grammar.  PyCharm doesn't know how to use a/an before acronyms, and wants more commas than I do.


Korijn (approved):

fyellin:
Reminder that I don't have write access.

PR #615, opened 2024-10-10
Structure checking

fyellin:
First pass at creating apipatcher that understands that check_structs() can be in helper code.  This code uses AST.parse
to parse the file and find all calls to helper functions and all calls to check_struct().

This may be overkill.

This PR is built on top of #614 and shouldn't be merged until it is.


fyellin:
Hoping to get some sort of feedback as to whether the change to apipatcher.py is overkill or useful.  I think that _api.py is desperately in need of refactoring.

almarklein:
Yes, from what I see in this PR, the amount of extra code is not that much, and it allows more refactoring. So IMO its worth it.

codegen/apipatcher.py
```diff
@@ -674,3 +675,41 @@ def _get_sub_structs(self, idl, structname):
...
+        structure_checks = defaultdict(list)
+
+        for key, method_ast in top_level_methods.items():
+            for node in ast.walk(method_ast):
+                if isinstance(node, ast.Call):
+                    name = ast.unparse(node.func)
+                    if name.startswith("self._"):
+                        method_helper_calls[key].append(name[5:])
+                    if name == "check_struct":
+                        if isinstance(node.args[0], ast.Constant):
```
almarklein, line 704: What would the `else` represent? From what I can see, `check_struct()` is always used with a string as first arg, so maybe we can assert this here?

codegen/apipatcher.py
```diff
@@ -674,3 +675,41 @@ def _get_sub_structs(self, idl, structname):
...
+        module = ast.parse(file_cache.read("backends/wgpu_native/_api.py"))
+        # We only care about top-level classes and their top-level methods.
+        top_level_methods = {
+            # (class_name, method_name) -> method_ast
+            (class_ast.name, method_ast.name): method_ast
+            for class_ast in module.body
+            if isinstance(class_ast, ast.ClassDef)
+            for method_ast in class_ast.body
+            if isinstance(method_ast, (ast.FunctionDef, ast.AsyncFunctionDef))
+        }
```
almarklein, line 690: Awesome dict comprehension 🚀

wgpu/backends/wgpu_native/_api.py
```diff
@@ -2173,16 +2180,15 @@ def read_mapped(self, buffer_offset=None, size=None, *, copy=True):
         src_m = get_memoryview_from_address(src_address, size)
 
         if copy:
-            # Copy the data. The memoryview created above becomes invalid when the buffer
-            # is unmapped, so we don't want to pass that memory to the user.
-            data = memoryview((ctypes.c_uint8 * size)()).cast("B")
-            data[:] = src_m
+            data = memoryview(bytearray(src_m)).cast("B")
```
almarklein, line 2183: Since data copies should be avoided where possible, but are needed sometimes, I like to be very explicit in comments on what happens and why, even if it may feel obvious sometimes.

The diff for this suggestion looks weird; I only added a comment.
```suggestion
            # Copy the data. The memoryview created above becomes invalid when the buffer is unmapped
            # bytearray() creates a copy, memoryview() creates a view on the byte-array.
            data = memoryview(bytearray(src_m)).cast("B")
```

wgpu/backends/wgpu_native/_api.py
```diff
@@ -2173,16 +2180,15 @@ def read_mapped(self, buffer_offset=None, size=None, *, copy=True):
...
-            # Return view on the actually mapped data
-            data = src_m
-            if hasattr(data, "toreadonly"):  # Py 3.8+
-                data = data.toreadonly()
+            # Return view on the actually mapped data.
+            try:
+                data = src_m.toreadonly()
+            except AttributeError:
+                # support for ≤ 3.7.  Do we still care??
+                data = src_m
```
almarklein, line 2191: No, we don't :)
```suggestion
            data = src_m.toreadonly()
```

almarklein (commented):
Nice work! Just a few comments.

codegen/apipatcher.py
fyellin: I try to be judicious in my use of "assert".  I prefer not to use it unless it really means a programmer error.  But I think you're right here.

fyellin:
Changes made.  Added a couple of more comments.

almarklein (approved):

Korijn (approved):

PR #617, opened 2024-10-14
Fix snake case for 1d/2d/3d suffix

almarklein:
In responding to https://github.com/pygfx/pygfx/issues/863, I realized that `
gfx.renderers.wgpu.set_wgpu_limits(max_texture_dimension_3d=...)`  does not work, but `
gfx.renderers.wgpu.set_wgpu_limits(max_texture_dimension3d=...)` does. While in fact wgpu uses the former notation, as is shown in the error message:
Caused by:
  Limit 'max_texture_dimension_3d' value 3000 is better than allowed 2048
```


This PR fixes that, and keeps the old notation working too: `max_texture_dimension3d` still resolves to `maxTextureDimension3D`. 

We have two implementations of `to_snake_case()` and `to_camel_case()`, because the codegen does not import from wgpu. This PR also makes them the same again (the diverged a little.

The codegen is actually not affected. Only the way that custom limits and such are handled.

### DIFF
--- a/codegen/tests/test_codegen_utils.py
+++ b/codegen/tests/test_codegen_utils.py
@@ -22,7 +22,8 @@ def test_to_snake_case():
     assert to_snake_case("_foo_bar_spam") == "_foo_bar_spam"
     assert to_snake_case("fooBarSpam") == "foo_bar_spam"
     assert to_snake_case("_fooBarSpam") == "_foo_bar_spam"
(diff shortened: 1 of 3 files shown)


almarklein:
Locally the failing test works. I'll wait for the nice diff error in #614.

Korijn (approved):

PR #618, opened 2024-10-14
Refactor wgpu.gui for scheduling

### LINKED ISSUE #355: Sync events with draws

### LINKED ISSUE #471: Blocking render call
Make it possible to render a new frame synchronously; in other words, a public API to call `canvas._draw_frame_and_present` directly. Do we have this?

Relates to https://github.com/pygfx/pygfx/discussions/667

almarklein:
## Refs

* This finally addresses the main points in wgpu for https://github.com/pygfx/pygfx/issues/495.
* Closes #355 
* closes #471

### Canvas

* Remove `WgpuAutoGui`, since its no simply a part of `WgpuCanvasBase`
* Implementation of `WgpuCanvasBase` is cleaner because it moves events and scheduling out.

### Events

* Event handling logic is moved into a separate class. 
* The event handling logic can merge events by itself (Canvas implementations don't have to do this).
* Events are queued until they are explicitly flushed (by the scheduling mechanics).
* EventType enum!

### Scheduling

* The `WgpuCanvasBase` implements a scheduling mechanic (Canvas implementations no longer have to do this).
* Multiple update modes are implemented: fastest, continuous, ondemand, manual.
* The `max_fps` applies (only) to continuous and ondemand mode.
* New `force_draw()` method for a blocking render call. 

### Public loop API

* `run()` and `call_later()` functions are replaced with a `loop` object that has both as a method.
* The `call_later` method returns a timer object, so the callback can be cancelled.
* Also added `call_repeated()` that returns a timer object that keeps ticking. We *really* needed an abstract timer object for the scheduling. Might as well provide it to our users.

## Design decisions

* We assume that users use the event-loop corresponding to the canvas, and don't use two GUI systems at the same time (e.g. Qt with glfw). We *can* accommodate for this, but it will complicate the code and feels like a bad idea in general.
* Similarly, our glwf window could run on any event loop, but we restrict it to asyncio (at least for now).
* Do we keep `canvas.add_event_handler()` or do we switch to `canvas.events.add_handler()` or `canvas.events.connect()`?
  * I think `canvas.add_event_handler()` feels a bit long, but it is explicit, and helps the canvas feel flat/simple.
* How to allow users to control scheduling, like setting mode or max_fps?
  * Proposal: use `canvas.scheduler.max_fps`,  `canvas.scheduler.mode`, etc.

## API changes

* New `canvas.force_draw()`.
* Replaced `from wgpu.gui.xx import WgpuCanvas, run` with `from wgpu.gui.xx import WgpuCanvas, loop`
* The glfw canvas is more friendly on your laptop battery (thanks to a proper event loop).
* The `WgpuAutoGui` mixin class is removed.

## Tasks

* [x] Update base canvas.
* [x] Take threading into account.
* [x] Update glfw gui
* [x] Update qt gui
* [x] Update wx gui
* [x] Update jupyter gui
* [x] Update offscreen gui
* [x] Don't draw when minimized.
* [x] Update docs 
* [x] Show/test force_draw in an example.
* [x] Can we make 'ondemand' with calling `request_draw()` in the draw function equal 'continuous'. -> we test that this is the case.

## After this PR

* Handle sigint.
* Animate events.
* Allow `draw_func` to be an asyn function.
* Can we move `present()` out of `CanvasContext()`?
* Register the draw function as a draw event handler instead of with `request_draw()`?
* Allow requesting a draw by setting a field in the event object?
* Tracking statistics (track time spent on different things). Plus visualizing them.


almarklein:
I've been working on this for more than a week. This is hard, but it's beginning to take shape now.

almarklein:
This implements (amongst other things) the ideas presented in http://gameprogrammingpatterns.com/game-loop.html

As robert says:

image

In addition to that, we also gave multiple different GUI systems and event loops.

> Do you own the game loop, or does the platform?

In our case that actually depends 😅  The scheduling code assumes it runs on an externally provided eventloop. For glfw we do provide the loop, but the scheduling logic does not know this.

Korijn:
> How to allow users to control scheduling, like setting mode or max_fps?
> * Proposal: use `canvas.scheduler.max_fps`,  `canvas.scheduler.mode`, etc.

Does scheduler even need to be a prop on canvas? Can it live on its own?

almarklein:
> Does scheduler need to be a prop on canvas? Can it live on its own?

Yeah, I'd rather hide it. So you mean the prop can be something like `canvas.update_mode`?

Korijn:
> > Does scheduler need to be a prop on canvas? Can it live on its own?
> 
> Yeah, I'd rather hide it. So you mean the prop can be something like `canvas.update_mode`?

I meant, it could be just `scheduler`, instead of `canvas.scheduler` :)

almarklein:
> > > Does scheduler need to be a prop on canvas? Can it live on its own?
> > 
> > 
> > Yeah, I'd rather hide it. So you mean the prop can be something like `canvas.update_mode`?
> 
> I meant, it could be just `scheduler`, instead of `canvas.scheduler` :)

There's no point for a scheduler without a canvas. It's really a little helper that I preferably hide from the user.

wgpu/gui/base.py
```diff
@@ -130,73 +133,182 @@ def __del__(self):
...
+
+        # Cannot draw to a closed canvas.
+        if self.is_closed():
+            return
+
+        # Process special events
+        # Note that we must not process normal events here, since these can do stuff\
+        # with the canvas (resize/close/etc) and most GUI systems don't like that.
+        self._events.submit({"event_type": "before_draw"})
+        self._events.flush()
```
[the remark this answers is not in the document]
Korijn: Well... all is well that ends well. Good find!

almarklein:
> Proposal: use canvas.scheduler.max_fps, canvas.scheduler.mode, etc.

These args can be passed when the canvas is instantiated. For now, we can leave it at that (no changing these at runtime).

almarklein:
Ok, this is almost done. All backends are working and tested on different platforms. I'm going to do a round of self-review and cleanup.

I think we will never merge this here, but in the new repo: #627

wgpu/gui/jupyter.py
```diff
@@ -46,21 +44,29 @@ def handle_event(self, event):
...
         # The result is either a numpy array or None, and this matches
         # with what this method is expected to return.
+
+        # When we had to wait relatively long for the drawn to be made,
+        # we do another round processing events, to minimize the perceived lag.
+        # We only do this when the delay is significant, so that under good
+        # circumstances, the scheduling behaves the same as for other canvases.
+        if time.perf_counter() - self._draw_request_time > 0.02:
+            self._process_events()
+
```
almarklein, line 61: cc @kushalkolar

wgpu/backends/wgpu_native/_api.py
```diff
@@ -519,6 +519,8 @@ class GPUCanvasContext(classes.GPUCanvasContext):
     # we can give meaningful errors/warnings on invalid use, rather than
     # the more cryptic Rust panics.
 
+    _surface_id = ffi.NULL
+
```
almarklein, line 523: This changes should be applied to wgpu

almarklein:
This is ready for review. It does not actually break compatibility so much. You can run pygfx examples with it.

When this pr is approved, I'll re-recreate the pr at `rendercanvas` so we can apply it there.

wgpu/backends/wgpu_native/_api.py
almarklein: #635

almarklein:
Merged in rendercanvas.

PR #622, opened 2024-10-17
Fix install of mesa drivers

almarklein:
Ref https://github.com/fastplotlib/fastplotlib/pull/651


kushalkolar:
github has [decided to rollback latest from 24.04 to 22.04 for now](https://github.com/actions/runner-images/issues/10636#issuecomment-2417303444), but this package seems to cover both 22.04 and 24.04.

almarklein:
And we don't render on RTD (yet), so there actually was not a problem in this repo. But there will be at some point, so let's just fix it now.

Korijn (approved):

PR #626, opened 2024-10-21
If data in create_buffer_with_data isn't a multiple of 4, just round up.

### LINKED ISSUE #512: create_buffer_with_data requires data to be a multiple of COPY_BUFFER_ALIGNMENT
The helper function `create_buffer_with_data` gives an error if the data it passes is not a multiple of 4.

Here is a reasonable use case for why I would want to do that:

I have an external model that includes vertices, normals, and indices of triangles.  These models are often written by others
```
     vertices_array = np.array(model.vertices, dtype=np.float32)
     vertices_buffer = device.create_buffer_with_data(vertices_array, usage=BufferUsage.VERTEX)
      . . .
     indices_array = np.array(model.indices, dtype=uint16)
     indices_buffer = device.create_buffer_with_data(indices_array, usage=BufferUsage.INDEX)
      . . . 
     encoder.set_index_buffer(indices_buffer, "uint16")
     encoder.draw_index(len(indices_array))
```
This works fine, unless the model happens to have an odd number of indexes. In that case, the byte size of `indices_array` will not be a multiple of 4, and the code fails.

There are solutions around the problem, but they are not always pretty:

1.  I can test the length of `model.indices` and add a junk element to the end if it is odd.  But note that this information belongs to someone else, and may be readonly.
2. I can just avoid using `create_buffer_with_data`, but perform the mapping myself.

But since `create_buffer_with_data` is a convenience function and not subject to a WebGPU specification, I think it's reasonable for it to always round up the size to a multiple of 4.  This could either (1) always be done, since it is invisible to the user or (2) be done if the user specifies `allows_padding` or something like that, indicating that they are aware of the problem.

fyellin:
FIXES #512 

If size of data isn't a multiple of 4, just round up.
Got rid of size argument to map_write, which no one was using.  Allow data to not be a multiple of 4 bytes.
Fixed a couple of typos, simplified some tests, added more tests.

The fix is essentially that used in the C WGPU API.

### DIFF
--- a/wgpu/_classes.py
+++ b/wgpu/_classes.py
@@ -810,6 +810,9 @@ class GPUDevice(GPUObjectBase):
             size (int): The size of the buffer in bytes.
             usage (flags.BufferUsage): The ways in which this buffer will be used.
             mapped_at_creation (bool): Whether the buffer is initially mapped.
+
+        Alignment: the size must be a multiple of 4.
+
         """
         raise NotImplementedError()
 
@@ -1448,9 +1451,6 @@ class GPUBuffer(GPUObjectBase):
             buffer_offset (int): the buffer offset in bytes. Must be at least
                 as large as the offset specified in ``map()``. The default
                 is the offset of the mapped range.
-            size (int): the size to write (in bytes). The default is the size of
-                the data, so this argument can typically be ignored. The
-                resulting range must fit into the range specified in ``map()``.
 
         Alignment: the buffer offset must be a multiple of 8.
 
--- a/wgpu/_classes.py
+++ b/wgpu/_classes.py
@@ -820,7 +820,8 @@ class GPUDevice(GPUObjectBase):
                 includes bytes, bytearray, ctypes arrays, numpy arrays, etc.).
             usage (flags.BufferUsage): The ways in which this buffer will be used.
 
-        Alignment: the size (in bytes) of the data must be a multiple of 4.
+        Alignment: if the size (in bytes) of data is not a multiple of 4, the buffer
+                   size is rounded up to the nearest multiple of 4.
 
         Also see `GPUBuffer.write_mapped()` and `GPUQueue.write_buffer()`.
         """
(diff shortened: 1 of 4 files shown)


wgpu/_classes.py
```diff
@@ -1425,7 +1426,7 @@ def read_mapped(self, buffer_offset=None, size=None, *, copy=True):
         raise NotImplementedError()
 
     @apidiff.add("Replacement for get_mapped_range")
-    def write_mapped(self, data, buffer_offset=None, size=None):
+    def write_mapped(self, data, buffer_offset=None):
```
almarklein, line 1429: Yeah, we can do this, because this method is not in the WebGPU spec. But we must also remove the arg from the docs below.

wgpu/_classes.py
```diff
@@ -820,7 +820,8 @@ def create_buffer_with_data(self, *, label="", data, usage: "flags.BufferUsage")
                 includes bytes, bytearray, ctypes arrays, numpy arrays, etc.).
             usage (flags.BufferUsage): The ways in which this buffer will be used.
 
-        Alignment: the size (in bytes) of the data must be a multiple of 4.
+        Alignment: if the size (in bytes) of data is not a multiple of 4, the buffer
+                   size is rounded up to the nearest multiple of 4.
```
almarklein, line 824: I see that the info on alignment is missing in the docstring for `create_buffer()`. Let's add that while we're at it.

almarklein (commented):
Nice! Just two minor comments.

wgpu/_classes.py
fyellin: Fixed

wgpu/_classes.py
fyellin: Thanks for catching that. Yes, I realized that this was our function and that the size arg made no sense.

fyellin:
In addition to addressing your comments, I also added a test.

Korijn:
Sorry for pushing that merge commit, I wanted to get this in and didn't notice you were still working on it.

fyellin:
All done now.

fyellin:
Something screwy is going on.  I'm not seeing my most recent changes here.  Let me see if I can get everything cleared up.  Don't do anything until I give the okay.

fyellin:
Everything is okay.

Korijn (approved):

PR #628, opened 2024-10-21
Only keep alive those objects that are necessary.

fyellin:
As of wgpu 1.25, we no longer need to keep alive as many objects internally as we used to.  I have deleted several pieces of code that kept a copy of its arguments to that they wouldn't be released.  

The only exception is RenderBundleEncoder.  I have filed a [bug](https://github.com/gfx-rs/wgpu/issues/6433) about it.  

Added a test that agressively deletes objects even though they are needed by the GPU, just to make sure that things work correctly.


Korijn (approved):
Very cool!!

PR #629, opened 2024-10-22
Add pre-commit hooks for ruff

claydugo:
I did this with their recommended setup where you specify a revision. 

You can alternatively set it up to use the users `ruff` binary with something like:

```
repos:
  - repo: local
    hooks:
      - id: ruff
        name: ruff
        entry: ruff check .
        language: system
```

This is opt-in and requires running `pre-commit install`. Otherwise the current manual workflow still applies.

### DIFF
--- a/README.md
+++ b/README.md
@@ -120,8 +120,9 @@ This code is distributed under the 2-clause BSD license.
   binaries.
   * You can use `python tools/download_wgpu_native.py` when needed.
   * Or point the `WGPU_LIB_PATH` environment variable to a custom build of `wgpu-native`.
-* Install the `pre-commit` hooks with `pre-commit install`.
-* Linting and formatting will be automatically applied on `git commit` but you can manually verify compliance with `pre-commit run --all-files`.
+* Use `ruff format` to apply autoformatting.
+* Use `ruff check` to check for linting errors.
+* Optionally, if you install `pre-commit` hooks with `pre-commit install`, lint fixes and formatting will be automatically applied on `git commit`.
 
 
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -120,8 +120,8 @@ This code is distributed under the 2-clause BSD license.
   binaries.
   * You can use `python tools/download_wgpu_native.py` when needed.
   * Or point the `WGPU_LIB_PATH` environment variable to a custom build of `wgpu-native`.
-* Use `ruff format` to apply autoformatting.
-* Use `ruff check` to check for linting errors.
+* Install the `pre-commit` hooks with `pre-commit install`.
+* Linting and formatting will be automatically applied on `git commit` but you can manually verify compliance with `pre-commit run --all-files`.
```
almarklein, line 124: Since this is opt-in, let's leave the original tips on using `ruff`, and add how to install the pre-commit hook to do formatting and linting automatically on each commit.

almarklein (approved):

Korijn (approved):

fyellin:
It's probably too late.  But I just saw the README.md, and I have absolutely no idea what it means.  

> Optionally, if you install `pre-commit` hooks with `pre-commit install`, lint fixes and formatting will be automatically applied on `git commit`.

Is `pre-commit install` a command I'm supposed to type at the shell? (It doesn't work).  Is it a git command?  (There is no such git command). From searching the web, it appears that this is a Python package that I'm supposed to install.  We shouldn't make readers of READER.md have to search the web. Maybe pre-commit should be part of the `pip -e .[dev].` ?

PR #631, opened 2024-11-01
Clean up async code

fyellin:
Asynchronous code in _api needed to know too much about the internal workings of the AwaitHandler.  Changed it so that there's a little bit more separation. 

Also made some asynchronous functions really be asynchronous.

Added on_submitted_word_done. 

Commenting out implementation of create_xxx_pipeline_async because these are unimplemented in wgpu-native.  The code is still there, but a constant forces the synchronous implementation.

Added tests.  Added dependency on pytest-asyncio to run those tests.


wgpu/backends/wgpu_native/_helpers.py
```diff
@@ -230,53 +232,62 @@ def to_camel_case(name):
...
+        self.event.set()
+
+    def set_error(self, error):
+        self.result = (None, error)
+        self.event.set()
+
+    def wait_sync(self):
+        if not self.poll_function:
+            # The result must come back without any polling.
+            assert self.event.is_set()
```
Korijn, line 259: In runtime code it's preferable to raise an exception over assertions

Korijn:
Very nice improvements!

fyellin:
While I'm cleaning up.  Why do both `enumerate_adapters_sync` and `enumerate_adapters_async` exist?   There seems to be no value added by the async version.  Can these just be merged into `enumerate_adapters"?

When I look at the [Rust documentation](https://docs.rs/wgpu/latest/wgpu/struct.Instance.html#method.enumerate_adapters), there seems to be only a single synchronous operation.

wgpu/backends/wgpu_native/_helpers.py
```diff
@@ -228,55 +230,69 @@ def to_camel_case(name):
...
+    def __init__(self, title, callback, finalizer, poll_function=None, timeout=5.0):
         self.title = title  # for context in error messages
-        self.result = result  # a dict that the callbacks writes to
         self.callback = callback  # only used to prevent it from being gc'd
-        self.poll_function = poll_function  # call this to poll wgpu
         self.finalizer = finalizer  # function to finish the result
-        self.maxtime = time.perf_counter() + float(timeout)
+        self.poll_function = poll_function  # call this to poll wgpu
+        self.timeout = timeout
+        self.event = asyncio.Event()
```
almarklein, line 248: There's a couple of problems with this implementation:

* It relies on `asyncio`, which basically closes the door for other aync frameworks. I really want to avoid that.
* AFAIK the event does not help, because it can only be set from the poll function.
* A sleep time of 0.1 s is a *realy* long time :)

tests/test_async.py
```diff
@@ -0,0 +1,106 @@
...
+    def poll_function():
+        nonlocal count
+        count += 1
+        if count >= 3:
+            callback(10)
+
+    awaitable = WgpuAwaitable("test", callback, finalizer, poll_function)
+
+    if use_async:
+        result = await awaitable.wait_async()
```
almarklein, line 31: I think we can make this work using `result = await awaitable`. That's how it worked earlier.

wgpu/backends/wgpu_native/_api.py
```diff
@@ -1079,6 +1067,8 @@ class GPUDevice(classes.GPUDevice, GPUObjectBase):
     # GPUObjectBaseMixin
     _release_function = libf.wgpuDeviceRelease
 
+    CREATE_PIPELINE_ASYNC_NOT_IMPLEMENTED = True
```
almarklein, line 1070: Let's make this private (prefix with an underscore).

almarklein (commented):
Nice work, I like the `set_error()` and `set_result()`. I think that crossed my mind when I worked on this, but I can't recall why I went with that weird result dict.

almarklein:
> Why do both enumerate_adapters_sync and enumerate_adapters_async exist?

Good question. The reason is that in order to implement this method in JS, we must probably just call `request_adapter` with different power settings, and `request_adapter` is async.

tests/test_async.py
fyellin: The problem is that I'm not sure *why* what we had before worked.  

According to [this documentation](https://peps.python.org/pep-0492/#await-expression), the value returned by `__await__` has to be an iterator.  Anything else is a TypeError.  

Likewise, according to [this stackoverflow answer](https://stackoverflow.com/questions/63964011/precise-specification-of-await), the value returned by `__await__` is library specific, and the value returned by trio might be different than the value returned by asyncio.  

But I can't argue with the fact that the code seems to work.  So I'll keep my changes to _api.py, but undo _helper.py to be a mixture of the old and new.  And I'll keep the tests.

New version in a couple of hours.

fyellin: Hmm.  As I feared, the current __await__ is asyncio only.  When I switched test-async to test both asyncio and trio, I got the error message:

```
TypeError: trio.run received unrecognized yield message None. Are you trying to use a library written for some other framework like asyncio? That won't work without some kind of compatibility shim.
```
This failed running with `trio` and calling `async awaitable` where `awaitable` has something similar to the current implementation of __await__ .

Korijn:
Just an open suggestion: maybe it would be helpful to leverage [anyio](https://anyio.readthedocs.io/en/stable/index.html) if our goal is to support both trio and asyncio? It has a pytest plugin too. It might make our lives simpler.

> Applications and libraries written against AnyIO’s API will run unmodified on either asyncio or trio

This quote from their docs seems very promising.

fyellin:
I was going to ask about anyio.  It seems like a reasonable solution, but I
didn’t want to bring in a third-party library without prior approval.

And pytest-anyio conveniently runs both asyncio and trio.

Ignore the push I just did. I’ll rewrite to use anyio.

almarklein:
> I was going to ask about anyio.

For testing It definitely makes sense. For implementation, only if it remains somewhat contained and an implementation detail.

There's a lot of changes coming in how wgpu-native implements futures, and I'm also not sure what we'd need to do to support this running in JS. My point is that this will code will evolve quite a bit in the coming months, and I don't want to lock into a dependency.

fyellin:
The Python documentation on writing implementation-agnostic asynchronous
code is very thin. I'm still investigating if there is an easy way of
saying in an async function "Let other coroutines run and then come back to
me" that doesn't tie the code down into a specific implementation.

This is one of the more painful areas of Python development.  It's a lot
easier if you know which async library you're going to use.

fyellin:
I need some advice on how to finish this.  I'm temporarily using anyio, which can deal with both trio and asyncio and it's mostly working.

The problem I'm running into is on the build.  I've included "anyio" in [pyproject.toml ](https://github.com/fyellin/wgpu-py/blob/async_handler/pyproject.toml#L12), and the build and tests run fine.  But when the wheel is being built, it doesn't include anyio.  How do I tell the wheel-builder to include this?

Just as an aside, I went to a clean environment and did "pip install wgpu".  I was surprised to see pycparser included.  I can't find where that dependency is coming from.

Korijn:
> The problem I'm running into is on the build. I've included "anyio" in [pyproject.toml ](https://github.com/fyellin/wgpu-py/blob/async_handler/pyproject.toml#L12), and the build and tests run fine. But when the wheel is being built, it doesn't include anyio. How do I tell the wheel-builder to include this?

Wheels don't ship their dependencies, they're only listed in their metadata files.

> Just as an aside, I went to a clean environment and did "pip install wgpu". I was surprised to see pycparser included. I can't find where that dependency is coming from.

I'm guessing it's a transient dependency of cffi.

almarklein:
> I'm guessing it's a transient dependency of cffi.

It is: https://github.com/python-cffi/cffi/blob/88f48d22484586d48079fc8780241dfdfa3379c8/pyproject.toml#L12-L14

almarklein:
It looks like `yield None` works for Trio too. The below runs two tasks simultaneously. One uses the awaitable that constantly switches back to the event-loop using `yield None`, the other prints the time on an interval.

```py
import time
import trio

class Awaitable:    
    def __await__(self):
        while True:
            yield None

async def main():
    async with trio.open_nursery() as nursery:
        nursery.start_soon(child1)
        nursery.start_soon(child2)

async def child1():
    await Awaitable()

async def child2():
    while True:
        print(time.time())
        trio.sleep(0.01)
    
trio.run(main)
```

almarklein:
To be honest I'm reluctant for using `anyio` because it brings in 4 [extra dependencies](https://github.com/agronholm/anyio/blob/4d3dd2697a8985216bad5538fbfa4be4c9ebcae3/pyproject.toml#L29-L34).

If we really do want to sleep, we can maybe use `sniffio` (which anyio also uses), which is a tiny self-contained library, and use that to detect asyncio vs trio and then use either trio.sleep or asyncio.sleep.

But my preference would be to use the `yield None` trick if that's possible, because its simple and because the code will have to change anyway when the new future-implementation in wgpu-native arrives. From what I can see `yield None` works on Trio too, but maybe I'm missing something.

fyellin:
Okay.  I'm now happy with this.

I verified that when wgpu is install normally (`pip install /tmp/builder/wgpu-py/dist/wgpu-0.19.0-py3-none-manylinux_2_28_aarch64.whl`) that `anyio` is installed.  It is only the special `pip` instructions used in `cd.yml` that were causing problems.

For now, this works with asyncio and trio.  I'm still looking for a way to say "I don't care which async package I'm using, let someone else run for X seconds."  Such a think doesn't seem to exist.  There really doesn't even seem to be a way of finding out which asynchronous package you're using.

Note that `test-async` uses `pytest.anyio` which causes it to test the code using both asyncio and trio.

almarklein:
>

almarklein:
>  There really doesn't even seem to be a way of finding out which asynchronous package you're using.

That's what [sniffio](https://github.com/python-trio/sniffio) tries to do.

wgpu/backends/wgpu_native/_helpers.py
fyellin: 1. Now the code uses anyio, which can support trio and asyncio.  The only asynchronous call it makes is to sleep.
2. Yeah. There is a possibility that someone really is writing asynchronous code and that a different coroutine has called "poll".  But you're right, it's not really worth the complications.
3. Changed to 5ms.

wgpu/backends/wgpu_native/_api.py
fyellin: Done

fyellin:
Everything resolved.

Korijn:
I will agree with Almar on the dependency management topic here.

If you switch from anyio to sniffio, you can write something like the following:

```py
def get_async_backend():
    lib = sniffio.current_async_library()
    if lib == "asyncio":
        import asyncio
        return asyncio
    # and so on
```

That would reduce our dependency footprint quite a lot.

wgpu/backends/wgpu_native/_helpers.py
```diff
@@ -234,49 +233,63 @@ class WgpuAwaitable:
     truely async manner, as well as to support a synchronous version of them.
     """
 
-    def __init__(self, title, result, callback, poll_function, finalizer, timeout=5.0):
+    SLEEP_TIME = 0.005  # 5ms
```
Korijn, line 236: This should really be 0, we want to resolve the awaitables as fast as possible

almarklein: I think so too. If a user uses `buffer.map()` on every draw, than this wait-time hits performance on each frame. There's no way to come up with a "reasonable" timeout, so I think 0 makes the most sense. That way at least control is given back to the loop / other threads.

almarklein:
Interesting that CI is green; if I checkout this branch and run an example, it panics in `_wgpuDeviceCreateRenderPipeline`. Note the `invalid vertex format for vertex attribute: 0`:

```
thread '<unnamed>' panicked at src/lib.rs:2169:42:
invalid vertex format for vertex attribute: 0
thread '<unnamed>' panicked at library/core/src/panicking.rs:221:5:
panic in a function that cannot unwind
stack backtrace:
   0:        0x1034a14e8 - <std::sys::backtrace::BacktraceLock::print::DisplayBacktrace as core::fmt::Display>::fmt::h243268f17d714c7f
   1:        0x1034ba19c - core::fmt::write::hb3cfb8a30e72d7ff
   2:        0x10349fb84 - std::io::Write::write_fmt::hfb2314975de9ecf1
   3:        0x1034a2504 - std::panicking::default_hook::{{closure}}::h14c7718ccf39d316
   4:        0x1034a2128 - std::panicking::default_hook::hc62e60da3be2f352
   5:        0x1034a2fc8 - std::panicking::rust_panic_with_hook::h09e8a656f11e82b2
   6:        0x1034a28f0 - std::panicking::begin_panic_handler::{{closure}}::h1230eb3cc91b241c
   7:        0x1034a1974 - std::sys::backtrace::__rust_end_short_backtrace::hc3491307aceda2c2
   8:        0x1034a25e0 - _rust_begin_unwind
   9:        0x1034df108 - core::panicking::panic_nounwind_fmt::h91ee161184879b56
  10:        0x1034df180 - core::panicking::panic_nounwind::heab7ebe7a6cd845c
  11:        0x1034df224 - core::panicking::panic_cannot_unwind::hedc43d82620205bf
  12:        0x1031ec070 - _wgpuDeviceCreateRenderPipeline
  13:        0x1a5051050 - <unknown>
  14:        0x1a5059b04 - <unknown>
  15:        0x101c51188 - _cdata_call
  16:        0x100ed20d0 - __PyObject_Call
  17:        0x100fc4358 - __PyEval_EvalFrameDefault
  18:        0x100eeb4b0 - _gen_send_ex2
  19:        0x100eeb114 - _gen_send_ex
  20:        0x100f20524 - _cfunction_vectorcall_O
  21:        0x100fe3124 - _context_run
  22:        0x100fc3e30 - __PyEval_EvalFrameDefault
  23:        0x100eeb4b0 - _gen_send_ex2
  24:        0x100eeb114 - _gen_send_ex
  25:        0x100fc3cbc - __PyEval_EvalFrameDefault
  26:        0x100fb942c - _PyEval_EvalCode
  27:        0x100fb6d44 - _builtin_exec
  28:        0x100f2024c - _cfunction_vectorcall_FASTCALL_KEYWORDS
  29:        0x100ed1e60 - _PyObject_Vectorcall
  30:        0x100fc2968 - __PyEval_EvalFrameDefault
  31:        0x100fb942c - _PyEval_EvalCode
  32:        0x101019e20 - _run_mod
  33:        0x1010183ec - __PyRun_SimpleFileObject
  34:        0x101017e20 - __PyRun_AnyFileObject
  35:        0x10103bb50 - _Py_RunMain
  36:        0x10103bf10 - _pymain_main
  37:        0x10103bfb0 - _Py_BytesMain
thread caused non-unwinding panic. aborting.
```

almarklein:
I was wondering how our trio example was able to run, while your new test indicated that trio does not handle the `yield None`. Turns out that in `main`, the `Awaitable.is_done()` calls the poll function, and it seems to always resolve on the first try, so the None is never yielded. Conclusion: the `yield None` is a asyncio-specific trick.

fyellin:
Is there something holding this up?

Korijn:
> Is there something holding this up?

Yes, these two comments:

* https://github.com/pygfx/wgpu-py/pull/631#discussion_r1833891369
* https://github.com/pygfx/wgpu-py/pull/631#issuecomment-2464021896

wgpu/backends/wgpu_native/_helpers.py
fyellin: I have changed it to 0.  Sorry I didn't see that comment.  Github usually lets me know.

fyellin:
I double checked "yield None" with the simplest possible code:

```
>>> class Foo:
...     def __await__(self):  yield None; return 10
... 
>>> async def foo():
...     await Foo()
... 
>>> import trio
>>> trio.run(foo)
  File "<stdin>", line 1, in <module>
  File "/Users/fy/Pycharm/wgpu-py/.venv/lib/python3.12/site-packages/trio/_core/_run.py", line 2407, in run
    raise runner.main_task_outcome.error
  File "<stdin>", line 2, in foo
  File "<stdin>", line 2, in __await__
TypeError: trio.run received unrecognized yield message None. Are you trying to use a library written for some other framework like asyncio? That won't work without some kind of compatibility shim.
>
```
Meanwhile `asyncio.run(foo())` worked just fine.  So the `yield None` is definitely asyncio specific.

fyellin:
@almarklein .  What are you doing to see your crash?   I haven't seen that on either my Mac or on the headless Linux machines I occasionally test on.

almarklein:
> What are you doing to see your crash?

`python examples/triangle.py` shows a black window. `python examples/cube.py` crashes.

fyellin:
@almarklein.  Hmm.  It works just fine on my Mac. . . .  Let me see if I can re-create this on a headless Linux device.  If not, I'll just close this PR.  Not worth the hassle it's causing.

fyellin:
@almarklein.  Can you give me details on OS, Python version, and GPU?  I rewrote cube.py to be windowless, ran it on a Linux machine, and it just worked.  The bug you've found sounds serious, but I just can't reproduce it. . .

almarklein:
I'm on MacOS (M1), python 3.12.4. This is very strange indeed. I'm having a quick look. (will update this comment)

almarklein:
I found and fixed the problem by adding a `new_array()` function. It makes sure that the substructs are associated with the array so they are not gc'd prematurely. It also returns null when the list of substructs is empty, which simplifies the code in a few other places (and produces a cleaner struct where we did not consider the case of the empty array). This change should also make further refactoring easier.

almarklein:
The last remaining issue is to replace `anyio` with `sniffio`.

wgpu/backends/wgpu_native/_helpers.py
```diff
@@ -234,49 +233,62 @@ class WgpuAwaitable:
...
+            if not self.result:
+                raise RuntimeError(f"Waiting for {self.title} timed out.")
+            result, error = self.result
+            if error:
+                raise RuntimeError(error)
+            else:
+                return self.finalizer(result)
+        finally:
+            # Reset attrs to prevent potential memory leaks
+            self.callback = self.finalizer = self.poll_function = self.result = None
```
almarklein, line 265: FYI I added this because these functions are likely closures, so this is somewhat prone to holding references. I think this is the reason why I used that `result_dict` earlier.

fyellin: I've spent my life programming in Lisp, Java and Python rather than C, C++, Rust, etc.  I tend to (foolishly?) trust the GC to do its job better than I can.  Unless it's a big object, I tend not to worry.. The callback has a reference to the awaitable and the awaitable has a reference to the callback, but the GC will take care of that. 

However if that bothers you, I understand.

almarklein:
For the record, I'll try to explain what was happening a bit better. When you create an ffi struct, it does not hold a reference to its fields when these fields are pointers. So if you do:

```py
sub_struct = ffi.new("CSubType")
struct = ffi.new("CType")
struct.aField = sub_struct
```

then `struct` does not hold a reference to `sub_struct`. If `sub_struct` goes out of scope, its memory is removed, and `struct` is now pointing to unclaimed memory; using it can result into garbage or a segfault.

We solved this problem long ago using the `new_struct()` and `new_struct_p()` functions in `_api.py`. One of the things these functions do is associated the fields with the struct, using a `WeakKeyDictionary`.

However, for arrays we had code like this in many places:
```py
list_of_sub_structs = [new_struct("CSubType", ...) for ... in ...]
array_of_sub_structs = ffi.new("CSubType[]", list_of_sub_structs)
struct.a_field = array_of_sub_structs
```

These arrays are contiguous copies of the list of structs. The problem is not that the sub-structs in the list get cleared by the gc (as I thought earlier), but that any sub-sub-structs and/or sub-sub-arrays of these sub-structs get cleared when the sub-struct gets cleared.

almarklein:
Wow, this PR is interesting, in being quite technical/deep on two very different topics 😅

wgpu/backends/wgpu_native/_helpers.py
almarklein: The gc can handle circular refs, yeah, but they do take longer for the gc to remove. 

I also trust the gc in most code, but most code in wgpu is potentially called a lot of times per second, so I'm more on my guard as usual 😉

almarklein:
I replaced `anyio` with `sniffio` and made sure it worked with with the changed in https://github.com/pygfx/rendercanvas/pull/41. This is ready from my end.

Korijn (approved):

PR #633, opened 2024-11-04
Merge enumerate_adapters_async and _sync into just enumerate_adapters.

fyellin:
Speculative PR.  Feel free to not approve if you think this is a bad idea.

There is no reason for enumerate_adapters_async.  In Rust, there is only a single function that returns Vec<Adapter>, and it is not asynchronous in any way.

### DIFF
--- a/docs/start.rst
+++ b/docs/start.rst
@@ -99,7 +99,7 @@ You can verify whether the `"DiscreteGPU"` adapters are found:
     import wgpu
     import pprint
 
-    for a in wgpu.gpu.enumerate_adapters_sync():
+    for a in wgpu.gpu.enumerate_adapters():
         pprint.pprint(a.info)
 
 If you are using a remote frame buffer via `jupyter-rfb <https://github.com/vispy/jupyter_rfb>`_ we also recommend installing the following for optimal performance:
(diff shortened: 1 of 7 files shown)


almarklein:
I'll just copy the comment I made in the other pr: The reason for the `_sync` and `_async` version of `request_adapters` is that in order to implement this method in JS, we must probably just call request_adapter with different power settings, and request_adapter is async.

almarklein:
Nevertheless, thanks for keeping an eye out, and help cleaning things up!

fyellin:
Before I delete this for good.  Just double checking.   This PR wasn't
about request_adapters.  They're staying just the way they are.

This was about enumerate_adapters, which is a wgpu extension and whose only
implementation in both Rust and in our code is synchronous and
callback-free.

almarklein:
Ah, I made a typo in my comment: I wrote `request_adapters` where I meant `enumerate_adapters` in one spot. I updated the comment.

The `enumerate_adapters` is actually in our public API (on the `gpu` object). It is a feature that is valuable for scientific applications (multi-GPU setups), and we can easily mimic it on WebGPU. The only catch is that the mimicing involves a call to `request_adapter`, so we'd therefore have to do the `_sync` / `_async` stuff here too. I hope this makes it more clear?

PR #639, opened 2024-11-05
Alwats use rtd theme

almarklein:
Not sure why it was disabled ...


almarklein:
Much better: https://wgpu-py--639.org.readthedocs.build/en/639/

PR #640, opened 2024-11-06
Combining more code

fyellin:
Several native-wgpu extensions apply to RenderPassEncoder and ComputePassEncoder, but not to RenderBundleEncoder.   Push Constants will soon join this group.  

Even though there is no Class that is the ancestor of just these two, we put the code into GPUBindingCommandsMixin.


fyellin:
Closing.  This accidentally got some code it wasn't supposed to.

PR #642, opened 2024-11-14
Refactor present-method mechanic

almarklein:
This does two things:

* Implement the changes in our `WgpuCanvasContext` to match with the new API in https://github.com/pygfx/rendercanvas/pull/21
  * This means that wgpu keeps working with rendercanvas (after the above change) 
* Backport the changes in our gui backends.
  * This means that `wgpu.gui` keeps working, so we can move to rendercanvas gradually.

### DIFF
--- a/docs/guide.rst
+++ b/docs/guide.rst
@@ -28,7 +28,7 @@ Next, we can setup the render context, which we will need later on.
 
 .. code-block:: py
 
-    present_context = canvas.get_context()
+    present_context = canvas.get_context("wgpu")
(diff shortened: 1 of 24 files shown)


almarklein:
I changed the scope of this PR. Before `rendercanvas`, the context-canvas interaction was an implementation detail. In https://github.com/pygfx/rendercanvas/pull/21 I propose an improved and documented version. This PR implements that.

Korijn (approved):

PR #651, opened 2024-12-09
Remove timeout in awaitable

### LINKED ISSUE #650: device.queue.read_buffer times out
### Describe the bug
`device.queue.read_buffer` times out

### To Reproduce
```
git clone https://github.com/tinygrad/tinygrad.git
cd tinygrad
python3 -m pip install -e .
PYTHONPATH=. WEBGPU=1 python3 examples/stable_diffusion.py --seed 0 --steps 1
```

Output:

>   File "/Users/ahmedharmouche/Documents/tinygrad/tinygrad/runtime/ops_webgpu.py", line 56, in _copyout
>     buffer_data = self.dev.queue.read_buffer(src, 0)
>                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
>   File "/Users/ahmedharmouche/Documents/tinygrad/venv/lib/python3.12/site-packages/wgpu/backends/wgpu_native/_api.py", line 3427, in read_buffer
>     tmp_buffer._map("READ_NOSYNC").sync_wait()
>   File "/Users/ahmedharmouche/Documents/tinygrad/venv/lib/python3.12/site-packages/wgpu/backends/wgpu_native/_helpers.py", line 268, in sync_wait
>     return self.finish()
>            ^^^^^^^^^^^^^
>   File "/Users/ahmedharmouche/Documents/tinygrad/venv/lib/python3.12/site-packages/wgpu/backends/wgpu_native/_helpers.py", line 260, in finish
>     raise RuntimeError(f"Waiting for {self.title} timed out.")
> RuntimeError: Waiting for buffer.map timed out.
> zsh: segmentation fault  PYTHONPATH=. WEBGPU=1 python3 examples/stable_diffusion.py --seed 0 --steps 1

### Observed behavior
When running tinygrad stable_diffusion.py, the buffer read times out when trying to get the [output of the decode step](https://github.com/tinygrad/tinygrad/blob/077e7e8ed2b2f0c211ee86f877b27e6147893f61/examples/stable_diffusion.py#L283). But it is not the buffer reading that takes that long, but to actually run the compute. Manually increasing the timeout from 5.0s [here](https://github.com/pygfx/wgpu-py/blob/c5108ac2e5de363a6b6c4d117c5521bf65d84dda/wgpu/backends/wgpu_native/_helpers.py#L237) solves it, but in `0.18.1` this just worked (the timeout wasn't there?). Now, we have stable diffusion working on faster machines, but on my local computer, it times out, so I have to manually increase this timeout value, so for now we downgraded to `0.18.1`.
Can this timeout be increased/disabled?

### Your environment
OS: MacOS Sonoma 14.4.1
Python version: 3.12
wgpu-py version: >=0.19.0
wgpu backend: Metal

almarklein:
Closed #650

cc @fyellin

### DIFF
--- a/wgpu/backends/wgpu_native/_helpers.py
+++ b/wgpu/backends/wgpu_native/_helpers.py
@@ -242,12 +242,11 @@ class WgpuAwaitable:
     truely async manner, as well as to support a synchronous version of them.
     """
 
-    def __init__(self, title, callback, finalizer, poll_function=None, timeout=5.0):
+    def __init__(self, title, callback, finalizer, poll_function=None):
         self.title = title  # for context in error messages
         self.callback = callback  # only used to prevent it from being gc'd
         self.finalizer = finalizer  # function to finish the result
         self.poll_function = poll_function  # call this to poll wgpu
-        self.maxtime = time.perf_counter() + float(timeout)
         self.result = None
 
     def set_result(self, result):
@@ -258,7 +257,7 @@ class WgpuAwaitable:
 
     def _is_done(self):
         self.poll_function()
-        return self.result is not None or time.perf_counter() > self.maxtime
+        return self.result is not None
 
     def _finish(self):
         try:


Korijn (approved):

fyellin:
Now that we have eliminated timeouts, I really do hate the fact that we are doing a busy wait.  Would it make sense to do a very slow backoff?  Maybe call `poll()` ten times in a row, and only then do we start to sleep, say 1ms between calls to poll().

For synchronous calls, I was wondering if it makes sense for us to use `poll_wait` instead of `poll`, but it really doesn't.  `poll_wait` waits until *Everything* on the queue is done, rather than waiting until just the action we're interested in is done.

Korijn:
It's probably more efficient to invest our time and energy into other things until upstream wgpu-native implements callbacks/futures/promises or something of that nature. Then we can build our polling-free implementation on top.

almarklein:
@Korijn I agree that we (hopefully) eventually get a polling-free implementation, but in the mean time, adding a bit of logic to sleep a bit longer than zero seconds sounds like a good idea to me. Apparently in some cases the future may take over 5 s to resolve, and that's a much longer time to busy-wait that we first anticipated.

@fyellin do you have time to create a pr for that? 

Maybe something like:
```sleep_time = min(0.01, sleep_time + 0.001)
```

fyellin:
Yes.  That was pretty much my thoughts exactly.  I didn't mind constant polling when we believed the results would always be returned quickly.  But now that we realize that polling may take 5s or more, than's not reasonable.  I'll come up with something reasonable (and easy to modify if you don't like it) and submit a PR.

Korijn:
It is clear that the use case in #650 is compute shaders right? For real time rendering use cases, results should always be returned quickly.

Regardless, I think both of you are probably right that some kind of back off mechanism is appropriate.

fyellin:
Probably, but not necessarily.

I could imagine someone attempting to do an extremely complicated rendering
with multiple light sources and complicated shadows and rendering this to a
texture rather than the screen.  The only way to get a texture from the GPU
to the CPU is via a buffer.

Korijn:
I guess you are right! 👍🏻

PR #655, opened 2024-12-13
Bring sleep back to WgpuAwaitable.  Slowly back off on the amount of time sleeping.

fyellin:


fyellin:
I've probably overengineered this.  

1. Should I be using `async_sleep` (defined in that file) or should I change `async_sleep` to use `anyio`? Now that we're actually sleeping, it's more important to use `Event` when sleeping.  (Some other operations `poll()` might cause this callback to happen).  Does everything that sniffio supports have an Event?

almarklein:
Yeah, we should only rely on `sniffio`, because the async stuff might actually be running on `rendercanvas.utils.asyncadapter`, which anyio does not know about. Both `sleep`  and `Event`  can be obtained in the way we now get the sleep func for `async_sleep` (it works for asyncio/trio/rendercanvas).

almarklein:
Ah, I see what you mean. We could technically add support for this to `rendercanvas`, but the simpler solution is to not use `Event`  here. Since we scale up the wait time, the possibly time we waited too long should only be a fraction of the total time we had to wait, so I don't see harm in that.

Could even make that explicit:
```py
sleep_time = time_since_started / 100
```

pyproject.toml
```diff
@@ -13,6 +13,7 @@ dependencies = [
     "cffi>=1.15.0",
     "rubicon-objc>=0.4.1; sys_platform == 'darwin'",
     "sniffio",
+    "anyio>=4.7.0"
```
almarklein, line 16:
```suggestion
```

Korijn (approved):

wgpu/backends/wgpu_native/_helpers.py
```diff
@@ -278,27 +273,52 @@ def sync_wait(self):
...
-        # can certainly copy from a function that *does* know what to return
+        # can certainly copy from a function that *does* know what to return.
+        # It would also be nice if wait_for_callback and sync_wait() could be merged,
+        # but Python has no wait of combining them.
         async def wait_for_callback():
-            # In all the async cases that I've tried, the result is either already set, or
-            # resolves after the first call to the poll function. To make sure that our
-            # sleep-logic actually works, we always do at least one sleep call.
-            await async_sleep(0)
+            # Set self.event before checking self.result to prevent data race.
```
almarklein, line 295: Leftover comment :)
```suggestion
            # Set self.event before checking self.result to prevent data race.
```

Korijn (approved):

almarklein:
Thanks @fyellin !

PR #659, opened 2024-12-30
Actually run the tests instead of fully skipping them

hmaarrfk:
Its easy to "fix tests" by completely skipping them ^_^_^_^_^

image


hmaarrfk:
I'm not sure what else in https://github.com/pygfx/wgpu-py/pull/604 needs to be "reworked"

examples/gui_auto.py
```diff
@@ -6,9 +6,10 @@
 
 from wgpu.gui.auto import WgpuCanvas, run
 
-from triangle import setup_drawing_sync
-# from cube import setup_drawing_sync
-
+try:
+    from .triangle import setup_drawing_sync
+except ImportError:
+    from triangle import setup_drawing_sync
```
hmaarrfk, line 12: you should likely just move these into a `wgpu/_testing_utils.py` file so you can import them correctly.


Its pretty annoying to not have a "real package structure". 

I honestly don't know a good way to "fix" this without putting these utilities INSIDE the `wgpu` package.

examples/cube.py
```diff
@@ -453,7 +453,7 @@ async def draw_frame_async():
 uniform_data = np.zeros((), dtype=uniform_dtype)
 
 
-if __name__ == "__main__":
+if True or __name__ == "__main__":
```
hmaarrfk, line 456: Its probably been a long day for me working on graphcis, but I can't understand why this won't run....

examples/gui_auto.py
hmaarrfk: I guess the point is to have this stuff in the example itself and not "hidden away" in some packaged file to make it feel like you have a very high level API...

hmaarrfk: I guess this is the warning:

```
examples/tests/test_examples.py::test_examples_run[triangle]
  /home/mark/miniforge3/envs/dev/lib/python3.10/runpy.py:126: RuntimeWarning: 'examples.triangle' found in sys.modules after import of package 'examples', but prior to execution of 'examples.triangle'; this may result in unpredictable behaviour
    warn(RuntimeWarning(msg))
```

examples/cube.py
```diff
@@ -453,21 +455,21 @@ async def draw_frame_async():
...
-        canvas.request_draw()
+def animate():
+    draw_frame()
+    canvas.request_draw()
 
-    canvas.request_draw(animate)
 
+canvas.request_draw(animate)
+
+if __name__ == "__main__":
```
hmaarrfk, line 474: I mostly untabbed and ran ruff.

Korijn (approved):

PR #660, opened 2025-01-15
use default lavapipe for screenshot

Vipitis:
this was missed since #350 and caused a panic. wgpu-py now has a different solution but this seems to be the root cause in shadertoy (which used the same ci workflows) https://github.com/pygfx/shadertoy/pull/37#issuecomment-2594147817

this PR is mostly to try if CI is still happy here.


Korijn (approved):

almarklein:
Thanks!

PR #666, opened 2025-01-26
Add patches for conda compatibility

hmaarrfk:
For the conda-forge packages I package wgpu-native myself and install them in the "system" libraries.

This should continue to allow you to use the "wheel" vendored headers but then also look for the headers in the "system" locations.

I've been dragging on a simpler (less upstreamable) version of this patch for about a year but I'm hoping I don't have to with this addition.

### DIFF
--- a/wgpu/backends/wgpu_native/_ffi.py
+++ b/wgpu/backends/wgpu_native/_ffi.py
@@ -4,7 +4,11 @@ import os
 import sys
 import logging
 
-from ..._coreutils import get_resource_filename, logger_set_level_callbacks
+from ..._coreutils import (
+    get_library_filename,
+    logger_set_level_callbacks,
+    get_header_filename,
+)
 
 from cffi import FFI, __version_info__ as cffi_version_info
 
@@ -19,8 +23,8 @@ if cffi_version_info < (1, 10):  # no-cover
 def get_wgpu_header():
     """Read header file and strip some stuff that cffi would stumble on."""
     return _get_wgpu_header(
-        get_resource_filename("webgpu.h"),
-        get_resource_filename("wgpu.h"),
+        get_header_filename("webgpu.h"),
+        get_header_filename("wgpu.h"),
     )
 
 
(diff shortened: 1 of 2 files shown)


Korijn:
Is there any (sane) way to test the different "deployment types" we want to support?

hmaarrfk:
I can add a conda forge based CI if you want

hmaarrfk:
I also have an other patch to disable the hatchling script.

Korijn:
Is it possible to unit test the required functionality? It seems like that should be sufficient?

hmaarrfk:
I’m not sure I can think of a test that is guaranteed to run successfully to completion on your local setups. 

Even to unit test the behavior. 

Would it be ok for me to add a CI job with a comment to delete it if/when I become inactive and unable to maintain it?

wgpu/backends/wgpu_native/_ffi.py
```diff
@@ -66,36 +66,41 @@ def get_wgpu_lib_path():
...
-        return embedded_path
+        raise RuntimeError(hints)
 
 
 def _maybe_get_hint_on_download_script():
-    root_dir = os.path.join(get_resource_filename(""), "..", "..")
+    root_dir = get_wgpu_py_root_dir()
     filename = os.path.abspath(
         os.path.join(root_dir, "tools", "download_wgpu_native.py")
     )
```
almarklein, line 106: I don't think this path is correct. We can make this bit more explicit, and drop `get_wgpu_py_root_dir`:
```suggestion
    lib_module = sys.modules[__name__.split(".")[0]]
    lib_dir = os.path.abspath(os.path.dirname(lib_module.__file__))
    print(lib_dir)
    filename = os.path.abspath(
        os.path.join(lib_dir, "..", "tools", "download_wgpu_native.py")
    )
```

almarklein:
Personally, I don't mind not-testing the conda-specific behaviour here.

Its (I assume) already tested in the conda forge [feedstock](https://github.com/conda-forge/wgpu-py-feedstock). When it fails, an update is made relatively easily. A test here (in the current repo) might catch such errors sooner, but I doubt whether the maintenance overhead is worth it, since I don't expect these paths to change very often.

hmaarrfk:
Ok let me test with conda-forge CIs before I unmark this as draft.

hmaarrfk:
https://github.com/conda-forge/wgpu-py-feedstock/pull/27

hmaarrfk:
Ok i backported the patch and it seems like it both tested the new path (since I found a bug in lib vs bin on windows) so this seems good to go from my end if it is cool with all of you!

almarklein (approved):

hmaarrfk:
thank you!

Korijn:
Can you explain how this works in the conda distribution? I would expect the new code to still prefer vendored headers when they're there. Does the conda package not ship the vendored headers?

hmaarrfk:
I've packaged the headers and the library and depend on those "c-like" packages in the recipe: https://github.com/conda-forge/wgpu-py-feedstock/blob/main/recipe/meta.yaml#L36

hmaarrfk:
and followup: https://github.com/conda-forge/wgpu-py-feedstock/pull/29

PR #669, opened 2025-01-31
Allow disabling downloading lib when creating wheel

almarklein:
Ref #666

@hmaarrfk Would this simple tweak allow the feedstock to drop another patch?

Not sure if there's an env var that we can check that a conda build is in progress. But in any case I imagine you can set an env var in the recipe.

### DIFF
--- a/tools/hatch_build.py
+++ b/tools/hatch_build.py
@@ -42,8 +42,11 @@ class CustomBuildHook(BuildHookInterface):
         # We only do our thing when this is a wheel build from the repo.
         # If this is an sdist build, or a wheel build from an sdist,
         # we go pure-Python mode, and expect the user to set WGPU_LIB_PATH.
+        # We also allow building an arch-agnostic wheel explicitly, using an env var.
 
-        if self.target_name == "wheel" and is_git_repo():
+        if os.getenv("WGPU_PY_BUILD_NOARCH") in ("1", "true"):
+            pass  # Explicitly disable including the lib
+        elif self.target_name == "wheel" and is_git_repo():
             # Prepare
             check_git_status()
             remove_all_libs()
--- a/tools/download_wgpu_native.py
+++ b/tools/download_wgpu_native.py
@@ -6,8 +6,6 @@ import tempfile
 import platform
 from zipfile import ZipFile
 
-import requests
-
 
 DEFAULT_UPSTREAM = "gfx-rs/wgpu-native"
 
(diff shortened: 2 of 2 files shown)


Korijn (approved):

tools/hatch_build.py
```diff
@@ -42,8 +42,11 @@ def initialize(self, version, build_data):
         # We only do our thing when this is a wheel build from the repo.
         # If this is an sdist build, or a wheel build from an sdist,
         # we go pure-Python mode, and expect the user to set WGPU_LIB_PATH.
+        # We also allow building an arch-agnostic wheel explicitly, using an env var.
 
-        if self.target_name == "wheel" and is_git_repo():
+        if os.getenv("WGPU_PY_BUILD_NOARCH") in ("1", "true"):
```
hmaarrfk, line 47:
```suggestion
        if os.getenv("WGPU_PY_BUILD_NOARCH") in ("1", "true", "True", "TRUE"):
```

Yeah this should work. sorry for the added spellings.

hmaarrfk: You are right. Sorry i forgot about the second argument.

hmaarrfk:
Would it be possible to also lazyload requests:
```patch
diff --git a/tools/download_wgpu_native.py b/tools/download_wgpu_native.py
index dfc17c0..6a18acf 100644
--- a/tools/download_wgpu_native.py
+++ b/tools/download_wgpu_native.py
@@ -6,8 +6,6 @@ import tempfile
 import platform
 from zipfile import ZipFile
 
-import requests
-
 
 DEFAULT_UPSTREAM = "gfx-rs/wgpu-native"
 
@@ -49,6 +47,8 @@ def write_current_version(version, commit_sha):
 
 
 def download_file(url, filename):
+    import requests
+
     resp = requests.get(url, stream=True)
     with open(filename, mode="wb") as fh:
         for chunk in resp.iter_content(chunk_size=1024 * 128):

```

I'm very weary of having requests as a dependency at build time since many (like this package) will pull in dependencies during the build process. Avoiding the dependency alltogether is an easy smoke test.

hmaarrfk:
can confirm that with the requests patch this works.

I mean I don't want to focus our attention on building the best build system. there is always some customization on this front for the build files.

Thanks for helping on the maintenance front!

almarklein:
Added the lazy-import.

Korijn:
If I may be so bold to make another suggestion, I would instead recommend to make the import conditional on the same environment variable. That way it will be clear  to future maintainers what the relationship is between the import and the build code.

almarklein:
I applied your **bold**  suggestion.

hmaarrfk:
Thanks this removes all the patching from conda-forge's distribution.

Korijn (approved):

Korijn:
Time for a release?

PR #673, opened 2025-02-21
update to wgpu-native 24.0.0.2

Vipitis:
Again trying to start the work on this... (in progress and will be draft for a while)
there is a release for 24.0.0.1 but wgpu23 never had a release for wgpu-native.
I am mainly motivated by the inclusion of https://github.com/gfx-rs/wgpu/pull/6755 

## useful links:
* diff [wgpu-core](https://github.com/gfx-rs/wgpu/compare/wgpu-v22.1.0...v24.0.0)
* diff [wgpu-native](https://github.com/gfx-rs/wgpu-native/compare/v22.1.0.5...v24.0.0.1)
* diff [headers](https://github.com/webgpu-native/webgpu-headers/compare/043af6c77e566f707db36759d9c9f161ebb616fd...bac520839ff5ed2e2b648ed540bd9ec45edbccbc) (is this correct via the ffi submodule in `wgpu-native`?)

### from template (skip upstream and IDL): 

* [x] Run `python tools/download_wgpu_native.py --version xx` to download the latest webgpu.h and DLL.
* [x] Run `python codegen` to apply the automatic patches to the code.
* [x] It may be necessary to tweak the `hparser.py` to adjust to new formatting.
* [x] Diff the report for new differences to take into account.
* [x] Diff `wgpu_native/_api.py` to get an idea of what structs and functions have changed.
* [x] Go through all FIXME comments that were added in `_api.py`:
        * Apply any necessary changes.
        * Remove the FIXME comment if no further action is needed, or turn into a TODO for later.
* [x] Run `python codegen` again to validate that all is well. Repeat the steps above if necessary.
* [x] Make sure that the tests run and provide full coverage.
* [x] Make sure that the examples all work.
* [x] Make sure that pygfx works (again).
* [x] Update release notes for changes in behavior.

## Wrapping up

* [ ] Update pygfx. not needed.
* [x] Release new wgpu-py.
* [x] This can be a good moment to deal with certain issues, or perhaps some issues can be closed.

## API changes

Nope. The `compilation_hints` arg got removed, but we did not support that anyway.


wgpu/backends/wgpu_native/_ffi.py
```diff
@@ -33,8 +33,8 @@ def _get_wgpu_header(*filenames):
     # Read files
     lines1 = []
     for filename in filenames:
-        with open(filename) as f:
-            lines1.extend(f.readlines())
+        with open(filename, "r") as f:
+            lines1.extend(f.read().replace("\\\n", "").splitlines(True))
```
Vipitis, line 37: this same code logic exists in codegen and here, and is also slightly different. So I have two different fixes too... especially since the file read stuff can be different between OS seem a bit janky.

wgpu/backends/wgpu_native/_mappings.py
```diff
@@ -3,107 +3,114 @@
...
-    "BlendOperation.min": 3,
-    "BlendOperation.reverse-subtract": 2,
-    "BlendOperation.subtract": 1,
-    "BufferBindingType.read-only-storage": 3,
-    "BufferBindingType.storage": 2,
-    "BufferBindingType.uniform": 1,
-    "BufferMapState.mapped": 2,
-    "BufferMapState.pending": 1,
-    "BufferMapState.unmapped": 0,
+    "AddressMode.clamp-to-edge": 1,
```
hmaarrfk, line 9: I’ve been curious about these updates so sorry for the non constructive comment. 

Did clap to edge really change in constant value?

Vipitis: this is just the bit for the mapping - they will change all for various reasons but shouldn't really impact behaviour. Sometimes new options are added and defaults change.

A good way to dig deeper is to find the diff (I linked the above) and then via blame get to the PR that made this.
For example this PR changed the order and added the undefined to `AddressMode`: https://github.com/webgpu-native/webgpu-headers/pull/364

Sometimes you find review comments or even meeting minutes that try to give reason. Othertimes you will just have to read the spec and follow it. hope this helps

hmaarrfk: Thank you!

wgpu/backends/wgpu_native/_api.py
```diff
@@ -452,33 +491,46 @@ def _request_adapter(
...
+                msg = "-" if message.data == ffi.NULL else ffi.string(*message).decode()
                 awaitable.set_error(f"Request adapter failed ({status}): {msg}")
             else:
                 awaitable.set_result(result)
 
+        # H: nextInChain: WGPUChainedStruct *, mode: WGPUCallbackMode, callback: WGPURequestAdapterCallback, userdata1: void*, userdata2: void*
+        callback_info = new_struct(
+            "WGPURequestAdapterCallbackInfo",
+            nextInChain=ffi.NULL,
+            mode=1,
```
Vipitis, line 519: since I have never done Async or much experience with it... I am hardcoding all these modes for now. Will have to read and try to understand the concepts here https://webgpu-native.github.io/webgpu-headers/Asynchronous-Operations.html and then figure out how they match to the concepts in python

Vipitis:
made some more progress... will try to keep at it but now run into the surface stuff which gave some trouble last time and is also really OS specific.

It might be worth to automatically generate the code for all these `callback` and `callbackInfo` structs or at least add the header info - I haven't tried to look into how the codegen actually works.

codegen/hparser.py
```diff
@@ -120,18 +122,30 @@ def _parse_from_h(self):
...
                     assert name.startswith("WGPU")
-                    name1 = name[4:-1]  # xxFlags -> xxFlag
-                    name2 = name[4:-5]  # xxFlags -> xx
-                    name = name1 if name1 in self.enums else name2
-                self.flags[name] = self.enums.pop(name)
+                self.flags[name] = self.enums.pop[name] if name in self.enums else {}
+
+        #fill flags
+        # schema: static const WGPUFlagName WGPUFlagName_Value = 0x0000000000000001;
+        for line in code.splitlines():
```
Vipitis, line 138: don't need two loops here

wgpu/backends/wgpu_native/_helpers.py
```diff
@@ -158,36 +158,36 @@ def get_surface_id_from_info(present_info):
             cv.setLayer(metal_layer)
             cv.setWantsLayer(True)
 
-        struct = ffi.new("WGPUSurfaceDescriptorFromMetalLayer *")
+        struct = ffi.new("WGPUSurfaceSourceMetalLayer *")
```
Vipitis, line 161: I made these changes just by following the schema - I don't have any Mac to test this myself. Also haven't tested the Linux variants just yet.

Vipitis:
more progress for today... got the first examples to run:
image

will hopefully find sometime tomorrow and keep at this pace.

wgpu/backends/wgpu_native/_api.py
```diff
@@ -3034,27 +3145,25 @@ def copy_texture_to_buffer(
...
             origin=c_origin,
             aspect=0,
-            # not used: nextInChain
         )
 
         c_destination = new_struct_p(
-            "WGPUImageCopyBuffer *",
-            buffer=destination["buffer"]._internal,
-            # H: nextInChain: WGPUChainedStruct *, offset: int, bytesPerRow: int, rowsPerImage: int
+            "WGPUTexelCopyBufferInfo *",
```
Vipitis, line 3158: codegen seems to only get the inner structs in these nested constructions

wgpu/backends/wgpu_native/_api.py
```diff
@@ -288,30 +309,65 @@ def _get_limits(id: int, device: bool = False, adapter: bool = False):
...
+    c_limits = new_struct_p(
+        "WGPULimits *",
+        # nextInChain=ffi.addressof(c_limits_native.chain), #seems to cause a crash with enumerate_adapter_async
```
Vipitis, line 326: this sorta works, but causes a rust panic - so I will have to look for a different solution

Vipitis:
all the examples have passed today, also most tests.
remaining issues with tests: limits for constants, something related to entry points (which changed to the `WGPUStringView`)

Vipitis:
okay. I think I have reached the point there I need some help.

To get the `WGPUNativeLimits` the pointer is needed, but does cause a crash when running the whole suite locally (could be gc stuff or something?). So uncommenting that one line and testing just the failing cases with `pytest tests -vvvsk "constants"` gives only one failure case: 
```py
FAILED tests/test_set_constant.py::test_render_bundle_push_constants_fails - Failed: DID NOT RAISE <class 'RuntimeError'>
```
I stepped through this test (be careful as the debug binaries seem to not work with native only limits -.-) and it doesn't raise anything and seems to work. So I wonder if this test is meant to work - not fail? 
It was added in #641 where the native functions were still missing. So can be be expected to work now as the functions have been added and do seemingly work? maybe cc: @fyellin 
(also why does the `._set_push_constants` function use additional `data_offset` when an `offset` is included? I feel like that can be dangerous with adding to the pointer?)

fyellin:
To answer the easy question.  There are two offsets, the offset of the data source and the offset of the constants pool that you're writing to.  The Rust interface (if I remember correctly) specifies a slice for the source.  The C interface provides a raw pointer.  The closest way I could implement that in Python was with two arguments data and data_offset.  If the Rust interface has changed, then I can fix that.

As to the test not failing.  At the time I wrote that code, you could not push constants with a RenderBundle, and  there would be an error if you tried to do that.  If that has since been implemented, then yes, this test needs to be rewritten!

Let me know if you have any other questions.

kushalkolar:
Thanks! :smile:  I was wondering why overrides for `@workgroup_size` weren't working, turns out it was added just a few months ago and isn't in native yet https://github.com/gfx-rs/wgpu/pull/6635

fyellin:
Once features are implemented in the C interface, I'm happy to help implement and test them in the Python interface.

almarklein:
Thanks for all the hard work, Jan! I'll have a look soon!

Vipitis:
I think I figured it out... test suite no longer crashes and I can continue with working through the failure cases now in the next few days.

*the ffi code was never the problem, it was a dictionary access -.-*

wgpu/backends/wgpu_native/__init__.py
```diff
@@ -11,8 +11,8 @@
 
 
 # The wgpu-native version that we target/expect
-__version__ = "22.1.0.5"
-__commit_sha__ = "fad19f5990d8eb9a6e942eb957344957193fe66d"
+__version__ = "24.0.0.1"
```
hmaarrfk, line 14: ps it seems that 24.0.0.1 was just released.

Vipitis:
All tests are finally passing. It will probably take me a few more days to get the new features to work (including examples/tests). So should I focus on cleanup and get this ready for review and open additional PRs for new features or try and get it all in here?

wgpu/backends/wgpu_native/__init__.py
hmaarrfk: this comment meant read 24.0.0.2

Vipitis: I pushed a commit to get the latest built. However I don't think any of the changes matter

wgpu/backends/wgpu_native/_api.py
Vipitis: upon reading this and giving it a think - I chose the 2nd option `WGPUCallbackMode_AllowProcessEvents` for most callbacks. And the least restrictive option `WGPUCallbackMode_AllowSpontaneous` for the device lost callback with the hope of avoiding crashes due to device lost - although I haven't had a test for this yet and might take a look at #603 again

almarklein:
> So should I focus on cleanup and get this ready for review and open additional PRs for new features or try and get it all in here?

Let's try to get it merged, and add new support for the new features in new prs. I plan to review this tomorrow.

wgpu/backends/wgpu_native/_ffi.py
almarklein: What case does the "\\\n" fix?

Vipitis: this escaped newline within a macro https://github.com/pygfx/wgpu-py/blob/78651f0ffeaf5ca25d15ea40b068c2e8a508932c/wgpu/resources/webgpu.h#L132

Turns out `.readlines()` doesn't really care and just escapes the backslash too

wgpu/backends/wgpu_native/_api.py
```diff
@@ -250,19 +252,37 @@ def to_c_string(string: str):
     return ffi.new("char []", string.encode())
 
 
+def to_c_string_view(string: str):
+    if string == ffi.NULL or not string or string is None:
```
almarklein, line 256:
```suggestion
    if not string:
```

almarklein:
Thanks for this amazing piece of work @Vipitis! This was not a small update 😉. I have read through all the changes, and made myself a list to look at some points more closely. I'll just push to this branch.

wgpu/backends/wgpu_native/_helpers.py
almarklein: Works on a mac!

codegen/hparser.py
```diff
@@ -128,26 +128,31 @@ def _parse_from_h(self):
                 parts = line.strip().strip(";").split()
                 assert len(parts) == 3
                 name = parts[-1]
-                if name.endswith("Flags"):
-                    assert name.startswith("WGPU")
-                self.flags[name] = self.enums.pop[name] if name in self.enums else {}
+                assert name.startswith("WGPU")
+                name = name[4:]
+                assert not name.endswith("Flags"), "XxxxFlags should not longer exist"
```
Vipitis, line 133: I suspect this might be handled by updating .idl too. I only briefly looked at it since updates to the parser were needed

almarklein: No, this is just because the headerfile used to define flags in a similar way to enums, so the flags were popped from of the list of eums. Now flags and enums have their own logic. I just left a few asserts in place to safeguard that.

almarklein:
This is ready from my end. I ran the tests, examples, also for pygfx, on MacOS, WIndows, and Linux, and all is looking good!

almarklein:
@Vipitis feel free to comment on my changes. Let's not merge this until you're happy too.

wgpu/backends/wgpu_native/_api.py
```diff
@@ -249,11 +249,12 @@ def _get_override_constant_entries(field):
 
 
 def to_c_string_view(string: str):
-    # if string is None:
-    #     # The null-string
-    #     data = ffi.NULL
-    #     length = lib.WGPU_STRLEN
-    if not string:
+    if string is None:
```
Vipitis, line 253: Is there any situation where this might receive `ffi.NULL`? Or does it evaluate to `None` too. I never checked what it's `__bool__()` is

Vipitis:
@almarklein thanks for the improvements. I followed this closely and it looks good to me! 

Regarding the difference between null string and empty string: there is no test that specifically requires the empty string, not sure where this would even be the case. when not specifying an entry point for example you specifically need it to be this null string. Perhaps empty string is more of an output use case for `WGPUStringView` 

For the view usages I branched off my example and will modify it in the next couple of days to test and show it's functionality and then hopefully have it in shadertoy soon.

wgpu/backends/wgpu_native/_api.py
almarklein: The intend is that `to_c_string_view()` converts a Python string or None to a ctypes object, so `ffi.NULL`  should not be used. I checked that this is the case right now. If `ffi.NULL`  would be used, it evaluates to false, resulting in an empty string.

almarklein:
> there is no test that specifically requires the empty string, not sure where this would even be the case. 

I guess the behavior depends on the case. Thinking of it, maybe it would be cleaner to also use the null-string if no label is provided.

As for the `entry_point`, if the user sets it to None, it will become the null-string, so this should be fine.

wgpu/backends/wgpu_native/_api.py
```diff
@@ -2522,6 +2522,8 @@ def create_view(
             elif dimension in ("2d-array", "cube-array"):
                 array_layer_count = self._tex_info["size"][2] - base_array_layer
 
+        usage = 0  # WGPUTextureUsage_None - assuming none means use the texture's usage
```
Vipitis, line 2527: using the to flag directly and avoid magic numbers should help spot issues quicker should the mapping change. Happened to me a few times with `status = 0`

almarklein:
@Korijn This is ready from our (Jan and me) end 🚀

Korijn (approved):

almarklein:
Let's make a release real soon. I'll have a quick look a updating idl.

PR #677, opened 2025-02-27
add simple matmul example

kushalkolar:
Very naive simple implementation of matrix multiplication for arbitrary sized matrices in wgsl.


Korijn (approved):
Nice example!

PR #679, opened 2025-03-09
Avoid special characters in gpu report

hmaarrfk:
Unicode is cute, but still to this day I find incompatibilities with it.


hmaarrfk:
I guess i should just find out why my font doesn't suppor the full unicode block:

█ U+2588

almarklein:
The tofu placeholder glyph comes pretty close 😉

hmaarrfk:
> The tofu placeholder glyph comes pretty close 😉

Which unicode glyph is that?

almarklein:
https://unicodedecode.com/tofu

PR #680, opened 2025-03-11
Improve performance for 'bitmap' render method

almarklein:
I did some experiments and this is a relatively easy way to improve performance by about 20%, going from ~24 to ~30 fps on fullscreen window on a 4k monitor on my M1 MacBook.

Tricks applied:
* ~~re-use the target texture~~ -> let's not for memory concerns, also seems to not help much.
* Reuse (and share) the buffer needed to copy the texture data to the CPU.
* Avoid a data-copy when bytes-per-row is not a multiple of 256.
* ~~use numpy for copying~~ does not seem to help much, and would introduce a new optional dependency.

We can go much faster, but for that we need to have our async stuff sorted out better.
More details here: https://github.com/pygfx/rendercanvas/issues/40#issuecomment-2714064032

### DIFF
--- a/wgpu/_classes.py
+++ b/wgpu/_classes.py
@@ -226,7 +226,6 @@ class GPUCanvasContext:
 
         # The last used texture
         self._texture = None
-        self._bitmap_texture = None  # only used when method is 'bitmap'
 
         # Determine the present method
         self._present_methods = present_methods
@@ -388,7 +387,6 @@ class GPUCanvasContext:
             self._unconfigure_screen()
         self._config = None
         self._drop_texture()
-        self._bitmap_texture = None
 
     def _unconfigure_screen(self):
         raise NotImplementedError()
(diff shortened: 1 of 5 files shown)


wgpu/_classes.py
```diff
@@ -457,25 +467,32 @@ def present(self):
         self._drop_texture()
         return result
 
-    def _present_bitmap(self):
-        texture = self._texture
-        device = texture._device
-
-        size = texture.size
-        format = texture.format
+    def _get_bytes_per_pixel(self, texture_format):
```
almarklein, line 470: This bit of the diff looks complicated, but all I did was move the logic that computes bytes_per_pixel in a method.

hmaarrfk:
Is this "bitmap" method the current path for Wayland + Qt?

almarklein:
> Is this "bitmap" method the current path for Wayland + Qt?

yes

almarklein:
Solved this by using a shared copy-buffer. It looks like re-using the texture does not do much; re-using the temporary buffer was what contributed by far the most to the speed-up.

hmaarrfk:
> e-using the temporary buffer was what contributed by far the most to the speed-up.

can you point to the one you are referring to in the diff (for my own learning)

wgpu/backends/wgpu_native/_api.py
```diff
@@ -3550,27 +3551,50 @@ def write_texture(
...
+        if is_present_texture:
+            copy_buffer, time_since_size_ok = self._shared_copy_buffer
+            if copy_buffer is None:
+                pass  # No buffer
+            elif copy_buffer.size < data_length:
+                copy_buffer = None  # Buffer too small
+            elif copy_buffer.size < data_length * 4:
+                self._shared_copy_buffer = copy_buffer, time.time()  # Bufer size ok
+            elif time.time() - time_since_size_ok > 5.0:
+                copy_buffer = None  # Too large too long
```
almarklein, line 3584: @hmaarrfk this one. If this is a texture used for presenting, it checks the shared buffer. If it's large enough, and not too large for a too long time, it will re-use it.

almarklein:
This does not seem to clash with #673, except for a change in `tests_mem/testutils.py` which seems to be exactly the same.

Korijn (approved):

almarklein:
I tested performance on Windows and Linux, and did not observe a difference. That's on a machine with integrated graphics. Will test on a Windows machine with a GPU later today.

PR #686, opened 2025-03-17
fix newlines

Vipitis:
I updated my fork and local branches after the #673 merge and redownloaded the lib using the script. Got hit with an `pycparser.plyparser.ParseError: <cdef source string>:11:1: Illegal character '\r'` on my windows machine.

here is a fix, but likely not the best (still error prone due to the duplicated code for reading the headers).
Also maybe we can get a Windows runner for CI?


Korijn (approved):

almarklein:
> Also maybe we can get a Windows runner for CI?

👌

PR #697, opened 2025-03-30
Add return type hints

### LINKED ISSUE #696: return type hints in `_classes.py` and `_api.py` ?
Hey, something that has bothered me for a while now: 

Some of the functions that return a different object, such as `GPUDevice.create_command_encoder()` returns a `GPUCommandEncoder`. So is there any issue with providing this type annotation in general? I only played around with a few manual examples. (I can look into code gen improvements to add that, time permitting).
The main point being that in the editor I am always worried about white text for methods. And I also can't easily jump to the definition to check the arguments needed... lot's of ctrl+F 
I have been using the following pattern 
```py
encoder: wgpu.GPUCommandEncoder = device.create_command_encoder()
```
which gives me the unimplemented function in `_classes.py` which as least as a docstring and the args. But getting to the actual version in `wgpu_native/_api.py` would be even more helpful.

So more like a question to the other devs, do you have a editor based solution or some other trick?

Vipitis:
closes #696 

found where the type hints can be added in codegen, had to work through a bunch of edge cases. Will see if these cases can be added to already existing utils.
also not sure if `-> None` is really wanted everywhere.

still in progress... hopefully I can get this done early next week.

as for the 2nd question in the issue, I looked through some ideas with stub classes in mypy, .pyi files and even using `if "TYPE_CHECKING"` to import the autobackend. But doesn't feel like that solves the issue if having the "go to declaration" button go to the implementation - I might keep this an option issue.

### DIFF
--- a/codegen/apipatcher.py
+++ b/codegen/apipatcher.py
@@ -432,6 +432,19 @@ class IdlPatcherMixin:
         idl_line = functions[name_idl]
         args = idl_line.split("(", 1)[1].split(")", 1)[0].split(",")
         args = [Attribute(arg) for arg in args if arg.strip()]
+        return_type = idl_line.split()[0]
+        if return_type.startswith("["):
+            # skip the [NewObject] and [SameObject] parts?
+            return_type = idl_line.split()[1]
+        if return_type.startswith("Promise"):
+            # resolve Promise<Type> to Type (should be in .resolve_type() I think)
+            return_type = return_type[8:-1]
+        return_type = return_type.removeprefix("constructor(")
+        if return_type == "optional":
+            return_type = idl_line.split()[1]
+        return_type = self.idl.resolve_type(return_type)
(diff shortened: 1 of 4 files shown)


almarklein:
> also not sure if `-> None` is really wanted everywhere.

Normally I'd say don't bother, but for an API like this I like the explicitness.

Vipitis:
I think this should be good mostly. please have a look as some of the edge cases feel hacky as usual. 
For the `request_adapter_` functions I added the return type manually due to `@apidiff.change`.

Will look into adding typing for properties/attributes (I think also the args to some of the constructors like `GPUDevice`) in a different PR to finally get the type for `device.queue`

almarklein (approved):
Nice work!

Korijn (approved):

Korijn:
There are merge conflicts !

Vipitis:
ah, got to rebase and likely rerun codegen. But will probably only get to that later today

Issue #695, opened 2025-03-26
No access to subgroup_min_size, subgroup_max_size


fyellin:
This is actually several interlocked bugs that I noticed when playing around with subgroups.

#1) The two classes named `GPUAdapterInfo` are never actually used. There is no reference to them anywhere else in the code.  When creating an Adapter, the current Python code pulls the information it needs out the C structure and then releases the C structure. 

#2) The never used class GPUAdapterInfo has properties `subgroup_min_size` and `subgroup_max_size`, but there is no such field in the corresponding `WGPUAdapterInfo` structure. Presumably this is a bug in the Rust -> C interface that this field was forgotten?

almarklein:
> The two classes named GPUAdapterInfo are never actually used. 

Good point! the `adapter.info` property simply returns a dict. We can improve this by instead returning `GPUAdapterInfo`. And then implement `__getitem__` and `__repr__` so it looks a bit more like a dict. So this is relatively easy to fix.

> The never used class GPUAdapterInfo has properties subgroup_min_size and subgroup_max_size, but there is no such field in the corresponding WGPUAdapterInfo structure. Presumably this is a bug in the Rust -> C interface that this field was forgotten?

IIRC these fields were added only recently to the WebGPU spec. They have been added to [webgpu.h](https://github.com/webgpu-native/webgpu-headers/blob/main/webgpu.h) on [jan 31](https://github.com/webgpu-native/webgpu-headers/pull/509), but wgpu-native uses `webgpu.h` commit [bac5208](https://github.com/webgpu-native/webgpu-headers/tree/bac520839ff5ed2e2b648ed540bd9ec45edbccbc) from 12-11-2024. So someone needs to bump `webgpu.h in wgpu-native.

almarklein:
I think both problems have been addressed by now. Feel free to re-open if you disagree.

PR #699, opened 2025-04-02
Use GPUAdapterInfo.

fyellin:
Bug #695 
Use GPUAdapterInfo.  Add subgroup_min_size and subgroup_max_size.  

For now, we can use default sizes from Spec, but eventually these will be available in the C WGPUAdapterInfo.


Korijn (approved):

almarklein (approved):
Thanks!

PR #701, opened 2025-04-09
Implement proper support for alt texture formats for offscreen use

### LINKED ISSUE #698: How to render to rgba16float with offscreen canvas?
I am working in a team where we have a [scientific visualisation code](https://github.com/pynbody/topsy) where we sometimes choose to render to rgba16float to make an HDR interface. (We realise color space and tone mapping is a work in progress, but in practice it works just fine for our use case.)

Previously we were using wgpu 0.15 and this would work either for onscreen or offscreen canvases. On updating to 0.21, we can still do this onscreen but when offscreen we fall over due to a [check against the capabilities of the device](https://github.com/pygfx/wgpu-py/blob/main/wgpu/_classes.py#L341). On investigation, it seems like these capabilities are actually [hard-coded](https://github.com/pygfx/wgpu-py/blob/main/wgpu/_classes.py#L257) for offscreen canvases. The main purpose of rendering offscreen is for testing, so we are in a position where the intended functionality actually continues to work but we can't test it.

I am sure we can hack around this, but is there any accepted/correct approach?

Many thanks indeed for any guidance/hints, and also for this excellent package.

almarklein:
Closes #698

* [x] Offscreen canvas defines that it supports more formats.
* [x] CanvasContext queries the formats from the canvas to get canvas capabilities (instead of assuming only `rgba8unorm`.
* [x] CanvasContext converts the downloaded data to the appropriate dtype.
* [x] Update triangle example to allow setting format.
* [x] Add example.

Will also adust the `rendercanvas` to make it compatible.

### DIFF
--- a/examples/offscreen_hdr.py
+++ b/examples/offscreen_hdr.py
@@ -1,0 +1,44 @@
+"""
+Render offscreen using a 16bit (HDR) render target
+--------------------------------------------------
+
+Render a wgpu example offscreen, safe to 16bit PNG, and show the result.
+
+"""
+
+import os
+import tempfile
+import webbrowser
+
+import numpy as np
+import png  # provided by the pypng package
+
+# from rendercanvas.offscreen import RenderCanvas
+from wgpu.gui.offscreen import WgpuCanvas as RenderCanvas
+from triangle import setup_drawing_sync
+
+
+canvas = RenderCanvas(size=(640, 480), pixel_ratio=2)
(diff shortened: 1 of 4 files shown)


Korijn (approved):

almarklein:
By the way, we will need to make adjustments in Pygfx too, eventually. E.g. allow changing the format of the internal textures to f16. But for now this opens up some HDR support in wgpu-py.

PR #703, opened 2025-04-09
add type hints for properties

Vipitis:
follow up to #697 

goal is to add type annotations to properties in `_classes.py` via codegen. 
A lot of properties resolve to simple `int` or `str` python types while others are enums.

Will look into expanding `resolve_type` where it's needed or find better workarounds for features/limits. I am not 100% sure how type hints get inherited from these stubs, as there have been changes every single python release.

### DIFF
--- a/codegen/idlparser.py
+++ b/codegen/idlparser.py
@@ -241,6 +241,8 @@ class IdlParser:
             name = name.strip("()")
             names = [self.resolve_type(t).strip("'") for t in name.split(" or ")]
             names = sorted(set(names))
+            if len(names) == 1:
+                return names[0]
             return f"Union[{', '.join(names)}]"
         if name.startswith("Promise<") and name.endswith(">"):
             name = name.split("<")[-1].rstrip(">")
--- a/codegen/idlparser.py
+++ b/codegen/idlparser.py
@@ -223,7 +223,7 @@ class IdlParser:
         name = pythonmap.get(name, name)
 
         # Is this a case for which we need to recurse?
-        if name.startswith("sequence<") and name.endswith(">"):
+        if (name.startswith("sequence<") or name.startswith("FrozenArray<")) and name.endswith(">"):
             name = name.split("<")[-1].rstrip(">")
             name = self.resolve_type(name).strip("'")
             return f"List[{name}]"
--- a/codegen/idlparser.py
+++ b/codegen/idlparser.py
@@ -219,7 +219,7 @@ class IdlParser:
             "GPUExternalTexture": "object",
             "undefined": "None",
             "ArrayBuffer": "memoryview",
-            "WGSLLanguageFeatures" : "set",
+            "WGSLLanguageFeatures": "set",
             "GPUSupportedFeatures": "set",
             "GPUSupportedLimits": "dict",
             "EventHandler": "None",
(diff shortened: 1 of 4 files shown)


codegen/idlparser.py
```diff
@@ -219,11 +219,13 @@ def resolve_type(self, typename):
...
             "undefined": "None",
             "ArrayBuffer": "memoryview",
+            "WGSLLanguageFeatures" : "set",
+            "GPUSupportedFeatures": "set",
         }
         name = pythonmap.get(name, name)
 
         # Is this a case for which we need to recurse?
-        if name.startswith("sequence<") and name.endswith(">"):
+        if (name.startswith("sequence<") or name.startswith("FrozenArray<")) and name.endswith(">"):
```
Vipitis, line 228: the spec says this is not a sequence type: https://webidl.spec.whatwg.org/#idl-frozen-array but I believe in python it's list either way. Plus is is currently just used for `GPUCompilationInfo` which is unimplemented in wgpu-native

wgpu/resources/codegen_report.md
```diff
@@ -11,7 +11,10 @@
 ### Patching API for _classes.py
 * Diffs for GPU: add enumerate_adapters_async, add enumerate_adapters_sync, change get_preferred_canvas_format, change request_adapter_async, change request_adapter_sync
 * Diffs for GPUCanvasContext: add get_preferred_format, add present
+* Error resolving type for GPUAdapter.limits: ('Encountered unknown IDL type: ', 'SupportedLimits')
```
Vipitis, line 14: the situation for Limits sounds more complex: https://github.com/gpuweb/gpuweb/issues/4277
But I guess it will be set, maybe list eventually. We can also just not annotate it for now.

almarklein: Limits would be a dict (name -> limit-value). Features is a `set`. But yeah, I think we can map `SupportedLimits` to dict in that big mapping-dict in the codegen?

Vipitis: makes sense, I will only get to this next week as I am away from any computer right now.

almarklein: No worries! I was just reminded of this comment because I looked at `AdapterLimits` (which is arguably also just a dict).

wgpu/_classes.py
```diff
@@ -623,19 +623,19 @@ def __init__(self, internal, features, limits, adapter_info):
 
     # IDL: [SameObject] readonly attribute GPUSupportedFeatures features;
     @property
-    def features(self):
+    def features(self) -> set:
```
Vipitis, line 626: since the IDL line gets used for the property down here, it isn't used for the args to `__init__()` above. I feel like this could be useful to have - so I will think about a solution for that.

almarklein: The `__init__` is not public API (adapter objects are only created via `request_adapter`), so I guess it's not needed, right? Arguably its handy for when developing this code itself, but then does it justify the added complexity of the codegen?

almarklein (approved):
Nice work!

wgpu/_classes.py
Vipitis: I was just wondering why all the other methods had nice type hints already, it doesn't matter in practice.

Korijn (approved):

PR #711, opened 2025-05-02
fix imgui example

Vipitis:
tiny issue I encountered with one of the examples, seems to be from #706

not sure if this is 100% the correct way but it seems to mostly work. The odd part is that I was able to run this example a few days ago on my other machine which I strongly think was on main branch already...

### DIFF
--- a/examples/imgui_backend_sea.py
+++ b/examples/imgui_backend_sea.py
@@ -402,8 +402,9 @@ def render():
     render_pass.draw(3, 1)
 
     # draw imgui
+    psize = canvas.get_physical_size()
     imgui_data = gui(app_state)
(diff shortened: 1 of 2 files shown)


Vipitis:
I should probably add a docstrings to the `.render` method too... Just saw that, will update the PR in maybe two days when I am back

almarklein (approved):

panxinmiao:
Uh, I feel the `psize` parameter is a bit odd, and it might be inconsistent with the state of `draw_data`... 🤔

Korijn:
Is this PR still needed now that #713 is merged?

Vipitis:
no, it's no longer needed

PR #713, opened 2025-05-06
Set the size via `imgui.io` instead of using the `psize` parameter of the render method.

panxinmiao:
See: https://github.com/pygfx/wgpu-py/pull/711#issuecomment-2853486448
We also need to update the `display_size` in `draw_data`, as it is used when setting the camera state.

I noticed that this change was made in #706, but I couldn’t really tell the difference. 🤔 

Also, in theory, if we update the size value every time the render method is called, we shouldn’t need to handle the canvas `on_resize` event anymore — but it seems we still can’t remove that event handler.

### DIFF
--- a/wgpu/utils/imgui/imgui_backend.py
+++ b/wgpu/utils/imgui/imgui_backend.py
@@ -380,7 +380,6 @@ class ImguiWgpuBackend:
         self,
         draw_data: imgui.ImDrawData,
         render_pass: wgpu.GPURenderPassEncoder,
-        psize: tuple,
     ):
         """
         Render the imgui draw data with the given render pass.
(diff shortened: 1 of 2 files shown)


almarklein:
For context, #706 was done because currently a draw can occur after a resize, before the resize event is emitted. This is a problem to solve at `rendercanvas`.

I like this implementation. I feel it'd be nicer to store physical size and then derive the logical size from the physical size and ratio, because it eliminates viewport mismatches based on rounding errors, but I guess that's how imgui works.

> we shouldn’t need to handle the canvas on_resize event anymore — but it seems we still can’t remove that event handler.

Yeah I think so. Why can't we remove the handler?

panxinmiao:
When the window is resized to a smaller size, the size retrieved in the backend's render method may be outdated, which can result in validation errors related to the viewport or scissor_rect.
In contrast, the `on_resize` method does not have this issue. I suspect this is because, during a window shrink, the canvas always triggers the `resize` event before invoking the `present` (i.e., render) method.

PR #714, opened 2025-05-10
Update examples to use rendercanvas

### LINKED ISSUE #710: Update examples to use rendercanvas
Move #627 a bit closer.

Vipitis:
closes #710 

I am mainly just applying the conversion instruction.
A few more todos:
* [x] update tests
* [ ] ~~update ci~~ (?)
* [x] subprocess example
* [x] fix HDR example
* [x] decide on scheduling pattern
* [ ] update docs/readme ?


examples/offscreen_hdr.py
```diff
@@ -13,8 +13,8 @@
 import numpy as np
 import png  # provided by the pypng package
 
-# from rendercanvas.offscreen import RenderCanvas
-from wgpu.gui.offscreen import WgpuCanvas as RenderCanvas
+from rendercanvas.offscreen import RenderCanvas
+# from wgpu.gui.offscreen import WgpuCanvas as RenderCanvas
```
Vipitis, line 17: not sure what is going on, since I am getting the `Configure: unsupported texture format: rgba16float not in ['rgba8unorm-srgb', 'rgba8unorm']` issue here (with either option). And this does work in pytest. So I am thinking that non offscreen canvas is requested via `.auto` in the triangle.py script.

However even setting the env var here to force offscreen didn't help. It could be unrelated as I never run this test standalone before.

examples/gui_glfw.py
```diff
@@ -1,16 +1,16 @@
...
 """
 
 # run_example = false
 
-from wgpu.gui.glfw import WgpuCanvas, run
+from rendercanvas.glfw import GlfwRenderCanvas, loop
 
 from triangle import setup_drawing_sync
-# from cube import setup_drawing_sync
+# from cube import setup_drawing_sync #TODO: this variant spawns and additional window!
```
Vipitis, line 10: the rendercanvas version of cube.py seems to spawn a window by just importing this, doesn't happen with the triangle variant.

so I believe the import might need to be moved inside the main function

examples/offscreen_hdr.py
almarklein: Have you updated to the most recent `rendercanvas`?

examples/gui_glfw.py
almarklein: I think the cube example was once written to be re-usable, and then later it evolved to create the window in the main module (not in `if __name__ ...`).

examples/offscreen_hdr.py
Vipitis: I got the 2.0.4 release from pypi, but this working through pytest suggest an env messup - will double check. 

Otherwise will step through this with a debugger later as it might load a non offscreen canvas via auto or something 🤔

Vipitis: was an env issue -.- (having multiple python versions and editable installs ins't proper env management)

examples/gui_glfw.py
Vipitis: turns out that public var is used screenshot testing, so I have to think about any possible solutions that don't feel awkward

examples/cube.py
```diff
@@ -459,17 +459,9 @@ async def draw_frame_async():
...
-def animate():
-    draw_frame()
-    canvas.request_draw()
-
-
-canvas.request_draw(animate)
 
 if __name__ == "__main__":
-    run()
+    canvas = RenderCanvas(size=(640, 480), title="wgpu cube example at $fps using $backend", update_mode="continuous", max_fps=-1, vsync=True)
```
Vipitis, line 464: so it seems like the default 30fps run at like 21 for me, and setting it to 60 gives me ~41. without vsync and fastest it easily runs over 1k fps... but setting -1 and contiuous with vsync on matches my monitor refreshrate just fine.

not sure if continuous, -1 and vsync on and fastest with vsync on are actually the same.

(I might file an issue if I understand the problem here...)

examples/gui_glfw.py
almarklein: Or we just don't re-use the cube example 🤷

examples/cube.py
almarklein: Interesting ... what system/os are you on again? In my experience it sometimes dips initially but then levels pretty quickly to the max-fps.

Vipitis: This is on like a 2015 Skylake i5 running windows 10 and an Intel dGPU (A750) but using modern python 3.12.9. the low framerate stays even after several seconds. 

I also have a laptop with slightly more modern CPU and Nvidia dGPU to observe this behavior.

I want to throw this through a profiler to see if this is an issue with queued frames and someone waiting for too long. Will also check what present mode is actually selected for Vsync on and off.

Vipitis: So not much of a performance issue today, seems to nearly approach 30 or 60 when I set that... will spent some more time to figure that out with profiling etc, but out of scope for this PR.

Updated to 2.1.1 where `-1` no longer works (it gets set to 1 in the new method), so I suppose we want to use `fastest` to get to the monitors refresh rate via vsync or keep it at the 30 default?

examples/gui_glfw.py
Vipitis: I put the canvas creation and draw func registration logic into the test function itself... in theory that should do the trick - will wait for CI here.

examples/gui_subprocess.py
```diff
@@ -26,31 +26,32 @@
...
-from wgpu.gui.qt import WgpuCanvas
+from PyQt6 import QtWidgets  # Use either PySide6 or PyQt6
+from rendercanvas.qt import RenderCanvas
 
 app = QtWidgets.QApplication([])
-canvas = WgpuCanvas(title="wgpu triangle in Qt subprocess")
+canvas = RenderCanvas(title="wgpu triangle in Qt subprocess", update_mode="ondemand")
 
-print(json.dumps(canvas.get_present_methods()))
+print(json.dumps(canvas._subwidget._rc_get_present_methods()))
```
Vipitis, line 35: since the wrapper doesn't include any of the `_rc_` methods I had to access the subwidget for the modes.

examples/gui_subprocess.py
```diff
@@ -84,4 +85,5 @@ def _request_draw(self):
 # Go!
 draw_frame = setup_drawing_sync(canvas)
 canvas.request_draw(draw_frame)
+canvas.force_draw()
```
Vipitis, line 88: not 100% sure what the solution is to have draws be shedulled from the qt loop, but the function gets registered and works if I force it here.

examples/cube.py
almarklein: 30 is the default because it is enough for basic tasks and can safe battery power compared to fastest.

examples/gui_subprocess.py
almarklein: we can delete this subprocess example; it worked at some point, but it was not supposed to, and won't in the future.

examples/gui_subprocess.py
almarklein: wait ... you *do* get this example to work????

examples/gui_auto.py
```diff
@@ -4,14 +4,14 @@
...
-from wgpu.gui.auto import WgpuCanvas, run
+from rendercanvas.auto import RenderCanvas, loop
 
 try:
     from .triangle import setup_drawing_sync
 except ImportError:
     from triangle import setup_drawing_sync
 
-canvas = WgpuCanvas(title=f"Triangle example on {WgpuCanvas.__name__}")
+canvas = RenderCanvas(title=f"Triangle example on {RenderCanvas.__name__}")
```
almarklein, line 14: Rendercanvas has a little placeholder DSL for the title:
```suggestion
canvas = RenderCanvas(title=f"Triangle example on $backend")
```

examples/tests/test_examples.py
```diff
@@ -81,8 +83,13 @@ def unload_module():
 
     request.addfinalizer(unload_module)
 
+    # create a offscreen canvas here and just use the imported .setup_drawing function to get the example
+    canvas = RenderCanvas(size=(640, 480), update_mode="manual")
+    draw_frame = example.setup_drawing_sync(canvas)
+    canvas.request_draw(draw_frame)
```
almarklein, line 89: I'd rather just use the canvas in the example. Because it's more generic than having the `setup_drawing_sync` function.

Because of  `RENDERCANVAS_FORCE_OFFSCREEN` the example's `canvas` will be the offscreen canvas (if imported from `rendercanvas.auto`.

FWIW the offscreen canvas does not scheduler draws by itself.

examples/gui_subprocess.py
almarklein: I just saw your other comment. Ok, we can try to keep this one working ...

examples/gui_subprocess.py
```diff
@@ -16,41 +16,42 @@
...
+from cube import setup_drawing_sync
 
 
 code = """
 import sys
 import json
-from PySide6 import QtWidgets  # Use either PySide6 or PyQt6
-from wgpu.gui.qt import WgpuCanvas
+from PyQt6 import QtWidgets  # Use either PySide6 or PyQt6
+from rendercanvas.qt import RenderCanvas
```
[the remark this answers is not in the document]
almarklein: Oh nevermind, you need the qt app, I think

examples/tests/test_examples.py
Vipitis: I had considered to do a try/hasattr for the test to make it work with either approach. But that seems really messy for a test. I guess the main concern was that the cube example is convenient to import like the triangle example - and when you don't have that env var set, creating the canvas does have side effects(which I think `wgpu.gui` didn't).

But I also came to the realization that the triangle and cube examples are fairly awful standalone, as you have to jump through like 4 functions to see the whole picture. Perhaps there should be a minimal script version (without any unnecessary functions). 
How I changed cube example now is basically provided by `rendercanvas.utils.cube` too, which should import without side effects. (and perhaps these are more rendercanvas examples than wgpu examples anyway?)

proposal: I revert screenshot test and cube to their previous behavior and then import the cube drawing function from rendercanvas. Alternatively I don't use the cube example at all and it will be all triangle.

almarklein: > I guess the main concern was that the cube example is convenient to import like the triangle example - and when you don't have that env var set, creating the canvas does have side effects(which I think wgpu.gui didn't).

What side effects do you mean? I do see that printing the adapter summaries could be moved into the `if __name__ ...`.

>  import the cube drawing function from rendercanvas

I'd rather have our own implementation of the cube example here, so we can adjust it to any API changes directly.

I was going to say that the triangle is fine, but the fact that the cube is animated makes it more interesting, e.g. for the pause button in the qt-embed example.

almarklein (approved):
Is this ready from your end?

Vipitis:
yeah

Korijn (approved):

PR #715, opened 2025-05-14
add compute texture example that converts RGBA -> YCbCr

kushalkolar:
Adds example that shows how to setup textures and a pipeline to run a compute shader on the textures to convert from RGBA -> YCbCr and also read the output textures.

image

Questions:

1. Would be nice to plot this directly with wgpu but is that easy to do :laughing: , or I can just leave the commented out fpl code at the bottom.
2. How do you debug shaders, I was doing a lot of things like this where workgroup invocations are performed on 8x8 blocks of the texture. Writing the workgroup_id value to the texture to make sure I've written my for loops correctly and structured my invocations correctly, looking at the output values and the image as well.

image

```wgsl
var L = f32(wid.x) + f32(wid.y);

// store luma channel
textureStore(tex_y, pos, vec4<f32>(L, 0, 0, 0));
```

### DIFF
--- a/examples/compute_textures.py
+++ b/examples/compute_textures.py
@@ -1,0 +1,279 @@
+"""
+Example that shows how to use textures in a compute shader to convert an RGBA image to YCbCr.
+The output chroma channels (CbCr) are subsampled by 2 along the rows and columns, typical of 4:2:0 formats.
+
+The shader uses workgroups to processes non-overlapping 8x8 blocks of the input rgba texture.
+"""
+
+import numpy as np
+import wgpu
+import imageio.v3 as iio
+
+
+def size_from_array(data, dim):
+    # copied from pygfx
+    # Check if shape matches dimension
+    shape = data.shape
+
+    if len(shape) not in (dim, dim + 1):
+        raise ValueError(
+            f"Can't map shape {shape} on {dim}D tex. Maybe also specify size?"
+        )
+    # Determine size based on dim and shape
+    if dim == 1:
+        return shape[0], 1, 1
+    elif dim == 2:
+        return shape[1], shape[0], 1
+    else:  # dim == 3:
+        return shape[2], shape[1], shape[0]
+
+
+# get example image, add alpha channel of all ones
+image = iio.imread("imageio:astronaut.png")
+image_rgba = np.zeros((*image.shape[:-1], 4), dtype=np.uint8)
+image_rgba[..., :-1] = image
+image_rgba[..., -1] = 255
+
+# wgpu texture size is (width, height) instead of (rows, cols) for whatever reason
+rgba_size = size_from_array(image_rgba, dim=2)
+
+# output
+y_size = size_from_array(image_rgba[:, :, 0], dim=2)
+cbcr_size = size_from_array(image_rgba[::2, ::2, 0], dim=2)
+
+# create device
+device: wgpu.GPUDevice = wgpu.utils.get_default_device()
+
+# create texture for input rgba image
+texture_rgb = device.create_texture(
+    label="rgba",
+    size=rgba_size,
+    usage=wgpu.TextureUsage.COPY_DST | wgpu.TextureUsage.TEXTURE_BINDING,
+    dimension=wgpu.TextureDimension.d2,
+    format=wgpu.TextureFormat.rgba8unorm,
+    mip_level_count=1,
+    sample_count=1,
+)
+
+# write input texture to device queue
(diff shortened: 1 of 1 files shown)


Korijn:
> 2\. Writing the workgroup_id value to the texture to make sure I've written my for loops correctly and structured my invocations correctly

This is pretty much the way. You can write unit tests on the intermediate outputs of your shaders.

kushalkolar:
> > 2. Writing the workgroup_id value to the texture to make sure I've written my for loops correctly and structured my invocations correctly
> 
> This is pretty much the way. You can write unit tests on the intermediate outputs of your shaders.

So basically write global invocation ids, workgroup ids, iterator values, etc. to a buffer and validate? This is almost like writing assembly but not as tedious and more fun!

almarklein:
I indeed tend to encode values that I want to check into the output color 🤷  I like the idea of writing values to a buffer, then loading that buffer to CPU, and rendering that as text! 🚀 

What would really be great, is a tool that can "emulate" wgsl so you can run it, set breakpoints, inspect variables, etc. Not sure if something like this exists ...

Vipitis:
There are tools that let you step through a shader pixel like Renderdoc, iGPA (likely also Nsight and Radeon analyzer...) but that's more tedious than trying to get the output onto the screen has a readable color or buffer of values that can be copied to CPU and printed. 

For prototyping compute shaders, maybe use compute.toys

Korijn (approved):

almarklein (approved):

almarklein:
```wgsl
    // chroma subsampling
    for (var x: u32 = start.x; x < stop.x; x += 2) {
        for (var y: u32 = start.y; y < stop.y; y += 2) {
            // convert to normalized uv coords for sampler
            let coords_sample: vec2f = (vec2f(f32(x), f32(y)) + 0.5) / vec2f(textureDimensions(tex_rgba).xy);
            var px_sample: vec4f = textureSampleLevel(tex_rgba, chroma_sampler, coords_sample, 0.0);
    ...
```

I've found that doing a texture sample in a loop is considerably slower than just writing out the `textureSampleLevel` calls below each-other, because it allows the compiler to optimize/combine the texture fetches. Just noting this for future optimizations.

kushalkolar:
> ```wgsl
>     // chroma subsampling
>     for (var x: u32 = start.x; x < stop.x; x += 2) {
>         for (var y: u32 = start.y; y < stop.y; y += 2) {
>             // convert to normalized uv coords for sampler
>             let coords_sample: vec2f = (vec2f(f32(x), f32(y)) + 0.5) / vec2f(textureDimensions(tex_rgba).xy);
>             var px_sample: vec4f = textureSampleLevel(tex_rgba, chroma_sampler, coords_sample, 0.0);
>     ...
> ```
> 
> I've found that doing a texture sample in a loop is considerably slower than just writing out the `textureSampleLevel` calls below each-other, because it allows the compiler to optimize/combine the texture fetches. Just noting this for future optimizations.

Thanks! So the wgsl compiler doesn't do loop unrolling?! Are optimizations documented anywhere or is this the early era and you just know from trying :sweat_smile: 

So should this be written as:

```wgsl
for x .... {
    textureSampleLevel(y1...)
    textureSampleLevel(y2...)
    textureSampleLevel(y3...)
    textureSampleLevel(yn...)
```

Or unroll _the entire thing_ for that workgroup?

Korijn:
You could use jinja to save yourself some typing :)

almarklein:
Unroll the entire thing :sweat_smile:   I know from running into extreme performance differences for two shaders that do almost the same thing, and going down the rabbit hole.  Many shading languages can mark a loop so the compiler unrolls it, but wgsl does not have that yet.

From my understanding, in a loop, the compiler sees *hey, I need to do a texture fetch, let's do that*, and then thus does that for every iteration. If the loop is unrolled, it sees *hey, I need to do N texture fetches, let's combine these in the most optimal way*. IIUC the same applies to an `if`.

almarklein:
While we're on this topic, another thing I learned the hard way is the limited amount of fast memory. The fast memory (registers) on most modern hardware fit about 128 f32's. So a vec4f takes four of those. Any variable/array/struct contributes to this (but constants and literals generally do not). 

In my case, I did some 9-25 fetches and stored the colors in an array, so I could use them in different ways in the shader.
But this made it much slower (like 100x).  So you may want to avoid storing the result of N texture fetches, and instead add their contribution to some aggregated value. Or only store the chrome values instead of the rgba.

PR #719, opened 2025-06-01
Adding missing enums to FeatureName and CanvasAlphaMode.

anadodik:
Hi friends,

Please let me know if I am missing something in this PR! I tried going through the how-to-contribute docs but it's highly likely I missed something.

I ran into an issue where `unpremultiplied` was the only canvas alpha mode available on my Mac, but the  check on line 350 of `_classes.py` inside of `GPUCanvasConfigure` would fail. Adding this to the enum fixes it and I am able to use alpha blending.

I also had to use `texture-adapter-specific-format-features` to enable multisampling on f32 textures. Using the string works, but I figured I'd add it too since I am changing things.

Looking forward to feedback!

Best,
Ana

### DIFF
--- a/wgpu/enums.py
+++ b/wgpu/enums.py
@@ -81,6 +81,9 @@ class FeatureName(Enum):
     clip_distances = "clip-distances"
     dual_source_blending = "dual-source-blending"
     subgroups = "subgroups"
+    texture_adapter_specific_format_features = (
+        "texture-adapter-specific-format-features"
+    )
 
 
 class BufferMapState(Enum):
@@ -406,6 +409,7 @@ class QueryType(Enum):
 class CanvasAlphaMode(Enum):
     opaque = "opaque"
     premultiplied = "premultiplied"
+    unpremultiplied = "unpremultiplied"
 
 
 class CanvasToneMappingMode(Enum):
--- a/wgpu/resources/codegen_report.md
+++ b/wgpu/resources/codegen_report.md
@@ -23,6 +23,7 @@
 * Validated 37 classes, 124 methods, 0 properties
 ## Validating backends/wgpu_native/_api.py
 * Enum field FeatureName.subgroups missing in wgpu.h
+* Enum field FeatureName.texture-adapter-specific-format-features missing in wgpu.h
 * Enum PipelineErrorReason missing in wgpu.h
 * Enum AutoLayoutMode missing in wgpu.h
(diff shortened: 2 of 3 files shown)


Vipitis:
hey,

I had a look and I believe the issue might actually be related to this line https://github.com/pygfx/wgpu-py/blob/d25edb102644f74c0d12e24996f757eac77b57fa/wgpu/backends/wgpu_native/_api.py#L790 perhaps we shouldn't just check against the `CanvasAlphaModes` in `_classes`.

Can you check what capabilities your canvas/adapter return?
```py
import wgpu
from wgpu.gui.auto import WgpuCanvas

canvas = WgpuCanvas()
adapter = wgpu.gpu.request_adapter_sync()
context = canvas.get_context("wgpu")
capabilities = context._get_capabilities(adapter)

print(capabilities)
```

as for the changes. .idl is an external file - so it can't be edited. Differences are marked in `._classes.py` with `@apidiff`


`texture-adapter-specific-format-features` is a native only feature, so it will not be part of the .idl and instead is from the external file `wgpu.h`. It does exist in `_mappings.py` under "NativeFeature" so should be requestable from there.

Vipitis:
I am more wondering how you get a difference between `CompositeAlphaMode` and `CanvasAlphaMode` in the first place, I suspect it is a screen vs off screen thing. Maybe there is some discussion upstream that gives insight. Or it could be specifically related to the native feature needing it. 

Alpha blending is one of these cases where you get undefined behavior by some GPU vendors and even spec violations. So being a little more lenient might be the way to go.

almarklein:
> [...] 'alpha_modes': ['opaque', 'unpremultiplied'],

Oh, glad to see 'opqaue' is in there 😅  I initially misread that you meant 'unpremultiplied' was the only one.

To provide some context on API 'discoverability'; the idea is that we stick to the WebGPU spec for the main API, and users can include native-specific features from `wgpu.backends.wgpu_native`. But this case is a little different, because (apparently) the windowing layer may not always support the alpa modes that WebGPU specified (opaque and premultiplied). 

> perhaps we can find a way to include all native only flags and enums without needing to manually whitelist them there. But some of the names are nearly the same and could be confusing.

Yeah, also these cases are rare, and tend to be edge-cases that are each special in their own way, so I guess just local whitelisting is the easiest way 🤷 

> This works on my machine too---happy to change the diff to this if y'all say it's the way to go.

Thanks for testing. That would be great!

PR #735, opened 2025-09-04
Remove wgpu.gui canvases

almarklein:
* [x] Users should use `rendercanvas` instead.
* [x] For now, doing `from wgpu.gui.auto import WgpuCanvas` will just provide a `RenderCanvas`.
* [x] Same for `wgpu.gui.WgpuCanvasBase`, because PyGfx still imports that.
* [x] Update tests, Remove from docs, remove more references etc.
* [x] Restore `WgpuCanvasInterface`.

### DIFF
--- a/examples/compute_int64.py
+++ b/examples/compute_int64.py
@@ -1,0 +1,70 @@
+"""
+simple example of using the int64 shader feature
+"""
+
+import wgpu
+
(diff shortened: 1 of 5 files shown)


almarklein:
@Korijn  let's merge this and then do a release.

Korijn (approved):

Korijn (approved):

PR #737, opened 2025-09-09
Drop support for Python 3.9

almarklein:
Makes typing a bit easier, and is EOL a month from now.


Vipitis:
would it make sense to have a release before dropping 3.9 so the final release at least includes the imgui fixes?

almarklein:
Sure, we can do that.

Korijn (approved):

PR #738, opened 2025-09-10
Switch to trusted publishing for PyPI publish in CI

### LINKED ISSUE #734: Switch to trusted publishing for package upload to PyPI in CI
Trusted publishing (with attestations) means I can know for certain that what I download from PyPI is the same artefact which was generated in GitHub CI, meaning that what I see in GitHub is the same as what is installed - handy for auditing (rather than having to manually review all of the installed files on each release).

See [the Python packaging documentation](https://packaging.python.org/en/latest/guides/publishing-package-distribution-releases-using-github-actions-ci-cd-workflows/#configuring-trusted-publishing), [the PyPI documentation](https://docs.pypi.org/trusted-publishers/), and [the official pypi-publish GitHub action documentation](https://github.com/pypa/gh-action-pypi-publish?tab=readme-ov-file#trusted-publishing) on trusted publishing - you'll need to configure an environment in PyPI and GitHub. You will be able to remove the `PYPI_PASSWORD` project secret.

Should be as simple as removing `user` and `password` arguments to the `Publish to PyPI` step of the `publish` job of the [`CD` workflow](/pygfx/wgpu-py/blob/main/.github/workflows/cd.yml), and adding environment and permissions to that job, in terms of code changes.

Related: pygfx/pygfx#1175

EpicWink:
You'll need to create a new GitHub environment called `pypi` ([GitHub docs](https://docs.github.com/en/actions/how-tos/deploy/configure-and-manage-deployments/manage-environments#creating-an-environment)), and add it to PyPI ([PyPI docs](https://docs.pypi.org/trusted-publishers/adding-a-publisher/#github-actions)).

Also enables computing and displaying published files' hashes.

Resolves #734

Related: pygfx/pygfx#1192

### DIFF
--- a/.github/workflows/cd.yml
+++ b/.github/workflows/cd.yml
@@ -101,6 +101,12 @@ jobs:
     runs-on: ubuntu-latest
     needs: [build-wheels, test-wheels]
     if: success() && startsWith(github.ref, 'refs/tags/v')
+    environment:
+      name: pypi
+      url: https://pypi.org/p/wgpu
+    permissions:
+      contents: write
+      id-token: write
     steps:
     - uses: actions/checkout@v4
     - name: Set up Python
@@ -134,5 +140,4 @@ jobs:
     - name: Publish to PyPI
       uses: pypa/gh-action-pypi-publish@release/v1
       with:
-        user: __token__
-        password: ${{ secrets.PYPI_PASSWORD }}
+        print-hash: true


almarklein (approved):
Thanks! LGTM and I made the necessary changes in the repo and on pypi.

almarklein:
Are you also going to do `rendercanvas`? 😉

Korijn (approved):

PR #739, opened 2025-09-11
Promise / Future API with .then()

### LINKED ISSUE #683: Support for callbacks in async functions
For the async API functions in wgpu, we currently offer two options: busy-wait for it, or async-wait for it.

In JS, an async function returns a promise, and this offers a 3d option: to provide a callback that is called when the promise resolves. I had completely overlooked this option. It's interesting, because its an option that works in sync (not-async) code, and can be made to work the same way in Python and JS/Pyodide.

I probably need to think about this more, also how this plays with rendercanvas and pygfx, but what I currently think would be a good plan is to have all the async API functions return a Promise-like object, which can then be used to:
* `promise.busy_wait()` - Python only, for code that does not care waiting a bit
* `await promise` - works in Python and JS
* `promise.then(callback)` - Python and JS

(we already have an object that does the first two things, but it's used internally (not public) and called `WgpuAwaitable`)

almarklein:
Closes #683, ref #761

Adds `GPUPromise` that is returned by all async methods. It handles all asynchronous logic in single place.

## Context

The idea is that the async methods of the wgpu API will no longer have `async` in front, but instead return the promise object. This object can be awaited, sync waited, and also handled with a callback.

The callback-approach is new, and makes it possible to handle things asynchronously when not in async code. This opens up doors for friedlier API, improved performance, and a more sync workflow that's still portable. It will e.g. allow making the rendercanvas "bitmap"  present-mode more performant, and allow us to resolve some issues related to timing events and draws.

Example of intended use:
```py
>>> p = wgpu.gpu.request_adapter_async()
>>> p
<GPUPromise pending  at 0x107ead400>

>>> await p:
<wgpu.backends.wgpu_native.GPUAdapter object at 0x107eaeba0>

>>> p.sync_wait()
<wgpu.backends.wgpu_native.GPUAdapter object at 0x107eaeba0>

>>> p.then(lambda adapter: print("got", adapter))
got <wgpu.backends.wgpu_native.GPUAdapter object at 0x107eaeba0>
```

## Polling and event loop

The promise objects have an optional `loop` and `poller` argument. If the poller is given, the `promise.async_wait()` method can poll until the promise resolves. And `await promise` can do the same, like we already do in `main`.

But for `promise.then()` to work, we must be able to schedule a new call with the event loop, i.e. call `loop.call_soon()`. If loop is given and the poller is not, `await promise` also works. 

In the browser backed, the `loop` still exists (asyncio has one main loop in pyodide). But we cannot poll, and we don't have to, so this is fine.

One caveat is that (on wgpu-native) we must somehow poll wgpu to process its internal events. This can be at spefic moments in the event loop, using a separate async tasks, or even in a separate thread. The tricky part is to come up with a strategy that does not strain the CPU, but also does not cause unnecessary delays.

## Code changes

* [x] Update codegen to allow adding classes using the apidiff mechanism (i.e. adding classes that are not in WebGPU).
* [x] Implement  `GPUPromise` in  new `_async.py`. 
* [x] Subclass `GPUPromise` in `_classes.py` so it's part of the formal API, and can be subclassed in backends.
* [x] Remove the private `WgpuAwaitable` class that we used for the same purpose.
* [x] All async methods like `request_adapter_async` return a `GPUPromise`. They no longer have the `async` keyword, because the methods themselves are no longer async.
* [x] All async methods in `_api.py` are adjusted to use `GPUPromise` instead of `WgpuAwaitable`.
* [x] `request_adapter_async` has a `loop` argument so that the promises can use `call_later`. 
  * [x] It is optional. Async code using `await` still works, using the same mechanism we've used  until now.
  * [x] Also optional for synchronous use.
  * [x] Required when using `.then()`.
* [x] Revive `backoff_time_generator`

### DIFF
--- a/wgpu/_classes.py
+++ b/wgpu/_classes.py
@@ -1157,12 +1157,14 @@ class GPUDevice(GPUObjectBase):
             vertex (structs.VertexState): Describes the vertex shader entry point of the
                 pipeline and its input buffer layouts.
             primitive (structs.PrimitiveState): Describes the primitive-related properties
-                of the pipeline. If `strip_index_format` is present (which means the
+                of the pipeline. If ``strip_index_format`` is present (which means the
                 primitive topology is a strip), and the drawCall is indexed, the
                 vertex index list is split into sub-lists using the maximum value of this
                 index format as a separator. Example: a list with values
-                `[1, 2, 65535, 4, 5, 6]` of type "uint16" will be split in sub-lists
-                `[1, 2]` and `[4, 5, 6]`.
+                ``[1, 2, 65535, 4, 5, 6]`` of type "uint16" will be split in sub-lists
+                ``[1, 2]`` and ``[4, 5, 6]``.
+                The primitive state supports an additional field ``polygon_mode`` that can be set to "line" or "point". To do so, the primitive state must be given as a dict,
+                and the the device must be requested with the native-only features "polygon-mode-line" or "polygon-mode-point".
             depth_stencil (structs.DepthStencilState): Describes the optional depth-stencil
                 properties, including the testing, operations, and bias. Optional.
             multisample (structs.MultisampleState): Describes the multi-sampling properties of the pipeline.
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -170,7 +170,7 @@ jobs:
             pyversion: '3.13'
           - name: Test Linux pypy3
             os: ubuntu-latest
-            pyversion: 'pypy3.10'
+            pyversion: 'pypy3.11'
     steps:
     - uses: actions/checkout@v4
     - name: Set up Python ${{ matrix.pyversion }}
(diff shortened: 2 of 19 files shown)


wgpu/_classes.py
```diff
@@ -553,6 +554,125 @@ def _release(self):
...
+        raise NotImplementedError()
+
+    def then(self, callback: Callable[Any, None]):
+        """Set a callback that will be called when the future resolves.
+
+        The callback will receive one argument: the result of the future.
+        """
+        # TODO: allow calling multiple times
+        # TODO: allow calling after being resolved -> tests!
+        # TODO: return another promise, so we can do chaining? Or maybe not interesting for this use-case...
```
Korijn, line 664: This seems to be pretty crucial to me!

Korijn (approved):
I'n a bit skeptical about the loop requirement (have to see how it pans out) but otherwise it seems well designed!

PR #744, opened 2025-09-15
Cleanup examples (remove most gui-specific ones)

almarklein:
The examples to demo all the various GUI toolkits etc. are covered in the docs of rendercanvas. I kept the one for embedding Qt in an application, but removed most of the rest.


Vipitis:
the subprocess example was a fun one to figure out. I don't think it has much utility but it might a be a good reference for how to do a custom canvas. Should I bring that back in `rendercanvas` and explain how you can communicate the relevant id via pipe?

almarklein:
> the subprocess example was a fun one to figure out.

I'm not even sure what the current status of that example was. I don't think it's even supposed to work 😄. But I agree that it demonstrated an interesting use-case. Maybe we can change it a bit so it actually works ... first thought: running the canvas in a subinterpreter! That way the process is the same, so it should "valid". And yeah, rendercanvas is likely the better place for an example like that.

PR #748, opened 2025-09-18
Better support for descriptors by using dataclasses 

### LINKED ISSUE #425: Allow using structs as API
We define structs in `wgpu.structs`, and use them as follows:
* In the wgpu-native backend, we use it to check incoming dicts.
* We generate docs off them, that users can use as a reference how to construct their dicts.

But they are not really part of the user-facing API, and (other than flags and enums)  they are not injected into the root namespace.

I propose that we make the objects in `wgpu.structs` callable, so user can type e.g. `wgpu.BindGroupEntry(` and let autocompletion kick in to show the possible arguments. The call returns a normal Python dict.

In an interactive shell you can also call such an object, get the dict repr-ed in the console, and copy-paste that into your code.

In summary it makes our API a bit more discoverable.

BTW It's probably better to generate these function's code (as opposed to creating the functions dynamically at startup) so that they can be statically discovered and used (easier) by IDE's to show signatures to the user.

almarklein:
Supersedes #746, closes #425

This PR defines the wgpu structs as dataclasses. This means that if a user uses the struct classes to create descriptors etc. they fully benefit from autocompletion, type checking, interactive docs etc. Further the backend api can check the incoming struct using isinstance instead of checking for the presence of fields.

Dicts are still supported, so this change should not break existing code. I think we should promote the new way to write structs. We could eventually deprecate using dicts, or not, let's decide later.

 * 🚫 In an interactive session, see the options by typing `>> wgpu.SomeStruct` and hitting enter.
 * 🆗 In an interactive session, get autocompletion by typing `>> wgpu.SomeStruct.x`.
 * 🆗 In an interactive session, get autocompletion by typing `>> wgpu.SomeStruct(foo`.
 * 🆗 Static type checking.
 * 🆗  Static autocompletion.
 * 🆗  Generation of clear API docs.

### Example from cube.py:

This used to be a nested dict. Now it's a nested class, but see how they are named, so it's a bit self-documenting.
In IDE's like VSCode, writing code like this is now quite a pleasant experience.

```py
return wgpu.RenderPipelineDescriptor(
        layout=pipeline_layout,
        vertex=wgpu.VertexState(
            module=shader,
            entry_point="vs_main",
            buffers=[
                wgpu.VertexBufferLayout(
                    array_stride=4 * 6,
                    step_mode="vertex",
                    attributes=[
                        wgpu.VertexAttribute(
                            format="float32x4",
                            offset=0,
                            shader_location=0,
                        ),
                        wgpu.VertexAttribute(
                            format="float32x2",
                            offset=4 * 4,
                            shader_location=1,
                        ),
                    ],
                ),
            ],
        ),
        primitive=wgpu.PrimitiveState(
            topology="triangle-list",
            front_face="ccw",
            cull_mode="back",
        ),
        fragment=wgpu.FragmentState(
            module=shader,
            entry_point="fs_main",
            targets=[
                wgpu.ColorTargetState(
                    format=render_texture_format,
                    blend={"alpha": {}, "color": {}},
                )
            ],
        ),
    )
```


almarklein:
@Korijn en @Vipitis curious to your thoughts. This makes wgpu code feel a bit less "light-Python", because you're instantiating (light) classes instead of dicts, but IMO it makes code more readable, and makes the experience with autocompletion and typing much more pleasant. 

The approach in #746 (the one based on TypedDict) also comes a long way: using dicts, but also having type checking. But I found that in nested dicts it does not work so well (e.g. autocompletion does not work).

wgpu/structs.py
```diff
@@ -1,27 +1,45 @@
...
+    def __len__(self):
+        return len(self.__dict__)
+
+    def get(self, key, default=None):
+        """Get method that returns the default when the acual value is None."""
+        # Get the stored value. Note that we don't allow unknown/invalid keys
+        val = self.__dict__[key]
+        # None means default
+        if val is None:
+            return default
```
Korijn, line 41: Are you sure about this? What if you explicitly need to set None, overriding a default?

almarklein: Good point. None means "use the default". For this to work, we need to be able to assume that there are no cases where None is a valid value and the default is something else. I'm pretty sure there are no such cases. I'm thinking of a way to check this ... 🤔

almarklein: All values in the "leafs" of wgpu are either:
* A GPUObject
* A struct
* A list
* An enum
* A flag
* An int
* A float

The only cases where None could be a valid value is in the top two. And in those cases None would also be the default.

Speaking in C-API terms, Null can only be valid for attributes that are pointers, and when it is allowed to use a null-pointer, that is the default.

Vipitis: Here is the diff for the webgpu.h repo of what we are using up to main: https://github.com/webgpu-native/webgpu-headers/compare/bac520839ff5ed2e2b648ed540bd9ec45edbccbc...main#diff-074fca0467717d8d789bdfb86bf9e2d4afec03055a2787e0b604fda4254d4ed9

And here is the WebGPU status talk from last month: https://youtu.be/gdMtgJTe1qY and the Face2Face meeting from this week: https://docs.google.com/document/d/1WnEj8jwUxgEeoMhUf5AuSoGoaHsZAAGmEk_P8ktlzbY

saying that `webgpu.h` is stable but `wgpu-native` being delayed and in need for contributors (plus a bunch of other interesting stuff if you are curious about what's happening upstream).

Korijn: @almarklein is this resolved?

almarklein: Yes, I'm confident that we can safely use `None` to mean default.

Issue #729, opened 2025-07-28
Way to display images(imgui.image) that are already on GPU


REDxEYE:
Hello, I haven't found a way to display image/texture that i have created in pygfx in wgpu-py/imgui-bundle.

Currently i'm manually adding my texture to ImguiWgpuBackend._texture_views like so
```py
    def _get_imgui_image_id(self, texture: gfx.Texture) -> int:
        """Get the ImGui image ID for the given texture."""
        if self._current_texture_id is None or self._current_texture_id != id(texture):
            self._current_texture_id = id(texture)
            texture_view = GfxTextureView(texture)
            self._current_texture_view = ensure_wgpu_object(texture_view)
            id_32 = ctypes.c_int32(id(self._current_texture_view)).value
            self.window.imgui_renderer.backend._texture_views[id_32] = self._current_texture_view
            return id_32
        return ctypes.c_int32(id(self._current_texture_view)).value
```

Please help me find correct way to display images

Vipitis:
hey, can you share an example where you use this?

If by display you mean drawing it to the screen, then you should be able to use `copy_texture_to_texture` where the target is whatever `context.get_current_texture()` gives you for that renderpass. Note that this will require you to have the usage `COPY_DST` which you can request in `context.configure`.

If instead I read it incorrectly and you want to use the imgui.Image widget, there is `.register_texture` which is similar to what you are doing in the code above: https://github.com/pygfx/wgpu-py/blob/8037a00404be1fa9199ce078de3bee5c0c5fab8f/wgpu/utils/imgui/imgui_backend.py#L104-L113

REDxEYE:
Yes, i meant use it in imgui.image. Looks like in version i'm using this method is not yet present

REDxEYE:
Also, will it be possible to use imgui window as 3d viewport/canvas?

Vipitis:
hey, I just had a little try for my imgui implementation and looks like you can easily put the past texture into the image widget, see https://github.com/pygfx/shadertoy/commit/a9656425eabdabea89e71136d4b0129e7c548fa6

<img width="849" height="747" alt="Image" src="https://github.com/user-attachments/assets/525c2e36-3780-4bd4-991c-c163eaab77a5" />

is this what you were looking for or something else? I think you could technically make a rendercanvas class to display inside this image widget - but not sure if that has any use.

REDxEYE:
Well, can you please also provide unregister method? I need resizable viewport and there is no way to unregister old texture

PR #749, opened 2025-09-24
Add unregister texture method

panxinmiao:
Implement https://github.com/pygfx/wgpu-py/issues/729#issuecomment-3293497600 and address https://github.com/pygfx/pygfx/issues/1197https://github.com/pygfx/pygfx/issues/1197.

BTW: I think the textures used by ImGui should be a fixed and limited set (corresponding to its GUI components). During the program’s runtime, these GPU texture objects generally should not be replaced—only their texture data needs to be updated. Therefore, it doesn’t seem necessary to provide a method for unbinding texture resources. When the ImGui backend is destroyed, the references to these resources will be naturally released. 
That said, it seems there are some special cases where this mechanism is useful, which is why this method was added.

### DIFF
--- a/wgpu/utils/imgui/imgui_backend.py
+++ b/wgpu/utils/imgui/imgui_backend.py
@@ -89,7 +89,6 @@ class ImguiWgpuBackend:
         # Texture management
         self._textures = {}
         self._texture_views = {}
-        self._texture_bind_groups = {}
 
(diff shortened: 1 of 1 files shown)


almarklein (approved):
Thanks Pan! I can confirm that this fixes https://github.com/pygfx/pygfx/issues/1197

PR #751, opened 2025-09-25
Remove mutable defaults

### LINKED ISSUE #747: Avoiding mutable defaults?
more like a discussion. 

`ruff check --select B006` is one of the rules we ignore and it currently finds 38 violations, a lot of them in `_classes.py` or with `wgpu.structs` arguments.
It's one of the python quirks that can be confusing and lead to hard to debug bugs: https://docs.astral.sh/ruff/rules/mutable-argument-default/
Not a fan of the recommended solution since it looks so awkward. 

I don't think this has ever happend to me and might not even be a problem with how args are used. 
Seeing all the typing improvements being made to typing and structs, perhaps this is worth a consideration.

almarklein:
Closes #747 

* [x] Default arg value of `[]` -> `()`, and the `list[..]` type becomes `Sequence[..]` to allow both lists and tuples.
* [x] Default arg value of `{}` -> `None`, and in function `val = {} if val is None else val`.

### DIFF
--- a/codegen/apiwriter.py
+++ b/codegen/apiwriter.py
@@ -124,6 +124,7 @@ def write_structs():
             # Get type to include in code
             raw_type = idl.resolve_type(field.typename).strip("'")
             code_type = raw_type.replace("GPU", "classes.GPU").replace("structs.", "")
+            code_type = code_type.replace("list[", "Sequence[")
             # Get type to include in docs
             docs_type = None
             if "flags." in raw_type or "enums" in raw_type:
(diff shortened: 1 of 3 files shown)


Korijn (approved):

wgpu/_classes.py
```diff
@@ -309,9 +310,9 @@ def configure(
         device: GPUDevice,
         format: enums.TextureFormatEnum,
         usage: flags.TextureUsageFlags = 0x10,
-        view_formats: list[enums.TextureFormatEnum] = [],
+        view_formats: Sequence[enums.TextureFormatEnum] = (),
         color_space: str = "srgb",
-        tone_mapping: structs.CanvasToneMappingStruct = {},
+        tone_mapping: structs.CanvasToneMappingStruct | None = None,
```
Vipitis, line 315: looks like this default breaks `configure_screen`, as it never gets replaced by `{}` and stays None:

`TypeError: Expecting CanvasToneMapping or dict, but got None`

PR #752, opened 2025-09-27
Handle pixel ratio in resize event for imgui adapter

hmaarrfk:
Add support for pixel ratio in resize event handling.


hmaarrfk:
oh i totally forgot about this....

many thx.

PR #760, opened 2025-10-04
add DXC and debug examples

Vipitis:
cleaner version of #728


almarklein (approved):
Nice work! Thanks for finishing this :)

PR #764, opened 2025-10-31
Canvas/context inversion of control

Korijn:
I know this branch is hideous and broken, but I wanted to put it up in draft anyway to communicate something that's bothered me about wgpu-py's API from the start. It's the hidden coupling between wgpu-py and the GUI layer. This self-referential class object thing just feels wrong. Below is an AI generated summary that I thought helps to illustrate my thoughts on this subject.

You can check out this branch and run the `gui_direct.py` example to see that it works (including resizes).

My main intention of opening this draft PR is to initiate a discussion with you @almarklein :)

In my opinion it's really not worth it to litter the codebase with class to get_context via a canvas weakref via a wrapper object, just to be able to call get_physical_size, _and_ require all users of wgpu-py to work with this awkward class interface. 

I strongly believe that implementing this decoupling API change will make wgpu-py much easier to apply in a wide variety of applications using libs such as glfw, sdl and so on.

## Summary
Replace the implicit self-referential canvas/context wrapper with a plain, caller-owned GPUCanvasContext. The application is now responsible for window/canvas lifetime and must explicitly supply canvas-related state (for example, physical framebuffer size) to the context.

## What changed
GPUCanvasContext no longer depends on being wrapped by a canvas object that calls back into itself.
Applications may construct a GPUCanvasContext directly and must call `context.set_physical_size((w, h))` and update it on resizes.

Example: `gui_direct.py` creates GPUCanvasContext directly and calls `set_physical_size()` from GLFW frame-buffer queries.

## Rationale
* Inverts control: the application manages window/canvas state instead of the context pulling it implicitly.
* Simplifies interop with GUI toolkits where the app already owns the window.
* Makes responsibilities explicit and reduces hidden coupling between GUI and wgpu-py.

## Example (conceptual)
* Before: canvas -> canvas.get_context("wgpu") -> context pulled size callbacks
* After: app owns window; app creates context; app calls context.set_physical_size(...) on resize.

### DIFF
--- a/wgpu/resources/codegen_report.md
+++ b/wgpu/resources/codegen_report.md
@@ -19,7 +19,7 @@
 * Diffs for GPUTextureView: add size, add texture
 * Diffs for GPUBindingCommandsMixin: change set_bind_group
 * Diffs for GPUQueue: add read_buffer, add read_texture, hide copy_external_image_to_texture
-* Validated 38 classes, 121 methods, 49 properties
+* Validated 38 classes, 120 methods, 50 properties
 ### Patching API for backends/wgpu_native/_api.py
 * Validated 38 classes, 115 methods, 0 properties
 ## Validating backends/wgpu_native/_api.py


almarklein:
Thanks @Korijn!

Executive summary: I agree!

I should start by saying that I share your feeling that the context feels somewhat awkward. Although for me the confusion lies more in how it ties `rendercanvas` and `wgpu-py` together, in a way that does not feel quite right yet ...

Have you seen https://github.com/pygfx/rendercanvas/issues/124? In short, it proposes that rendercanvas implements its own context class, and (only if necessary) wraps a `wgpu.GPUCanvasContext`. It fits quite well with your idea, in the sense that `rendercanvas` and `wgpu-py` are more independent, and can each decide on their own API for the the context (or even drop the concept of a context?).

The main reason why our `GPUCanvasContext` is the way it is, is because it follows the JS/WebGPU `GPUCanvasContext` API. But as I am composing this reaction, I've come to realize 💡 that this API has quite some historic influence, and is geared to the situation in the browser. And we can certainly ask ourselves whether its worth to strictly follow that API. Spoiler alert: I now think its probably not.

To go in a bit more detail ... in JS, both the `CanvasRenderingContext2D` and the `WebGLRenderingContext` are objects that have *a lot* of methods, which are used by user-code to perform the rendering. These objects get passed around a lot, and it's very convenient for such code to then also be able to access the canvas. The `GPUCanvasContext` was likely given a `.canvas` method mostly for compatibility with these other contexts; it has only a few methods which are used in a very specific places. Further, the only canvas in a browser is a `<canvas>` which is owned by the browser and a rather opaque object for the user. In contrast, in Python the "canvas" can be *many* different things, and is managed outside of wgpu-py. 

With that said, I think we can justify adopting a different API for wgpu-py. One that makes wgpu-py oblivious of any canvas, and that makes code outside of wgpu-py responsible for keeping the context up-to-date. That makes cases like `gui_direct.py` simpler, and `rendercanvas` can just abstract such boilerplate away for the user. 🚀 

A few more practical comments:
* I think we can drop the `get_physical_size()` since the code that handles the context knows the physical size, or calls `get_current_texture()` and checks its size. The context needs to know the physical size internally because it's needed to (re) configure the context.
* https://github.com/pygfx/rendercanvas/issues/124 means that the `wgpu.GPUCanvasContext` will only be used for presenting to screen, so the `present_methods` dict will be replaced with a dict that contains a smaller dict that conains the platform and winid (which we use to get the surface id).
* The one other thing that the context gets from the canvas is the `_vsync` prop. We can add that to the aforementioned present-dict.
* I'll think about the best order in which to apply these changes 🤔

wgpu/_classes.py
```diff
@@ -206,6 +206,13 @@ def wgsl_language_features(self) -> set:
         # Looks like at the time of writing there are no definitions for extensions yet
         return set()
 
+    @apidiff.add("Provide a low-level way to let wgpu render to screen")
+    def get_canvas_context(self, present_info: dict) -> GPUCanvasContext:
+        """Get the GPUCanvasContext object for the appropriate backend."""
+        from .backends.auto import gpu
+
+        return gpu.get_canvas_context(present_info)
```
Korijn: Can you explain what this is for? It's not apparent. Most likely my mental model of the wgpu object model is not accurate...

Korijn:
Semantically speaking, why is the class called GPUCanvasContext? Why isn't it just GPUContext? Is it in reference to a "canvas on the GPU"?

almarklein:
> Semantically speaking, why is the class called GPUCanvasContext? Why isn't it just GPUContext? Is it in reference to a "canvas on the GPU"?

That's just what the WebGPU spec calls it. In JS it makes perfect sense. For wgpu-py we try to ban the word 'canvas' from the codebase 😉, and we change its API somewhat, but it's still the same concept that we're exposing, so I'd prefer to keep the name the same as in the spec.

examples/cube.py
```diff
@@ -5,7 +5,7 @@
 because it adds buffers and textures.
 
 This example is set up so it can be run with any canvas. Running this file
-as a script will use the auto-backend.
+as a script will rendercanvas with the auto-backend.
```
Vipitis:
```suggestion
as a script will use rendercanvas with the auto-backend.
```

almarklein:
> I think this might allow wgpu to run with pyodide even without rendercanvas (minus the events). I could give that a try on the PR branch.

Definitely! Actually, I should probably just document what should be in the struct already.

almarklein (approved):

PR #768, opened 2025-11-10
Fix compatibility with rendercanvas 2.4.0

hmaarrfk:
I think there might be more places but now my application runs again

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -25,6 +25,7 @@ os.environ["RENDERCANVAS_FORCE_OFFSCREEN"] = "true"
 # Load wgpu so autodoc can query docstrings
 import wgpu  # noqa: E402
 import wgpu.utils.compute  # noqa: E402
+import wgpu.utils.glfw_present_info  # noqa: E402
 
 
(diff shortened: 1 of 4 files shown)


wgpu/utils/imgui/imgui_renderer.py
```diff
@@ -144,8 +144,8 @@ def render(self):
 
         imgui.set_current_context(self.imgui_context)
 
-        pixel_ratio = self._canvas_context.canvas.get_pixel_ratio()
-        lsize = self._canvas_context.canvas.get_logical_size()
+        pixel_ratio = self._canvas_context.pixel_ratio
+        lsize = self._canvas_context.logical_size
```
hmaarrfk: we can certainly add an if statement here too. but i'm always hesitant to adding "if statements' in the runtime path. I would probably add it around the entire `def render` so that there would be 2 versions of the function.

almarklein:
I'll add a few more cases to your pr, if you don't mind :)

Korijn (approved):

PR #778, opened 2025-11-21
Use a polling thread per device

almarklein:
This PR makes each device create a thread for polling the native device to resolve the promises. This means there are no busy-loops; i.e. its less strain on the CPU while obtaining faster responses.

The implemented technique is heavily inspired by the implementation in the Servo browser (https://github.com/servo/servo/pull/32266). The Firefox browser will likely adopt a similar strategy soon (https://bugzilla.mozilla.org/show_bug.cgi?id=1870699).

When passing a loop instance, this work relies on `loop.call_soon_threadsafe` from https://github.com/pygfx/rendercanvas/pull/146.

### DIFF
--- a/wgpu/resources/codegen_report.md
+++ b/wgpu/resources/codegen_report.md
@@ -21,7 +21,7 @@
 * Diffs for GPUQueue: add read_buffer, add read_texture, hide copy_external_image_to_texture
 * Validated 38 classes, 120 methods, 50 properties
 ### Patching API for backends/wgpu_native/_api.py
-* Validated 38 classes, 116 methods, 0 properties
+* Validated 38 classes, 117 methods, 0 properties
 ## Validating backends/wgpu_native/_api.py
 * Enum field FeatureName.core-features-and-limits missing in webgpu.h/wgpu.h
 * Enum field FeatureName.subgroups missing in webgpu.h/wgpu.h


Korijn (dismissed):

wgpu/backends/wgpu_native/_api.py
```diff
@@ -564,24 +565,25 @@ def request_adapter_callback(status, result, c_message, _userdata1, _userdata2):
...
-        libf.wgpuInstanceRequestAdapter(get_wgpu_instance(), struct, callback_info)
+        libf.wgpuInstanceRequestAdapter(instance, struct, callback_info)
+
+        # Note that although we claim this is an asynchronous method, the callback
+        # happens within libf.wgpuInstanceRequestAdapter.
+        # To be sure though, we tickle the instance and double-check that the promise is set.
+
+        # H: void f(WGPUInstance instance)
+        libf.wgpuInstanceProcessEvents(instance)
+        assert promise._state != "pending"
```
Korijn: are you sure about the `assert` usage instead of raise?

almarklein:
I made a small improvement to the `_sync_wait()` logic.

I can already see that the FPS in https://github.com/pygfx/rendercanvas/pull/138  increases a bit because the system simply reacts immediately (instead of the previous loop that repeatedly sleeps and checks status). Also, while FPS is nice, the greater advantage IMO is minimizing the delay between rendering a frame and putting it under the user's eyes. This delay has a big impact on the perceived responsiveness.

PR #788, opened 2025-12-24
Fix error when the window is minimized

panxinmiao:
When the window is minimized, the application crashes.
```
  File "F:\rendercanvas\rendercanvas\_coreutils.py", line 50, in log_exception
    yield
  File "F:\rendercanvas\rendercanvas\_loop.py", line 241, in wrapper
    await async_func(*args)
  File "F:\rendercanvas\rendercanvas\_scheduler.py", line 164, in __scheduler_task
    canvas._process_events()
  File "F:\rendercanvas\rendercanvas\base.py", line 390, in _process_events
    self.__maybe_emit_resize_event()
  File "F:\rendercanvas\rendercanvas\base.py", line 361, in __maybe_emit_resize_event
    self._canvas_context._rc_set_size_dict(self._size_info)
  File "F:\rendercanvas\rendercanvas\contexts\basecontext.py", line 54, in _rc_set_size_dict
    self._wgpu_context.set_physical_size(*size_info["physical_size"])
  File "F:\wgpu-py\wgpu\_classes.py", line 277, in set_physical_size
    raise ValueError("Physical size values must be positive.")
ValueError: Physical size values must be positive.
```

I think `physical_size` can be (0, 0).

### DIFF
--- a/wgpu/_classes.py
+++ b/wgpu/_classes.py
@@ -273,8 +273,8 @@ class GPUCanvasContext:
         The application must call this to keep the context informed about
         the window/framebuffer size.
         """
-        if width <= 0 or height <= 0:
-            raise ValueError("Physical size values must be positive.")
+        if width < 0 or height < 0:
+            raise ValueError("Physical size values must be non-negative.")
         self._physical_size = int(width), int(height)
         self._has_new_size = True
 


Korijn (approved):
I checked with git blame, this PR indeed restores the original behavior. Thank you!

almarklein:
mmm, from what I can find, the original behavior was different from bitmap and screen presenting:

For bitmap, the case was basically ignored, making a texture of at least 1 pixel:
https://github.com/pygfx/wgpu-py/blob/44a8943363042b16b8006c5aa6df3bde3c55e26f/wgpu/_classes.py#L461-L462

For screen it failed also for zero-size textures:
https://github.com/pygfx/wgpu-py/blob/44a8943363042b16b8006c5aa6df3bde3c55e26f/wgpu/backends/wgpu_native/_api.py#L881-L884

I think the latter was done because zero-sized textures cause a crash later in the pipeline. Not sure if this behavior has changed in recent releases of wgpu-native.

---- 

Anyway ... thanks for the fix! I will probably dig a bit more into it though, because I feel I don't fully understand the consequences, and in what cases we can expect a zero-sized array.

I suspect that a zero-sized canvas can happen, but it will never be drawn to by rendercanvas, so the aforementioned pipeline error should not happen...

I just tried on MacOS, and for both glfw and Qt, a minimized window does not change its size. I can test on Windows and Linux next week when I am back in my office. Was the original issue on Windows @panxinmiao ?

panxinmiao:
>  Was the original issue on Windows @panxinmiao ?

Yes, I encountered this issue on Windows.

PR #804, opened 2026-04-06
Fix required positional args

### LINKED ISSUE #803: some required positional arguments default to `None`
Hey, 
noticed this in the signatures just now:
Here is one example (and this is all over `_classes.py` and the native implementation):
```py
# IDL: undefined clearBuffer( GPUBuffer buffer, optional GPUSize64 offset = 0, optional GPUSize64 size);
    def clear_buffer(
        self, buffer: GPUBuffer | None = None, offset: int = 0, size: int | None = None
    ) -> None:
```
the first argument isn't optional, the other two are. But All of them are ` ... | None = None`.

It's changed in this commit https://github.com/pygfx/wgpu-py/pull/748/changes/f7cd69ae53fad61ccf172d491c41ce3f4e2e2dfc couldn't find a reason in the review comments either. I believe it might be an oversight for methods that don't use a descriptor but positional arguments directly.

My debugging adventure so far got me from the changed lines in the commit above to `Attribute.required`, which I don't think is parsed correctly for required positional args (since it doesn't start with `"required"` in the IDL).

https://github.com/pygfx/wgpu-py/blob/0d0af75db48807ec2618179e49452e373639dddc/codegen/idlparser.py#L48-L52

I can try a PR for this tomorrow

Vipitis:
fixes #803 

I think this is a fix, but it is pretty awkward. The parsing logic for `Attribute` simply couldn't really distinguish between struct field (dict members) and positional args for a function. because the required <-> optional logic is different between the two.

Maybe there should be another class for `Parameter` instead?

I didn't check if any of the implementations of the native api need fixes, because some seem to validate defaults twice already [example](https://github.com/pygfx/wgpu-py/blob/0d0af75db48807ec2618179e49452e373639dddc/wgpu/backends/wgpu_native/_api.py#L2763-L2764) while the vast majority simply never handled an incorrectly passed `None`.

Python still won't necessarily raise anything more useful and the error comes from the native layer, unless you enforce typing.


almarklein:
> Maybe there should be another class for Parameter instead?

Since for descriptors we turn a struct into flat args, it makes sense (I think) to unify them.

I'm not sure if I was myself explicitly aware of how it works:
* in structs, fields are optional by default, unless marked as `required`.
* function parameters are required by default, unless marked as optional.

almarklein (approved):
Nice work, big improvement! I made some small tweaks to make the difference more explicit, and force `Attribute` to be instantiated with a `source` param.

Issue #789, opened 2025-12-29
Performance considerations: startup


Vipitis:
This might be an issue on my specific old system, but perhaps it's worth a little discussion.

Running any small example (or larger project) has a couple seconds startup time. I want to run some more profiling but it looks like creating the instance and requesting the adapter are two of the main slowdowns.

One easy speedup I identified is to only request one backend using instance extras, this already saves 1-2 seconds.
Will need to look more into this, as this will likely depend on platform a lot. Async might also help, but I feel most of the startup tasks depend on others to finish first.

(this isn't a perfect comparison, but matches the subjective experience too)

<img width="1033" height="523" alt="Image" src="https://github.com/user-attachments/assets/fc0c1d05-ac6a-4878-9ec8-c80bb8d8b34a" />

<img width="1033" height="479" alt="Image" src="https://github.com/user-attachments/assets/1400dbe3-be0b-4258-9157-87f23f756364" />

Vipitis:
I tried to do a benchmark comparison, but there is all kinds of caching going on. so to really compare a coldstart I need another evening to do like multiple scripts and do it end to end (request everything, compile a model, render a frame) to get a real world first frame latency.

PR #805, opened 2026-04-10
Update to wgpu-native v29.0.0.0 (includes updated headers!)

Vipitis:
as usual... still in progress but I am happy to get into this again and enjoy my free time.

There was a release for v27 with just the header updated, but there were quite a few missing symbols etc, so that release won't really work. So one PR to do both. The headers are also now considered stable. So this might be last major update for a long while.

I will put quite a few code comments in for things I feel like could be improved in the codegen, these are also meant for discussion.

more to come, I will try to keep the commits tidy and topic, since it's quite helpful to look at the individual diffs

### TODOs:
* [x] tests green
* [x] examples green
* [x] mem tests green
* [x] instance flags extras changed quite a bit
* [x] changelog explaining the breaking changes
* [ ] copy the issue template here?
* [x] try joined cleaned header
* [ ] the backend extras documentation needs to be updated 
    * [x] for immediates
    * [ ] the instance extras


wgpu/backends/wgpu_native/_ffi.py
```diff
@@ -27,7 +27,8 @@ def get_wgpu_header():
         get_header_filename("wgpu.h"),
     )
 
-
+# TODO: can we avoid doing this every single time the lib is loaded? I feel like we should provide a cleaned header file to ship instead?
```
Vipitis: this particular. It's probably a small amount, but if we cleanup the header file(like this function does, as part of codegen perhaps), combine both headers and save to to something like `wgpu-native-min.h` which can be read by `cdef()` directly?

I want to try give this a try, also makes the preprepreprocessing here more visible and easier to debug

almarklein: I like the idea to save work at runtime!

Vipitis:
I get test failure locally in `tests_mem/test_gui_qt.py` but CI seems to be happy. I am on rendercanvas main, and there were quite a few changes since. CI seems to have installed 2.6.3 so I guess it's fine. I suspect these test are outdated anyway.

wgpu/backends/wgpu_native/_ffi.py
Vipitis: I added a commit to give this a rough try. seems to still work - but didn't thoroughly test this. Without some kind of time to first frame benchmark (also for #789) hard to tell if it makes a meaningful difference. Either way it would avoid the prepreprocessor code duplication.

docs/backends.rst
```diff
@@ -61,100 +61,73 @@ The wgpu_native backend provides a few extra functionalities:
     :return: Device
     :rtype: wgpu.GPUDevice
 
-The wgpu_native backend provides support for push constants.
-Since WebGPU does not support this feature, documentation on its use is hard to find.
-A full explanation of push constants and its use in Vulkan can be found
-`here <https://vkguide.dev/docs/chapter-3/push_constants/>`_.
-Using push constants in WGPU closely follows the Vulkan model.
+The wgpu_native backend provides support for immediates.
+Immediates are not yet part of the WebGPU spec, but the headers for native webgpu have converged officially.
```
Vipitis, line 65: it looks like this is rather close to becoming official spec https://github.com/gpuweb/gpuweb/pull/5423

wgpu/backends/wgpu_native/_api.py
```diff
@@ -1787,51 +1796,40 @@ def create_bind_group(
...
+                immediateDataSize=immediate_size,
             )
             c_pipeline_layout_extras.chain.sType = lib.WGPUSType_PipelineLayoutExtras
             # Note that the object returned by ffi.cast() does not own the memory, so we must keep a ref to the uncast object, until wgpu-native has consumed it.
             c_pipeline_layout_next_in_chain = ffi.cast(
                 "WGPUChainedStruct *", c_pipeline_layout_extras
             )
 
-        # H: nextInChain: WGPUChainedStruct *, label: WGPUStringView, bindGroupLayoutCount: int, bindGroupLayouts: WGPUBindGroupLayout *
+        # TODO: there is an immediateSize field in this struct too. so do we even need the extras?
```
Vipitis, line 1824: https://github.com/gfx-rs/wgpu-native/pull/583

wgpu/resources/wgpu.h
```diff
@@ -1,217 +1,1090 @@
...
+     * When this is set, @c wgpuQueueOnCompletedWorkDone callbacks will fire
+     * the next time the device is polled, not when work is actually done on
+     * the GPU.
+     */
     WGPUGLFenceBehaviour_AutoFinish = 0x00000001,
     WGPUGLFenceBehaviour_Force32 = 0x7FFFFFFF
 } WGPUGLFenceBehaviour;
 
-typedef enum WGPUDx12SwapchainKind {
+typedef enum WGPUDx12SwapchainKind
```
Vipitis, line 828: this gets added to extras in #775

Vipitis:
apart from a new instance extra `WGPUNativeDisplayHandle` this should be ready.

The "combined" header is nice for avoiding code duplication and it should theoretically be faster and thinner. I haven't updated the packing or contributor instructions (you need to run codegen locally after downloading the via the script I believe). There could be other problems with people bringing in a custom lib dir etc

wgpu/backends/wgpu_native/_mappings.py
```diff
@@ -113,101 +118,107 @@
...
+    "TextureFormat.etc2-rgb8a1unorm-srgb": 67,
+    "TextureFormat.etc2-rgb8unorm": 64,
+    "TextureFormat.etc2-rgb8unorm-srgb": 65,
+    "TextureFormat.etc2-rgba8unorm": 68,
+    "TextureFormat.etc2-rgba8unorm-srgb": 69,
+    "TextureFormat.r16float": 9,
+    "TextureFormat.r16sint": 8,
+    "TextureFormat.r16snorm": 6,
+    "TextureFormat.r16uint": 7,
+    "TextureFormat.r16unorm": 5,
```
Vipitis: so the mappings point to the enums from `webgpu.h` which don't work correctly in wgpu-native. While the `wgpu.h` enums do actually work. I filed this as an upstream issue: https://github.com/gfx-rs/wgpu-native/issues/596

we could adjust the codegen to use old mappings over the new mappings, or include both (but we generally strip the *native* prefix. Or just ignore it and wait for an upstream fix. I think it might even have worked correctly prior to this update...

CHANGELOG.md
```diff
@@ -17,6 +17,22 @@ Possible sections in each release:
 * Security: in case of vulnerabilities.
 
 
+## [unreleased]
+
+### Changed:
+* update wgpu-native to [29.0.0.0](https://github.com/gfx-rs/wgpu-native/blob/v29.0.0.0/CHANGELOG.md#changed) by @Vipitis in https://github.com/pygfx/wgpu-py/pull/805
+  * updates the `webgpu.h` header to a more recent version
+  * new texture formats for 16-bit uint/sint across r, rg and rgba variants
+  * new features: `texture-formats-tier1`, `texture-formats-tier2`
```
Vipitis: These are not actually supported, I got a commit at home which I forgot to push

wgpu/backends/wgpu_native/_api.py
```diff
@@ -322,23 +322,26 @@ def _get_limits(id: int, device: bool = False, adapter: bool = False):
     """Gets the limits for a device or an adapter"""
     assert device + adapter == 1  # exactly one is set
 
-    # H: chain: WGPUChainedStructOut, maxPushConstantSize: int, maxNonSamplerBindings: int
+    # NOTE: this will get even simpler in the near future: https://github.com/gfx-rs/wgpu-native/pull/575 as maxImmediateSize shouldn't exist in both
```
almarklein, line 326: Marking this; should do another release of wgpu-native and update for this one and for the easier immediates.

wgpu/backends/wgpu_native/_ffi.py
almarklein: Let's go with this approach. Not sure how much it will improve import-time, but won't hurt it in any case.

almarklein: > I haven't updated the packing or contributor instructions (you need to run codegen locally after downloading the via the script I believe)

Nope, the combined header is commited into the repo, and updated at each codegen.

wgpu/backends/wgpu_native/_api.py
Vipitis: I have https://github.com/gfx-rs/wgpu-native/pull/592 which just needs a changelog entry and should be ready for review. 

If this works as I think it should, we can update the idl and remove almost all of immediates from extras and perhaps also use it in the web.

wgpu/backends/wgpu_native/_mappings.py
Vipitis: even on main you can use `r16unorm` if you request the feature `TextureFormat16bitNorm` but you can't use the enum or string because they weren't mapped before. The enum value of `0x00030001` does work.

What changes here is that we now have them mapped - but that way won't work.
Plus it looks like you can't use them as render targets just yet (eventhough that should work).

As wgpu30 is releasing like this week or soon, I expect to get it with wgpu-native too with the improvements being made there. Otherwise I will find some time to clean up the old mappings so it works for us once we upgrade to v30 too

Vipitis: https://github.com/gfx-rs/wgpu-native/pull/600 should make it easier for us

almarklein: Maybe the easiest path forwards is to get that merged, relese wgpu-native, and rebase on that new release. What do you think?

Vipitis: Either is fine.
I think the current state of the PR is as functional as it gets, there is no regression. So we could merge this here and have a good version for v29.0.0.0
The wgpu-native improvements are just simplifications and additions. I haven't tried a local build for of wgpu-native but there is a couple new things I want to test first. 
I am slightly in favor of doing it in two steps, eventhough some of the changes made here will change again directly after. 

So perhaps we just have a wgpu-py release after both are done?

almarklein: Ok, let's merge this and improve later with new upstream fixes.

I did add a commit to fix the wrong texture-format int to be selected. It involves a tweak to the codegen, which will simply not be activated once https://github.com/gfx-rs/wgpu-native/pull/600 is included.

almarklein: 16bit norm *nearly* working in PyGfx: https://github.com/pygfx/pygfx/pull/1305.

Vipitis: I had a patch Similar to this which I threw away. But if this one works out, that alright from my end!

almarklein (approved):

almarklein:
@Vipitis  In terms of releasing:
* Just did a release *before* merging this.
* Let's do the smaller tweaks for the recent updates in wgpu-native.
* Maybe update IDL?
* Then a new release with only the upstream updates.

almarklein:
And thanks for the hard work on this one @Vipitis 🙏

Vipitis:
not sure how much changed in the idl. but some of the newest features around immediates are a mix of idl, webgpu.h and wgpu-core implementation. So you might end up needing a native feature for a bit. As it's likely to be out of sync. 

My wgpu-native PRs got merged so if we have a release for that we should make the simplifications here too. Avoids having breaking changes twice. 

v29 includes some naga fixes for bugs I filed like two years ago and then managed to fix myself over the winter holidays. So it's great to finally get them here.

almarklein:
I made a release of wgpu-native: https://github.com/gfx-rs/wgpu-native/releases/tag/v29.0.1.1

I also updated the IDL. It's not too much. Immediates are now public: https://github.com/pygfx/wgpu-py/pull/815

almarklein:
@Vipitis I'll leave it up to you to bump to latest wgpu-native, since the biggest change is related to the immediates, which you know very well now :)  Or you can wait for wgpu-core 30.0 to land.

PR #810, opened 2026-05-28
Make the triangle example simpler

### LINKED ISSUE #809: Provide minimal example?
This is somewhat of a proposal, to discuss it a bit. 

The triangle and cube examples are the most basic examples we provide. But they are also a bit convoluted because they serve as utilities that can be reused elsewhere. 

Now it's not too difficult to serialize the functions in your head(and skip all the asycnc variants), or when you copy these as a starting point. But for beginners it might add some amount of confusion. And those are the most viewed documents in the repository. 

I propose adding a standalone example like triangle_minimal and perhaps port one more complex example from https://webgpu.github.io/webgpu-samples/ to python. 

Let me know what you think, or if you have any suggestions. I can open a PR for this sometime this week.

Vipitis:
closes #809 

a few open questions:
* do we use the shared global device for convenience?
* should we stick with the colorful triangle or simplify to the red triangle seen in the other hello triangle examples
* the hdr example broke, should we just make the cube example also take the format arg?
* do we want a few more comments to make it a bit of a guide too?

(I don't expect to have any time over the weekend but will find a bit of free space next week)


almarklein:
> do we use the shared global device for convenience?

If different code/libraries in an application use the same device, they can share buffers and textures. E.g. a lib doing compute can visualize the result with a viz library without having to move the data to the CPU. Therefore I want to promote the global device, especially for libraries. Maybe keep it straightforward but mention the global device briefly in a comment?

> should we stick with the colorful triangle or simplify to the red triangle seen in the other hello triangle examples

I liked that it was a bit more colorful, and it demonstrates varyings nicely with a few extra lines in the shader. 

> the hdr example broke, should we just make the cube example also take the format arg?

yeah, if that fixes the hdr example, let's do that.


> do we want a few more comments to make it a bit of a guide too?

Sounds good! I think we're aiming for an example for ppl to explore to understand wgpu, so any comments that help understand what happens is good!

examples/gui_auto.py
```diff
@@ -12,14 +12,12 @@
 
 """
 
-# test_example = true
```
Vipitis, line 15: I disabled this and removed the screenshot. Since we get one for cube and one for triangle anyway. Perhaps not the best idea as we could miss problems

examples/offscreen_hdr.py
```diff
@@ -16,9 +16,9 @@
 from rendercanvas.offscreen import RenderCanvas
 
 try:
-    from .triangle import setup_drawing_sync
```
Vipitis, line 19: ... and problems we missed. I ran this locally and got a messed up image. As I was on main for rendercanvas I had to go back to v2.5.0 to get it to work again. Didn't bisect it fully but between v2.5.0 and v2.6.0 is what I could find on the quick.

I suspect something about the bitorder might be wrong:

Image

I will check the rendercanvas code when I find time. Maybe a numpy kwargs is all we need

almarklein: Maybe width/height reversed?

Vipitis: wasn't it. I tried around with different shapes and all. But the individual pixel values don't make sense. I think it's somewhere in the new bitmapcontext in rendercanvas. I had a quick glance and it looked like it always writes bytes as uint8... so it could also be a platform specific problem with order across bytes. 

I will try some variations and file this for rendercanvas hopefully in a few days.

examples/triangle.py
```diff
@@ -5,176 +5,108 @@
...
+fn fs_main(in: VertexOutput) -> @location(0) vec4f {
+    let physical_color = pow(in.color.rgb, vec3(2.2));  // gamma correct
+    return vec4(physical_color, in.color.a);
 }
 """
 
 
-if __name__ == "__main__":
-    from rendercanvas.auto import RenderCanvas, loop
+# adapter provides allows us to create a single device, which is the general entrypoint to create most wgpu objects.
```
Vipitis:
```suggestion
# adapter is required to create the device object, which is the general entry point to create most wgpu objects.
```

examples/triangle.py
```diff
@@ -5,176 +5,108 @@
...
+# this function gets called for every frame. It ends with submitting a buffer of work onto the GPU queue.
+def drawing_function():
+    command_encoder = device.create_command_encoder()
+    current_texture_view = context.get_current_texture().create_view()
+
+    render_pass = command_encoder.begin_render_pass(
+        color_attachments=[
+            wgpu.RenderPassColorAttachment(
+                view=current_texture_view,
+                clear_value=(0, 1, 0, 1),  # a green background
```
almarklein: Was the green background intentional? Just wondering if you were testing something :D

Vipitis: yeah, there a really odd issue which I have missed by the background being black in our examples. It also shows up with the wgpu-native example so I never raised it here. I think it's an intel bug since it first showed up after a driver update: https://github.com/IGCIT/Intel-GPU-Community-Issue-Tracker-IGCIT/issues/1441

examples/offscreen_hdr.py
Vipitis: the way I understand the current rendercanvas offscreen canvas is that it downloads 8bit bytes at a time. which course is problemaic with 16bit textures. I tried to look at the bytes directly and they were not reassembled and interpreted as 16bit again. This is a limitation with rendercanvas, so we should raise it there and maybe make this example read the texture directly insted of using the rendercanvas context downloader?

examples/triangle.py
almarklein: Odd, even with a solid black `(0, 0, 0, 1)`?  You mind if I make the green darker so it feels more like a background? The green is a bit overwhelming haha

Vipitis: it must be be something with the OS level compositor since the behaviour is to weird. I am not familiar with how these work or where you could even start to debug there, so it's hopefully addressed by them eventually.

feel free to make changes, as always.

examples/offscreen_hdr.py
almarklein: This was already broken on `main`, so let's not let this hold up this pr. I'll have a look into the problem.

almarklein (dismissed):

examples/gui_auto.py
Vipitis: turns out we don't get a triangle screenshot anymore... it's still in the artifacts just as a leftover I think

Vipitis: got it running with the simplified triangle script now. But this means we might lose legacy readmes or externally hosted readme (think pypi) that still point to the removed `triangle_auto.png` file.

Vipitis:
CI failure seems unrelated (also happend on main after the merge: https://github.com/pygfx/wgpu-py/actions/runs/27551541810/job/81438907026)... maybe I will wait a bit and then trigger a rerun later

examples/gui_auto.py
almarklein: They get fixed on a new release, I suppose.

almarklein (approved):

examples/gui_auto.py
Vipitis: okay, good from my end then

PR #812, opened 2026-06-10
fix GPUPromise type annotation

Vipitis:
Little typing improvement that hands down the generic `AwaitedType` for `GPUPromise` to the subclasses. 

I don't like that there is an edge case but couldn't find anything easier to read that handles `BaseGPUPromise[AwaitedType]` in that function with the existing logic.

before:
image

after: 
image

LSP is way happier this way and shows fewer white method names and red squiggles


almarklein (approved):
Nice!

PR #815, opened 2026-06-23
Update idl

almarklein:


docs/advanced.rst
```diff
@@ -0,0 +1,46 @@
...
+You must tell the adapter to create a device that supports immediates,
+and you must tell it the number of bytes of immediates that you are using.
+Overestimating is okay::
+
+    device = adapter.request_device_sync(
+        required_features=["immediates"],
+        required_limits={"max-immediate-size": 256},
+    )
+
+Creating a immediate data struct in your shader code is similar to the way you would create
```
Vipitis:
```suggestion
Creating an immediate data struct in your shader code is similar to the way you would create
```
not 100% sure on this one, and it was originally my own text.

wgpu/backends/wgpu_native/_api.py
```diff
@@ -2992,6 +2963,41 @@ def _end_pipeline_statistics_query(self):
...
+        m, address = get_memoryview_and_address(data)
+
+        # Deal with offset and size
+        range_offset = int(range_offset)
+        data_offset = int(data_offset)
+        size = int(data_size)
+
+        if not (0 <= data_size <= m.nbytes):
+            raise ValueError("Invalid data_size")
+        if not (0 <= data_size <= m.nbytes):
```
Vipitis: This is the same check twice

wgpu/backends/wgpu_native/extras.py
```diff
@@ -80,11 +80,12 @@ def create_pipeline_layout(
     bind_group_layouts: Sequence[GPUBindGroupLayout],
     immediate_size: int,
 ) -> GPUPipelineLayout:
-    return device._create_pipeline_layout(label, bind_group_layouts, immediate_size)
+    return device.create_pipeline_layout(label, bind_group_layouts, immediate_size)
 
 
 def set_immediates(
     render_pass_encoder: GPURenderPassEncoder,
+    visibility,
```
Vipitis: do we keep the old signature for the deprecation warning? If so, would it make sense to have the one from previous release instead of this in between step?

wgpu/_classes.py
```diff
@@ -837,7 +837,8 @@ def create_buffer_with_data(
         buf.unmap()
         return buf
 
-    # IDL: GPUTexture createTexture(GPUTextureDescriptor descriptor); -> USVString label = "", required GPUExtent3D size, GPUIntegerCoordinate mipLevelCount = 1, GPUSize32 sampleCount = 1, GPUTextureDimension dimension = "2d", required GPUTextureFormat format, required GPUTextureUsageFlags usage, sequence<GPUTextureFormat> viewFormats = []
+    # view_formats: Sequence[enums.TextureFormatEnum] = ()) -> GPUTexture:
```
Vipitis: is this additional comment part of the annotation or just a leftover?

Vipitis (commented):
Mostly some cosmetics, additionally the changelog entry could be extended so it reflects the breaking changes/deprecations

wgpu/backends/wgpu_native/extras.py
almarklein: I think it does:
Image

https://github.com/pygfx/wgpu-py/blob/v0.31.1/wgpu/backends/wgpu_native/extras.py#L89-L96

But also, immediates were not present in the last release, so maybe we should just drop it altogether?

wgpu/_classes.py
almarklein: yeah, sorry. Removed!

wgpu/backends/wgpu_native/_api.py
almarklein: Was existing code. I adapted it to what I think it was meant to do.

wgpu/backends/wgpu_native/extras.py
almarklein: I removed it and updated the changelo.

Vipitis (approved):

PR #816, opened 2026-06-23
Bump wgpu-native to v29.0.1.1

Vipitis:
upstream diff: https://github.com/gfx-rs/wgpu-native/compare/v29.0.0.0...v29.0.1.1

Mostly simplifies the immediates implementation code, plus the 16bit norm texture formats now map correctly (still needs the native only feature).

There is a few new native only features and functions I might look into later.

~~(still draft as I want to try the 16bit norm textures and maybe add a test for them)~~


wgpu/backends/wgpu_native/_mappings.py
almarklein, line 345: The "immediates" is still a native feature. Should it not be standard, or a feature in webgpu.h?

Vipitis: it's still a native feature in v29, it should no longer require a feature in the future. I think it might not make v30 either: https://github.com/gfx-rs/wgpu/issues/8556. If I read the spec correctly, it should become an enable extension for wgsl eventually

almarklein: I am confused ... the WebGPU api supports immediates and e.g. `maxImmediateSize`, but does not specify an "immediates" feature.

Vipitis: I believe this means it's always available. It's a wgsl language feature, not a device feature

almarklein: I don't see it in this list: https://gpuweb.github.io/gpuweb/wgsl/#language-extensions-sec, so maybe the idea is for it to be always available? I guess for now users will have to explicitly request it.

Vipitis: It's [`immediate_address_space`](https://gpuweb.github.io/gpuweb/wgsl/#language_extension-immediate_address_space)

on one of my devices it doesn't even show up on https://webgpureport.org/ while the other marks it with a warning. I think it's just so new that the browser implementations haven't caught up. 

There is more discussion here: https://github.com/gpuweb/gpuweb/pull/5423

I read a while ago that wgpu emulates it on openGL via uniform buffers.

Vipitis:
if I find the time tomorrow, I will start a tracking issue with all the native only features and functions that still need to be implemented, tested and explained (docs, examples). Perhaps we can decide which are worth looking into there. With v30 and https://github.com/gfx-rs/wgpu-native/pull/594 there will be even more like raytracing and passthrough which I am motivated in making work.

wgpu/backends/wgpu_native/_mappings.py
almarklein: Ah, looks like I missed it :)

almarklein (approved):

PR #817, opened 2026-06-24
Add preferred features for preconfigure_default_device

almarklein:
This solves two problems:

* Being able to select native features.
* Being able to select features, but providing a fallback if it's not available.

These can be though of as being the same; when running in the browser, the native features are not available.

Very practically, this will allow PyGfx to enable filtered uint16 textures, see https://github.com/pygfx/pygfx/pull/1305


wgpu/utils/device.py
```diff
@@ -74,6 +75,8 @@ def preconfigure_default_device(
                 Setting the adapter overrules all other adapter settings
                 (feature_level, power_preference, force_fallback_adapter, canvas).
             label (str): A human-readable label for the device.
+            preferred_features (list or str): the features (extensions) that you want but do not strictly need.
```
Vipitis:
```suggestion
            preferred_features (set of str): the features (extensions) that you want but do not strictly need.
```

wgpu/utils/device.py
```diff
@@ -195,12 +201,21 @@ def get_default_device(self) -> GPUDevice:
...
+            # Get adapter
             adapter: GPUAdapter = self._adapter_kwargs.pop("adapter", None)
             if adapter is None:
                 adapter = wgpu.gpu.request_adapter_sync(**self._adapter_kwargs)
-            self._the_device = adapter.request_device_sync(**self._device_kwargs)
+            # Handle preferred features
+            kwargs = self._device_kwargs.copy()
+            required_features = kwargs.get("required_features", set())
+            extra_features = kwargs.pop("preferred_features", set())
+            extra_features = {f for f in extra_features if f in adapter.features}
```
Vipitis: as `extra_features` should be a set, this can be an intersection. However I think this way other sequence types will still work

almarklein: I like that!

Vipitis (dismissed):

wgpu/utils/device.py
```diff
@@ -74,6 +75,8 @@ def preconfigure_default_device(
                 Setting the adapter overrules all other adapter settings
                 (feature_level, power_preference, force_fallback_adapter, canvas).
             label (str): A human-readable label for the device.
+            preferred_features (set of str): the features (extensions) that you want but do not strictly need.
+                Check ``device.features`` for its success. Native features are allowed too.
             required_features (list of str): the features (extensions) that you need.
```
almarklein:
```suggestion
            required_features (set of str): the features (extensions) that you need.
```

Vipitis (dismissed):

almarklein:
sry, I took some more time to clean it up just a bit more.

Vipitis (approved):
looks good to me, I like the utility. I got a [private branch](https://github.com/pygfx/wgpu-py/compare/main...Vipitis:wgpu-py:startup) with some ideas regarding #789 that could be added to it in the future.

PR #820, opened 2026-07-02
Context.get_current_texture() can now raise DrawCancelled

### LINKED ISSUE #819: Hiding a GLFW window causes surface error
* Run any wgpu-based example in the glfw canvas backend.
* Hide the window, as in selecting your IDE or something so the glfw window gets behind it.
* Errrors are raised.

This is present on MacOS on `main`. The latest release (v0.31.1) is fine.

The qt backend is fine, since that default to bitmap present-method. When using `present_method='screen'`, it raises a similar error, but only on startup. With the WIP native osx backend it raises both on startup and when hidden.

I suppose that something changed upstream, that causes invisible windows not to have a valid surface? Maybe to prevent spending CPU and GPU cycles when its not visible anyway?

@Vipitis are you aware of any changed related to the surface texture?

```
Draw error
  File "/Users/almar/dev/py/rendercanvas/rendercanvas/core/coreutils.py", line 51, in log_exception
    yield
  File "/Users/almar/dev/py/rendercanvas/rendercanvas/base.py", line 589, in _draw_and_present
    self._draw_frame()
    ~~~~~~~~~~~~~~~~^^
  File "/Users/almar/dev/py/wgpu-py/examples/cube.py", line 345, in draw_func
    draw_frame()
    ~~~~~~~~~~^^
  File "/Users/almar/dev/py/wgpu-py/examples/cube.py", line 308, in draw_frame
    context.get_current_texture().create_view(
    ~~~~~~~~~~~~~~~~~~~~~~~~~~~^^
  File "/Users/almar/dev/py/rendercanvas/rendercanvas/contexts/wgpucontext.py", line 127, in get_current_texture
    return self._get_current_texture()
           ~~~~~~~~~~~~~~~~~~~~~~~~~^^
  File "/Users/almar/dev/py/rendercanvas/rendercanvas/contexts/wgpucontext.py", line 165, in _get_current_texture
    return self._wgpu_context.get_current_texture()
           ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^
  File "/Users/almar/dev/py/wgpu-py/wgpu/_classes.py", line 447, in get_current_texture
    self._texture = self._create_texture_screen()
                    ~~~~~~~~~~~~~~~~~~~~~~~~~~~^^
  File "/Users/almar/dev/py/wgpu-py/wgpu/backends/wgpu_native/_api.py", line 1002, in _create_texture_screen
    raise RuntimeError(
        f"Cannot get surface texture: {status_str} ({status_int})."
    )
RuntimeError: Cannot get surface texture: Unknown (196609).
```

almarklein:
Closes #819.

Since the latest update, when a window is hidden, we cannot obtain the surface texture. We had two options:
* provide a dummy texture to render to.
* raise an exception so downstream code can react appropriately.

Since this is a reliable method to detect that a window is not actually visible, and not all GUI toolkits have a way to determine the visibility of a window, we decided for the latter.

This does mean downstream code needs to adjust. Any code based on rendercanvas can simply update rendercanvas, see https://github.com/pygfx/rendercanvas/pull/221.


Vipitis (approved):
I skimmed the diff and looks good to me. Is the whole dummy thing not needed anymore? Feels great.

Vipitis:
Also is there some reasonable way to automatically test for this? Or would you need gui automation etc? 
Not specifically to expect the error as it might not work with all backend and all guis, but just to spot any panics easier.

almarklein:
> Is the whole dummy thing not needed anymore?

No, because we can use the same frame-skip mechanism for these cases. The dummy texture was there to allow `get_current_texture()` to always Just Work, even if the result would not be visible.


> Also is there some reasonable way to automatically test for this?

The manual test is to run an example with glfw, hide the window, check that there are no errors, and observe that reported fps is low at first when you bring it back to front. Automating that in CI is very tricky at the least.

PR #821, opened 2026-07-02
Bump version

almarklein:
Creating this early to also bump dependency on yet to be released rendercanvas version.


pyproject.toml
```diff
@@ -12,7 +12,7 @@ requires-python = ">= 3.11"
 dependencies = [
     "cffi>=1.15.0",
     "rubicon-objc>=0.4.1; sys_platform == 'darwin'",
-    "rendercanvas >=2.4",                            # Temporarily depend on rendercanvas because we re-aligned apis. Remove in a few months
+    "rendercanvas >=2.7",                            # Temporarily depend on rendercanvas because we re-aligned apis. Remove in a few months
```
almarklein, line 15: Make sure this exists before we merge :)

Vipitis: would you consider it a major release for rendercanvas? (Renderview and new backend). I am not sure if there is anything breaking like with the context inversion we had previously

almarklein: I think there are no breaking changes, but will have a proper look at the time, and update this number accordingly. I mainly started this PR early so we don't forget pinning to a rendercanvas that can deal with the `DrawCancelled`.

PR #828, opened 2026-08-16
Allow `GPUTexture` as resource in bind group entries

### LINKED ISSUE #825: `GPUBindGroupEntry` spec divergence: what should we do?
Okay, since `webgpu.h` is considered stable it's different to the `webgpu.idl` in a few places. 
One place where this happened to me is bind group entries. The struct (derived from .idl) allows for both `texture` and `texture_view` while the function that uses this (`create_bind_group`) only accepts the view: 
https://github.com/pygfx/wgpu-py/blob/9d59b27fab7a15c92cd1fc147912b889f08e93d3/wgpu/backends/wgpu_native/_api.py#L1694
I asked the spec people and it won't be updated: https://github.com/webgpu-native/webgpu-headers/issues/597

For the user it's at most a single call to `create_view()`, however if you strictly follow the typing from the structs you might run into it. In the future this could become a problem when code written for pyodide no longer works with native. 

The obvious best solution would be to just add a two line check and call `create_view()` and allow it, following the .idl without needing to make it an apidiff.

If you check the upstream spec change PR, there is additional places where this might show up, so perhaps we need to also adjust our codegen to spot these differences as we continue to update .idl too. 

I got quite a lot of stress in in front of me with deadlines, so it might be several weeks for me to have a PR for this. But it's also a very low chance of impacting anyone.

Vipitis:
fixes #825 

Also adds a test for all the different resource types. (although not all layout combinations are included and I don't actually run anything there.
Had to help out the gc to reuse code, I tried a new variable but that also had trouble.

Also spotted that for color attachment (in begin render pass). there is similar logic: https://www.w3.org/TR/webgpu/#color-attachments and the spec even has a helper method for it we could adopt.


almarklein (approved):
Nice!

PR #830, opened 2026-08-18
Use len(draw_data.cmd_lists) in imgui backend

### LINKED ISSUE #829: cmd_list_counts is no longer available in latest imgui_bundle
https://github.com/pygfx/wgpu-py/blob/9d59b27fab7a15c92cd1fc147912b889f08e93d3/wgpu/utils/imgui/imgui_backend.py#L448

I think you need len(draw_data.cmd_lists)

hmaarrfk:
closes https://github.com/pygfx/wgpu-py/issues/829

see failing tests in https://github.com/pygfx/wgpu-py/pull/831

stays compatible with 1.19.X

<details><summary>Claude's draft</summary>

`ImDrawData.cmd_lists_count` was a legacy alias documented as `== CmdLists.Size`, and imgui_bundle removed it in 1.92.900. Every `ImguiWgpuBackend.render()` call now raises AttributeError against that release. Fixes #829.

`len(draw_data.cmd_lists)` is exactly what the removed attribute equalled and works on all releases in the supported `imgui-bundle>=1.92.0,<2` range, verified against 1.92.0, 1.92.3, 1.92.4, 1.92.5, 1.92.600, 1.92.601, 1.92.700, 1.92.801 and 1.92.900. No shim or version check is needed, and the pin does not need narrowing.

Add tests/test_util_imgui.py, which renders one real frame through ImguiRenderer on an offscreen canvas. The existing example tests never enter the backend's render() (the offscreen canvas does not pump a draw), so they pass even with the bug present. The "Test Examples" job is the only CI job that installs imgui_bundle, so it now runs that test file too; adding imgui-bundle to the `tests` extra was rejected because that extra is used by the 3.14 and pypy3.11 matrix jobs.

Resume this Claude session:
```
cd /home/mark/git/wgpu-py
claude --resume 590dd7bc-312c-419e-ad3d-2c69d00943cd
```
</details>


almarklein (approved):
Thanks!

PR #832, opened 2026-08-22
Implement view_formats in create_texture

hmaarrfk:
Allow users to create non-srgb textures through the usage of `view_formats`

-----------------------------------------------------------------

`create_texture()` accepts a `view_formats` argument and then raises `NotImplementedError` for any non-empty value, while the descriptor it fills already carries `viewFormatCount` / `viewFormats`, commented out as unused. This wires them up with the same conversion the surface configuration path in the same file already performs, which makes it possible to take a non-srgb view of an srgb target instead of copying the whole texture into a second one of the plain format every frame. The test covers a declared format working, the texture's own format, an undeclared format still being rejected, and a texture declaring none keeping the previous behaviour.


almarklein (approved):
Looks good to me, thanks! Support for view formats open up some new use-cases 🚀

